PR #37, opened 2024-01-29
Add INSTALLATION guide

joaopalet:


INSTALLATION.md
```diff
@@ -0,0 +1,56 @@
+# Installation
+
+## Linux
```
vicentepinto98, line 3: homebrew also works for Linux, so it would be worth mentioning it here

README.md
```diff
@@ -7,33 +7,7 @@ Your feedback is appreciated!
...
-   ```
-
-2. Skip building and run the Go application directly using:
-
-   ```bash
-   $ go run . <GROUP> <SUB-GROUP> <COMMAND> <ARGUMENT> <FLAGS>
-   ```
-
-We also plan to integrate the STACKIT CLI on package managers such as APT and Brew.
+Please refer to our [installation](./INSTALLATION.md) guide for instructions on how to install and get started using the STACKIT CLI.
```
vicentepinto98, line 10: I wonder if it makes sense to have some of the installation commands directly on the README, to be quicker to start using the CLI when opening the repo. How do others do it?

INSTALLATION.md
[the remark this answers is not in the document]
joaopalet: Good point, I will add it

README.md
joaopalet: Some seem to just redirect to a page with an installation guide while others show the instructions directly in the README

I moved it to another file so the README wouldn't get too long once we start adding instructions for all OSs and package managers

PR #38, opened 2024-01-29
Sign Linux packages

joaopalet:


.github/workflows/release.yaml
```diff
@@ -26,9 +26,16 @@ jobs:
         with:
           go-version-file: "go.mod"
           cache: true
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
```
vicentepinto98, line 30: there is no built-in action from GitHub to import a gpg key? Seems like basic functionality so its strange to use some third-party for this

joaopalet: I don't think so, this is the one we use in the Terraform provider pipeline too. Also e.g. Hashicorp had developed their own action but have recently deprecated it in favor of this one (see [here](https://github.com/hashicorp/ghaction-import-gpg))

vicentepinto98 (approved):

PR #40, opened 2024-01-30
Add instance type and replicas to MongoDB Flex instance describe pret…

joaopalet:
…ty output

### DIFF
--- a/internal/cmd/mongodbflex/instance/describe/describe.go
+++ b/internal/cmd/mongodbflex/instance/describe/describe.go
@@ -94,7 +94,8 @@ func outputResult(cmd *cobra.Command, outputFormat string, instance *mongodbflex
 
 		instanceType, err := mongodbflexUtils.GetInstanceType(*instance.Replicas)
 		if err != nil {
-			instanceType = "Unknown"
+			// Should never happen
+			instanceType = ""
 		}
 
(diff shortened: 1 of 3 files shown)


internal/cmd/mongodbflex/instance/describe/describe.go
```diff
@@ -91,6 +92,11 @@ func outputResult(cmd *cobra.Command, outputFormat string, instance *mongodbflex
 		acls := *instance.Acl.Items
 		strings.Join(acls, ",")
 
+		instanceType, err := mongodbflexUtils.GetInstanceType(*instance.Replicas)
+		if err != nil {
+			instanceType = "Unknown"
```
vicentepinto98, line 97: maybe worth asking the team what we should show in this case. Not sure if `unknown` is the best option. Could also be just "-" like we show in other cases, or just the empty string

vicentepinto98 (approved):

internal/cmd/mongodbflex/instance/describe/describe.go
joaopalet: This should never happen as only 1, 3 and 9 are allowed for number of replicas, I've added a comment and changed it to be `""`

vicentepinto98 (approved):

PR #41, opened 2024-01-30
Onboard OpenSearch service

joaopalet:


internal/cmd/config/set/set.go
```diff
@@ -23,6 +23,7 @@ const (
 	serviceAccountCustomEndpointFlag  = "service-account-custom-endpoint"
 	skeCustomEndpointFlag             = "ske-custom-endpoint"
 	resourceManagerCustomEndpointFlag = "resource-manager-custom-endpoint"
+	OpenSearchCustomEndpointFlag      = "opensearch-custom-endpoint"
```
hcsa73, line 26:
```suggestion
	openSearchCustomEndpointFlag      = "opensearch-custom-endpoint"
```
Unless there's a reason for this to be public

internal/cmd/opensearch/instance/create/create.go
```diff
@@ -0,0 +1,253 @@
...
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create an OpenSearch instance",
+		Long:  "Create an OpenSearch instance.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create an OpenSearch instance with name "my-instance" and specify plan by name and version`,
+				"$ stackit opensearch instance create --name my-instance --plan-name stackit-qa-opensearch-1.2.10-replica --version 2"),
```
hcsa73, line 67: Change plan name to not use one from QA

internal/cmd/opensearch/instance/create/create.go
```diff
@@ -0,0 +1,253 @@
...
+				`Create an OpenSearch instance with name "my-instance" and specify plan by ID`,
+				"$ stackit opensearch instance create --name my-instance --plan-id xxx"),
+			examples.NewExample(
+				`Create an OpenSearch instance with name "my-instance" and specify IP range which is allowed to access it`,
+				"$ stackit opensearch instance create --name my-instance --plan-id xxx --acl 192.168.1.0/24"),
+		),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			// Service name and operation needed for error handling
+			service := strings.Split(cmd.Parent().Parent().Use, " ")[0]
```
hcsa73, line 78: Isn't this just "opensearch"

hcsa73: Also below, `operation` is just `"create"`

I guess you're doing this so it can be copy-pasted for other DSA services. I don't like it, because you're making the code harder to read, while not really saving on maintenance effort (since changes here would have to be copy-pasted for other services). It's a middle ground that ends up taking the worst of both worlds. Also, this assumes this command has a parent's parent, and that the grandparent's `Use` starts with the name of the service, which is a very obscure dependency to have.

If you want to take advantage of the DSA APIs being the same, I'd suggest making a `services/dsa` package that holds all the commands core logic. This `RunE` would then just calls `CreateRunE` from `pkg/services/dsa` (or something similar). And that `CreateRunE` would get as input a `dsaClient` interface that opensearch/mariadb/... clients satisfy (which works because the APIs are the same).
Alternatively, you can just not rely on that similarity and have this as a normal command, use `"opensearch"` instead of the `service` variable and so on.

internal/pkg/services/opensearch/utils/utils.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			if strings.EqualFold(*plan.Name, planName) && plan.Id != nil {
+				return plan.Id, nil
+			}
+			availablePlanNames = fmt.Sprintf("%s\n- %s", availablePlanNames, *plan.Name)
+		}
+	}
+
+	if !isValidVersion {
+		details := fmt.Sprintf("You provided version '%s', which is invalid. Available versions are: %s", version, availableVersions)
```
hcsa73, line 50:
```suggestion
		details := fmt.Sprintf("You provided version %q, which is invalid. Available versions are: %s", version, availableVersions)
```

internal/pkg/services/opensearch/utils/utils.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, &errors.DSAInvalidPlanError{
+			Service: service,
+			Details: details,
+		}
+	}
+	details := fmt.Sprintf("You provided plan_name '%s' for version %s, which is invalid. Available plan names for that version are: %s", planName, version, availablePlanNames)
```
hcsa73, line 56:
```suggestion
	details := fmt.Sprintf("You provided plan_name %q for version %s, which is invalid. Available plan names for that version are: %s", planName, version, availablePlanNames)
```

internal/pkg/services/opensearch/utils/utils.go
```diff
@@ -0,0 +1,82 @@
...
+		for _, plan := range *offer.Plans {
+			if plan.Id != nil && strings.EqualFold(*plan.Id, planId) {
+				return nil
+			}
+		}
+	}
+
+	return &errors.DSAInvalidPlanError{
+		Service: service,
+		Details: fmt.Sprintf("You provided plan ID '%s', which is invalid.", planId),
```
hcsa73, line 23:
```suggestion
		Details: fmt.Sprintf("You provided plan ID %q, which is invalid.", planId),
```

hcsa73 (changes requested):

internal/cmd/opensearch/instance/create/create.go
joaopalet: We also had it for Postgres and MongoDB Flex but I don't remember the reason, I've replaced it by the explicit service name

In the end we also have it in other places such as the examples, and it can be easily replaced by using the replace tool

joaopalet: I'm not sure about having the common code for the DSA, as there's nothing enforcing them to behave similarly AFAIK, so I would suggest keeping it as is and re-evaluate in the future if we have too much effort to maintain them

internal/cmd/opensearch/instance/create/create.go
```diff
@@ -0,0 +1,253 @@
...
+		Version:              version,
+	}, nil
+}
+
+type OpenSearchClient interface {
+	CreateInstance(ctx context.Context, projectId string) opensearch.ApiCreateInstanceRequest
+	ListOfferingsExecute(ctx context.Context, projectId string) (*opensearch.ListOfferingsResponse, error)
+}
+
+func buildRequest(ctx context.Context, service string, model *inputModel, apiClient OpenSearchClient) (opensearch.ApiCreateInstanceRequest, error) {
```
hcsa73, line 205: Same comment as `parseInput`'s signature: `service` looks unnecessary in `buildRequest`'s signature

internal/cmd/opensearch/instance/create/create.go
```diff
@@ -0,0 +1,253 @@
...
+	var planId *string
+	var err error
+
+	offerings, err := apiClient.ListOfferingsExecute(ctx, model.ProjectId)
+	if err != nil {
+		return req, fmt.Errorf("get OpenSearch offerings: %w", err)
+	}
+
+	if model.PlanId == nil {
+		planId, err = opensearchUtils.LoadPlanId(service, model.PlanName, model.Version, offerings)
```
hcsa73, line 217: Same comment as `parseInput`'s signature: `service` looks unnecessary in `LoadPlanId`'s signature
Here it's particularly weird, because it's coming from a package named `opensearchUtils`, it really shouldn't need to be told what service this is

internal/cmd/opensearch/instance/create/create.go
```diff
@@ -0,0 +1,253 @@
...
+		if err != nil {
+			var dsaInvalidPlanError *cliErr.DSAInvalidPlanError
+			if !errors.As(err, &dsaInvalidPlanError) {
```
hcsa73, line 220: This isn't very Go idiomatic, I think:
- Error types usually are named `Err...`
- `errors.Is` is what should be used to check if a specific error type is in the chain. `errors.As` is used to extract an error type from the chain

For the second point, you can check how it would look like here: https://go.dev/play/p/oMyqrb9Znoq. Note the `Is` method is required for `errors.Is` to work

internal/pkg/services/opensearch/utils/utils.go
```diff
@@ -0,0 +1,82 @@
...
+
+func GetInstanceName(ctx context.Context, apiClient OpenSearchClient, projectId, instanceId string) (string, error) {
+	resp, err := apiClient.GetInstanceExecute(ctx, projectId, instanceId)
+	if err != nil {
+		return "", fmt.Errorf("get OpenSearch instance: %w", err)
+	}
+	return *resp.Name, nil
+}
+
+func GetCredentialUsername(ctx context.Context, apiClient OpenSearchClient, projectId, instanceId, credentialsId string) (string, error) {
```
hcsa73, line 76:
```suggestion
func GetCredentialsUsername(ctx context.Context, apiClient OpenSearchClient, projectId, instanceId, credentialsId string) (string, error) {
```
(Refactor in IDE so it changes references)

internal/pkg/services/opensearch/utils/utils_test.go
```diff
@@ -0,0 +1,144 @@
...
+				return
+			}
+			if output != tt.expectedOutput {
+				t.Errorf("expected output to be %s, got %s", tt.expectedOutput, output)
+			}
+		})
+	}
+}
+
+func TestCredentialUsername(t *testing.T) {
```
hcsa73, line 94:
```suggestion
func TestGetCredentialsUsername(t *testing.T) {
```

internal/cmd/opensearch/instance/create/create.go
```diff
@@ -0,0 +1,253 @@
...
+	SgwAcl               *[]string
+	Syslog               *[]string
+	PlanId               *string
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create an OpenSearch instance",
+		Long:  "Create an OpenSearch instance.",
```
hcsa73, line 62: Nitpick: Do we have a standard on what help descriptions end with full stop or not

joaopalet: The way I see it:
- `Short`: is not a full sentence, it's just a short description of the command that shows in one line for each of the child commands of a given command
- `Long`: might be composed of multiple sentences spanning over multiple lines, so I think it should be a full sentence and end with `.`s

joaopalet: Take e.g. `gh pr list`.

- `Short`
image

- `Long`: 
image

internal/cmd/opensearch/credentials/delete/delete.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("delete OpenSearch credentials: %w", err)
+			}
+
+			cmd.Printf("Deleted credential %s of instance %s\n", credentialLabel, instanceLabel)
```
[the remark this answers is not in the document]
joaopalet: Fixed for the `create` command too

hcsa73 (approved):

internal/cmd/opensearch/instance/update/update.go
```diff
@@ -0,0 +1,265 @@
...
+	sgwAcl := flags.FlagToStringSlicePointer(cmd, sgwAclFlag)
+	syslog := flags.FlagToStringSlicePointer(cmd, syslogFlag)
+	planId := flags.FlagToStringPointer(cmd, planIdFlag)
+	planName := flags.FlagToStringValue(cmd, planNameFlag)
+	version := flags.FlagToStringValue(cmd, versionFlag)
+
+	if planId != nil && (planName != "" || version != "") {
+		return nil, &cliErr.DSAInputPlanError{
+			Service:   service,
+			Operation: operation,
```
hcsa73, line 180: `cmd.Use` is `"update INSTANCE_ID"`, is this intended or is just `"update"` intended?

joaopalet: Yes, I think it makes sense so we can see it in the error help:
image

joaopalet: Similarly in mongodbflex:
image

joaopalet: But we could improve even more by providing the actual instance ID

joaopalet: But I would do it in a separate PR also, as it impacts more services

hcsa73 (approved):

internal/cmd/opensearch/instance/create/create.go
vicentepinto98: This not applied to the other added commands. They are missing the `.` on the long descriptions

joaopalet: I will create a separate PR to make it uniform in all CLI commands

PR #47, opened 2024-02-02
Rename Homebrew tap and Snap package to `stackit`

joaopalet:


.goreleaser.yaml
```diff
@@ -64,15 +64,15 @@ signs:
     args: ["-u", "{{ .Env.GPG_FINGERPRINT }}", "--output", "${signature}", "--detach-sign", "${artifact}"]
 
 brews:
-  - name: stackit-cli
+  - name: stackit
```
hcsa73, line 67: Do we also want to change the brew name?

hcsa73 (approved):

PR #52, opened 2024-02-02
Docs: Update INSTALLATION guide

joaopalet:


INSTALLATION.md
```diff
@@ -13,44 +15,54 @@ brew tap stackitcloud/tap
...
+> For more information, please refer to the [`go install` documentation](https://go.dev/ref/mod#go-install)
+
+Alternativelly, you can:
 
 1. Clone the repository
 2. Build the application locally by running:
 
    ```bash
-   $ make build
+   make build
```
joaopalet, line 49: We don't need the `$` as we are already inside a shell code block and the user can just copy the command using 
image

vicentepinto98 (approved):

PR #56, opened 2024-02-05
Implement `postgresflex instance`

hcsa73:


internal/pkg/services/postgresflex/utils/utils_test.go
```diff
@@ -0,0 +1,83 @@
+package utils
```
vicentepinto98, line 1: We should also test the other three functions in `utils`: `ValidateStorage`, `ValidateFlavorId` and `LoadFlavorId`

internal/cmd/postgresflex/postgresflex.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "postgresflex",
+		Short: "Provides functionality for PostgreSQL Flex",
+		Long:  "Provides functionality for PostgreSQL Flex",
```
vicentepinto98, line 15:
```suggestion
		Long:  "Provides functionality for PostgreSQL Flex.",
```

internal/cmd/postgresflex/instance/instance.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "instance",
+		Short: "Provides functionality for PostgreSQL Flex instances",
+		Long:  "Provides functionality for PostgreSQL Flex instances",
```
vicentepinto98, line 19:
```suggestion
		Long:  "Provides functionality for PostgreSQL Flex instances.",
```

internal/cmd/postgresflex/instance/describe/describe.go
```diff
@@ -0,0 +1,128 @@
...
+func buildRequest(ctx context.Context, model *inputModel, apiClient *postgresflex.APIClient) postgresflex.ApiGetInstanceRequest {
+	req := apiClient.GetInstance(ctx, model.ProjectId, model.InstanceId)
+	return req
+}
+
+func outputResult(cmd *cobra.Command, outputFormat string, instance *postgresflex.Instance) error {
+	switch outputFormat {
+	case globalflags.PrettyOutputFormat:
+		acls := *instance.Acl.Items
+		strings.Join(acls, ",")
```
vicentepinto98, line 92: don't you need to assign this to the `acls` variable? In the `mongodbflex` it should also be fixed

internal/cmd/postgresflex/instance/create/create.go
```diff
@@ -0,0 +1,266 @@
...
+	flavorIdFlag       = "flavor-id"
+	cpuFlag            = "cpu"
+	ramFlag            = "ram"
+	replicasFlag       = "replicas"
+	storageClassFlag   = "storage-class"
+	storageSizeFlag    = "storage-size"
+	versionFlag        = "version"
+	typeFlag           = "type"
+
+	defaultBackupSchedule = "0 0/6 * * *"
```
vicentepinto98, line 38: was this checked with the team?

internal/cmd/postgresflex/instance/describe/describe.go
```diff
@@ -0,0 +1,128 @@
...
+
+		table := tables.NewTable()
```
vicentepinto98, line 94: what about the `TYPE` and `REPLICAS` fields? We should show those, as we do for `mongofbflex`

internal/cmd/postgresflex/postgresflex.go
hcsa73: Applied, together with changes following #57

internal/cmd/postgresflex/instance/describe/describe.go
hcsa73: Good catch!

internal/cmd/postgresflex/instance/describe/describe.go
hcsa73: Replicas is just missing, by bad
For type I'll double-check with the team

internal/cmd/postgresflex/instance/create/create.go
vicentepinto98: you can ask @joaopalet why he decided to go for the hardcoded default version instead of fetching the latest one from the API. 

Fetching the latest would be the best I agree, but we should be careful to not introduce too many calls that would make the CLI slow

joaopalet: At the time I used the default value confirmed with the team and didn't think about always getting the most recent value from the API

internal/pkg/services/postgresflex/utils/utils_test.go
hcsa73: Added, for mongodb as well

internal/cmd/postgresflex/instance/create/create.go
hcsa73: Aligned defaults with Portal, for version it pulls the latest version from the API

internal/cmd/postgresflex/instance/describe/describe.go
hcsa73: Added, aligned mapping of replicas <--> type to what is in the Portal, and tested to check they work

internal/cmd/postgresflex/instance/create/create.go
hcsa73: Will change for mongodbflex in a separate PR

vicentepinto98 (approved):

PR #58, opened 2024-02-05
Refactor DSA input plan error

joaopalet:


internal/pkg/errors/errors.go
```diff
@@ -37,14 +38,14 @@ For more details run:
 
 	DSA_INVALID_INPUT_PLAN = `the instance plan was not correctly provided.
```
joaopalet, line 39: The output will be something like:
image

hcsa73 (approved):

PR #60, opened 2024-02-06
Onboard Redis and RabbitMQ

joaopalet:


internal/pkg/services/opensearch/utils/utils.go
joaopalet, line 1: New DSAs were copied over from OpenSearch implementation, only change made was removing the `service` parameter in this call and making it a `const` at the service util level

hcsa73 (commented):
LGTM, just docs missing

hcsa73 (approved):

PR #63, opened 2024-02-06
MongoDBFlex instance creation uses latest available version by default

hcsa73:


joaopalet (approved):

internal/cmd/mongodbflex/instance/create/create.go
```diff
@@ -100,6 +99,15 @@ func NewCmd() *cobra.Command {
 				}
 			}
 
+			// Fill in version, if needed
```
joaopalet, line 102: Is there any specific reason this was not added to the `buildRequest` function instead? We are already also filling the flavor ID there

joaopalet (approved):

PR #64, opened 2024-02-06
Refactor database input flavor error

joaopalet:


internal/pkg/errors/errors.go
```diff
@@ -57,13 +57,13 @@ For more details on the available plans, run:
 	DATABASE_INVALID_INPUT_FLAVOR = `the instance flavor was not correctly provided. 
 
 Either provide flavor ID by:
-  $ stackit %[1]s instance %[2]s --project-id xxx --flavor-id <FLAVOR ID> [flags]
+  $ %[1]s --flavor-id <FLAVOR ID> [flags]
```
hcsa73, line 60: The `--project-id` part isn't shown anymore, right?
I think it's fine, because we want to make it obvious we're talking about these flags

hcsa73 (approved):

internal/pkg/errors/errors.go
joaopalet: Yes, as well as all other possible flags the user was using (hence the `[flags]` at the end). I think it's easier to understand this way

PR #66, opened 2024-02-07
Send service account key creation to stderr

vicentepinto98:
- To support use case of saving the result of service-account key creation directly in a file

### DIFF
--- a/internal/cmd/service-account/key/create/create.go
+++ b/internal/cmd/service-account/key/create/create.go
@@ -86,7 +86,7 @@ func NewCmd() *cobra.Command {
 				return fmt.Errorf("create service account key: %w", err)
 			}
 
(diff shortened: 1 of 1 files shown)


internal/cmd/service-account/key/create/create.go
```diff
@@ -86,7 +86,7 @@ func NewCmd() *cobra.Command {
 				return fmt.Errorf("create service account key: %w", err)
 			}
 
-			cmd.Printf("Created key for service account %s\n", model.ServiceAccountEmail)
+			cmd.PrintErr(fmt.Sprintf("Created key for service account %s\n", model.ServiceAccountEmail))
```
joaopalet, line 89:
```suggestion
			cmd.PrintErrf("Created key for service account %s\n", model.ServiceAccountEmail)
```

joaopalet (approved):

joaopalet (approved):

PR #70, opened 2024-02-08
Postgresflex user fixes

hcsa73:
- Fix roles available
- Make `--username` flag required on creation
- Change `--roles` flag name to `--role`
- Remove incorrect UUID check of user ID
- Add check for args, check if args aren't empty

### DIFF
--- a/internal/cmd/mongodbflex/user/create/create.go
+++ b/internal/cmd/mongodbflex/user/create/create.go
@@ -113,7 +113,7 @@ func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().Var(flags.UUIDFlag(), instanceIdFlag, "ID of the instance")
 	cmd.Flags().String(usernameFlag, "", "Username of the user. If not specified, a random username will be assigned")
 	cmd.Flags().String(databaseFlag, "", "The database inside the MongoDB instance that the user has access to. If it does not exist, it will be created once the user writes to it")
-	cmd.Flags().Var(flags.EnumSliceFlag(false, nil, roleOptions...), roleFlag, fmt.Sprintf("Roles of the user, can be specified multiple times. Possible values are %q", roleOptions))
+	cmd.Flags().Var(flags.EnumSliceFlag(false, rolesDefault, roleOptions...), roleFlag, fmt.Sprintf("Roles of the user, can be specified multiple times. Possible values are %q", roleOptions))
 
 	err := flags.MarkFlagsRequired(cmd, instanceIdFlag, databaseFlag)
 	cobra.CheckErr(err)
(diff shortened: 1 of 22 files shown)


internal/cmd/postgresflex/user/delete/delete.go
```diff
@@ -45,7 +44,7 @@ func NewCmd() *cobra.Command {
 				`Delete a PostgreSQL Flex user with ID "xxx" for instance with ID "yyy"`,
 				"$ stackit postgresflex user delete xxx --instance-id yyy"),
 		),
-		Args: args.SingleArg(userIdArg, utils.ValidateUUID),
+		Args: args.SingleArg(userIdArg, nil),
```
joaopalet, line 47: Is this only for the PotsgresFlex? What is the form of a user ID then?

internal/pkg/args/args.go
```diff
@@ -30,6 +30,13 @@ func SingleArg(argName string, validate func(value string) error) cobra.Position
 				Count:    len(args),
 			}
 		}
+		if args[0] == "" {
+			return &errors.SingleArgExpectedError{
+				Cmd:      cmd,
+				Expected: argName,
+				Count:    len(args),
+			}
+		}
```
joaopalet, line 39: Probably could be merged into the condition above with `||`:

```go
if len(args) != 1 || args[0] == "" {
    ...
```

internal/cmd/postgresflex/user/create/create.go
joaopalet, line 1: The `CreateUser` endpoint response has a `database` field. This does not seem to be configurable on creation but the field can be updated in the PATCH endpoint.

So maybe we should:
- Add the `Database` field in the output here
- Allow the user to update the `database` in the `update` command

internal/cmd/mongodbflex/user/create/create.go
```diff
@@ -108,12 +108,12 @@ func NewCmd() *cobra.Command {
...
 
 func configureFlags(cmd *cobra.Command) {
-	rolesOptions := []string{"read", "readWrite"}
+	roleOptions := []string{"read", "readWrite"}
 
 	cmd.Flags().Var(flags.UUIDFlag(), instanceIdFlag, "ID of the instance")
 	cmd.Flags().String(usernameFlag, "", "Username of the user. If not specified, a random username will be assigned")
 	cmd.Flags().String(databaseFlag, "", "The database inside the MongoDB instance that the user has access to. If it does not exist, it will be created once the user writes to it")
-	cmd.Flags().Var(flags.EnumSliceFlag(false, rolesDefault, rolesOptions...), rolesFlag, fmt.Sprintf("Roles of the user, possible values are %q", rolesOptions))
+	cmd.Flags().Var(flags.EnumSliceFlag(false, nil, roleOptions...), roleFlag, fmt.Sprintf("Roles of the user, can be specified multiple times. Possible values are %q", roleOptions))
```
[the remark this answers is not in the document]
hcsa73: Ah, usage says it's of type `strings` instead of `String`. I thought it didn't show difference
Removed

internal/cmd/postgresflex/user/delete/delete.go
hcsa73: Seems to be 5 digit numbers

internal/pkg/args/args.go
hcsa73: Done, added empty arg test case as well

joaopalet (approved):

PR #75, opened 2024-02-08
Config fixes

hcsa73:
- Add `--session-time-limit` to `config unset`
- Change `config unset` behavior to set as the default, instead of setting to `""` (so it reverts `config set`)
- Fix `config list` not showing bool values
- Change `--async` default to bool type (so its value in the config file is always bool -- before it was bool or string)
- Rewrite descriptions. Most notably,
  - Move the config logic with env vars from `config set` to `config list`
  - Move flag details from examples to the flag descriptions


internal/cmd/config/list/list.go
```diff
@@ -19,8 +20,16 @@ func NewCmd() *cobra.Command {
...
 		Use:   "list",
 		Short: "Lists the current CLI configuration values",
-		Long:  "Lists the current CLI configuration values.",
-		Args:  args.NoArgs,
+		Long: fmt.Sprintf("%s\n%s\n%s\n%s\n%s\n%s\n%s",
+			"Lists the current CLI configuration values, based on the following sources (in order of precedence):",
+			"- Environment variable",
+			`  The environment variable is the name of the setting, with underscores ("_") instead of dashes ("-") and the "STACKIT" prefix.`,
+			"  Example: you can set the project ID by setting the environment variable STACKIT_PROJECT_ID.",
+			"- Configuration set in CLI",
```
joaopalet, line 28: Would it be clearer if rephrased to

```suggestion
			"- Configuration set in the config file",
```

internal/cmd/config/set/set.go
```diff
@@ -42,22 +42,17 @@ func NewCmd() *cobra.Command {
...
 		Use:   "set",
 		Short: "Sets CLI configuration options",
-		Long: fmt.Sprintf("%s\n%s\n%s\n%s",
-			"Sets CLI configuration options.",
-			"All of the configuration options can be set using an environment variable, which takes precedence over what is configured.",
-			`The environment variable is the name of the flag, with underscores ("_") instead of dashes ("-") and the "STACKIT" prefix.`,
-			"Example: to set the project ID you can set the environment variable STACKIT_PROJECT_ID.",
-		),
-		Args: args.NoArgs,
+		Long:  "Sets CLI configuration options.",
```
joaopalet, line 45: I think the information that was here could still be useful.
Even more if we added the note that this command writes **to the config file**, and that those values can be overritten by env variables

internal/cmd/config/set/set.go
```diff
@@ -42,22 +42,17 @@ func NewCmd() *cobra.Command {
...
-				`Set a project ID in your active configuration. This project ID will be used by every command, as long as it's not overridden by the "STACKIT_PROJECT_ID" environment variable or the command flag`,
+				`Set a project ID in your active configuration. This project ID will be used by every command (as long as it's not overridden by the "STACKIT_PROJECT_ID" environment variable)`,
 				"$ stackit config set --project-id xxx"),
 			examples.NewExample(
-				`Set the session time limit to 1 hour. After this time you will be prompted to login again to be able to execute commands that need authentication`,
+				`Set the session time limit to 1 hour`,
 				"$ stackit config set --session-time-limit 1h"),
 			examples.NewExample(
-				`Set the DNS custom endpoint. This endpoint will be used on all calls to the DNS API, unless overridden by the "STACKIT_DNS_CUSTOM_ENDPOINT" environment variable`,
+				`Set the DNS custom endpoint`,
```
joaopalet, line 55: Same for these example descriptions

internal/cmd/config/unset/unset.go
```diff
@@ -134,18 +142,20 @@ func configureFlags(cmd *cobra.Command) {
...
-	cmd.Flags().Bool(mariaDBCustomEndpointFlag, false, "MariaDB custom endpoint")
-	cmd.Flags().Bool(mongoDBFlexCustomEndpointFlag, false, "MongoDB Flex custom endpoint")
-	cmd.Flags().Bool(openSearchCustomEndpointFlag, false, "OpenSearch custom endpoint")
-	cmd.Flags().Bool(postgresFlexCustomEndpointFlag, false, "PostgreSQL Flex custom endpoint")
-	cmd.Flags().Bool(rabbitMQCustomEndpointFlag, false, "RabbitMQ custom endpoint")
-	cmd.Flags().Bool(redisCustomEndpointFlag, false, "Redis custom endpoint")
-	cmd.Flags().Bool(resourceManagerCustomEndpointFlag, false, "Resource Manager custom endpoint")
-	cmd.Flags().Bool(serviceAccountCustomEndpointFlag, false, "SKE custom endpoint")
-	cmd.Flags().Bool(skeCustomEndpointFlag, false, "SKE custom endpoint")
+	cmd.Flags().Bool(sessionTimeLimitFlag, false, fmt.Sprintf("Maximum time before authentication is required again. If unset, defaults to %s", config.SessionTimeLimitKey))
```
joaopalet, line 145: I think you meant
```suggestion
	cmd.Flags().Bool(sessionTimeLimitFlag, false, fmt.Sprintf("Maximum time before authentication is required again. If unset, defaults to %s", config.SessionTimeLimitDefault))
```

internal/cmd/config/list/list.go
hcsa73: I phrased this here to abstract from the config file, so you don't have to know of its existence
Basically tried to write "whatever was set using `stackit config set/unset`" (which ends up in the config file)
Not sure if it was the best choice, what do you think?

internal/cmd/config/set/set.go
hcsa73: Same thing as https://github.com/stackitcloud/stackit-cli/pull/75#discussion_r1483270839 regarding the config file
I can add that env vars take precedence, but I'm not sure this information should be repeated across multiple commands. And where that matters is in `stackit config list`, where the precedence is solved

joaopalet: I see the point, but I think it could also be relevant to note here that any value set using this command is not guaranteed to take effect (if there is an env var overriding it)

joaopalet: We could also not mention the config file specifically, but I think we should at least mention the possibility that the configuration values set here are being overitten by an env var

hcsa73: Re-added

internal/cmd/config/set/set.go
hcsa73: Added

joaopalet (approved):

PR #77, opened 2024-02-08
Auth, organization and service account fixes

vicentepinto98:
`auth`
- add user-friendly message to failed because of invalid storage key

`organization`
- fix examples

`service-account`
- fix example
- fix messages

### DIFF
--- a/internal/pkg/auth/user_token_flow.go
+++ b/internal/pkg/auth/user_token_flow.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/golang-jwt/jwt/v5"
 	"github.com/spf13/cobra"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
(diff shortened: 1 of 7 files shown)


internal/pkg/auth/user_token_flow.go
```diff
@@ -37,19 +38,19 @@ func UserTokenFlow(cmd *cobra.Command) *userTokenFlow {
 func (utf *userTokenFlow) RoundTrip(req *http.Request) (*http.Response, error) {
 	err := loadVarsFromStorage(utf)
 	if err != nil {
-		return nil, err
+		return nil, &errors.AuthError{}
```
vicentepinto98, line 41: all of these errors will be logged as debug logs once we implement it

vicentepinto98: Will keep it as it was while we dont have debug logs

joaopalet (approved):

PR #92, opened 2024-02-15
Misc fixes to database commands

hcsa73:
- Add hint to `postgresflex/mongodbflex instance update` confirmation prompt that instance might be down
- Add note on usages of `postgresflex/mongodbflex user reset-password` saying password cannot be retrieved later
- Capitalize instance state on `postgresflex instance list` pretty output
- Fix examples descriptions for `logme/mariadb/opensearch/rabbitmq/redis credentials describe`

### DIFF
--- a/internal/cmd/mongodbflex/instance/update/update.go
+++ b/internal/cmd/mongodbflex/instance/update/update.go
@@ -86,7 +86,7 @@ func NewCmd() *cobra.Command {
 			}
 
 			if !model.AssumeYes {
-				prompt := fmt.Sprintf("Are you sure you want to update instance %s?", instanceLabel)
+				prompt := fmt.Sprintf("Are you sure you want to update instance %s? (This may cause downtime)", instanceLabel)
 				err = confirm.PromptForConfirmation(cmd, prompt)
 				if err != nil {
 					return err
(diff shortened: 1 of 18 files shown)


internal/cmd/mongodbflex/instance/update/update.go
```diff
@@ -86,7 +86,7 @@ func NewCmd() *cobra.Command {
 			}
 
 			if !model.AssumeYes {
-				prompt := fmt.Sprintf("Are you sure you want to update instance %s?", instanceLabel)
+				prompt := fmt.Sprintf("Are you sure you want to update instance %s? (This may cause downtime)", instanceLabel)
```
joaopalet, line 89: Should we confirm with the team? Maybe for the MongoDBFlex this is not the case

internal/cmd/postgresflex/instance/list/list.go
```diff
@@ -126,11 +128,12 @@ func outputResult(cmd *cobra.Command, outputFormat string, instances []postgresf
 
 		return nil
 	default:
+		caser := cases.Title(language.English)
 		table := tables.NewTable()
 		table.SetHeader("ID", "NAME", "STATUS")
 		for i := range instances {
 			instance := instances[i]
-			table.AddRow(*instance.Id, *instance.Name, *instance.Status)
+			table.AddRow(*instance.Id, *instance.Name, caser.String(*instance.Status))
```
joaopalet, line 136: Probably could be done is describe command too

internal/cmd/mongodbflex/instance/update/update.go
hcsa73: Removed, will be added later if it applies

joaopalet (approved):

PR #94, opened 2024-02-22
SKE: update flatcar version in tests

robinschneider:
SKE will soon deprecate the old flatcar version, so we should update it before the test will fail.


vicentepinto98 (approved):

vicentepinto98:
Thanks for the contribution :)

PR #97, opened 2024-02-27
Add Object Storage `bucket` commands

joaopalet:


internal/cmd/object-storage/bucket/describe/describe.go
```diff
@@ -0,0 +1,113 @@
...
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("describe %s", bucketNameArg),
+		Short: "Shows details of an Object Storage bucket",
+		Long:  "Shows details of an Object Storage bucket.",
+		Args:  args.SingleArg(bucketNameArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Get details of an Object Storage bucket with name "my-bucket"`,
```
vicentepinto98, line 36: Here we say `Get details` but in the command description we say `Shows details`. Not sure if this is also the case for all of the other commands, but maybe it would be good to uniformize

vicentepinto98 (approved):

PR #105, opened 2024-03-05
Onboard Secrets Manager instance (create and list) and config commands

DiogoFerrao:


internal/cmd/mongodbflex/instance/list/list.go
```diff
@@ -106,7 +106,7 @@ func parseInput(cmd *cobra.Command) (*inputModel, error) {
 
 	return &inputModel{
 		GlobalFlagModel: globalFlags,
-		Limit:           flags.FlagToInt64Pointer(cmd, limitFlag),
+		Limit:           limit,
```
joaopalet, line 109: 👍

internal/cmd/secrets-manager/instance/list/list.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			return fmt.Errorf("marshal Secret Manager instance list: %w", err)
+		}
+		cmd.Println(string(details))
+
+		return nil
+	default:
+		table := tables.NewTable()
+		table.SetHeader("ID", "NAME", "STATE", "SECRETS")
+		// TODO: Add support for Number of Users (to match Portal)
```
joaopalet, line 132: We generally do not leave TODOs in the code, in this case we could either implement this in this (or a follow-up) PR, or create a story for it in case we want to do it at a later point

internal/cmd/secrets-manager/secrets_manager.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "secrets-manager",
+		Short: "Provides functionality for the Secrets Manager",
+		Long:  "Provides functionality for the Secrets Manager.",
```
joaopalet, line 15: I think we can omit the `the`:

```suggestion
		Short: "Provides functionality for Secrets Manager",
		Long:  "Provides functionality for Secrets Manager.",
```

internal/cmd/config/set/set.go
```diff
@@ -102,6 +103,7 @@ func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().String(rabbitMQCustomEndpointFlag, "", "RabbitMQ API base URL, used in calls to this API")
 	cmd.Flags().String(redisCustomEndpointFlag, "", "Redis API base URL, used in calls to this API")
 	cmd.Flags().String(resourceManagerCustomEndpointFlag, "", "Resource Manager API base URL, used in calls to this API")
+	cmd.Flags().String(secretsManagerCustomEndpointFlag, "", "Secrets Managert API base URL, used in calls to this API")
```
vicentepinto98, line 106:
```suggestion
	cmd.Flags().String(secretsManagerCustomEndpointFlag, "", "Secrets Manager API base URL, used in calls to this API")
```

internal/cmd/secrets-manager/instance/list/list.go
```diff
@@ -0,0 +1,144 @@
...
+			apiClient, err := client.ConfigureClient(cmd)
+			if err != nil {
+				return err
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("get Secret Manager instances: %w", err)
```
vicentepinto98, line 64:
```suggestion
				return fmt.Errorf("get Secrets Manager instances: %w", err)
```

internal/cmd/secrets-manager/instance/list/list.go
```diff
@@ -0,0 +1,144 @@
...
+	req := apiClient.ListInstances(ctx, model.ProjectId)
+	return req
+}
+
+func outputResult(cmd *cobra.Command, outputFormat string, instances []secretsmanager.Instance) error {
+	switch outputFormat {
+	case globalflags.JSONOutputFormat:
+		details, err := json.MarshalIndent(instances, "", "  ")
+		if err != nil {
+			return fmt.Errorf("marshal Secret Manager instance list: %w", err)
```
vicentepinto98, line 124:
```suggestion
			return fmt.Errorf("marshal Secrets Manager instance list: %w", err)
```

internal/cmd/secrets-manager/instance/list/list.go
vicentepinto98: It would be good to do in a follow-up PR

vicentepinto98 (dismissed):

vicentepinto98 (approved):

PR #111, opened 2024-03-06
Onboard Secrets Manager user create and list commands

DiogoFerrao:


internal/cmd/secrets-manager/user/create/create.go
```diff
@@ -0,0 +1,140 @@
...
+	Description  *string
+	Write        *bool
+	HidePassword bool
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates a Secrets Manager user",
+		Long:  "Creates a Secrets Manager user.",
```
joaopalet, line 41: Would be nice to add some information here mentioning that the username is always auto-generated, and that a description can be specified to help identifying the user

internal/cmd/secrets-manager/user/create/create.go
```diff
@@ -0,0 +1,140 @@
...
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a Secrets Manager user for instance with ID "xxx"`,
+				"$ stackit mongodbflex user create --instance-id xxx"),
```
joaopalet, line 45: I think we can omit this first example so the list doesn't get too long

joaopalet (approved):

PR #113, opened 2024-03-06
auth: login: use next free port

### LINKED ISSUE #80: Feat: Retry other ports if 8000 is occupied for `stackit auth login`
Hello,
`stackit auth login` needs to open a webUI on localhost:8000. If this port is occupied it fails with:
> Error: authorization failed: listen to port :8000: listen tcp :8000: bind: address already in use

 I now had to disable my development resources disrupting my workflow. A retry loop trying different ports would be convenient.
If you want that as a contribution let me know!

gfelbing:
fixes #80.

This PR tries to address the issue of a fixed redirect port by choosing a random free one.

- [x] Use listener on next free port
- [x] While being complete codewise, the oauth config does not work: When using a different port than 8000, the login results in an error: `Invalid redirect http://localhost:40235 did not match one of the registered values`. This needs to be fixed by maintainers first.

Screenshot:

image

### DIFF
--- a/internal/pkg/auth/user_login.go
+++ b/internal/pkg/auth/user_login.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -23,7 +22,6 @@ import (
 const (
 	authDomain              = "auth.01.idp.eu01.stackit.cloud/oauth"
 	clientId                = "stackit-cli-client-id"
-	redirectURL             = "http://localhost:8000"
 	loginSuccessPath        = "/login-successful"
 	stackitLandingPage      = "https://www.stackit.de"
 	htmlTemplatesPath       = "templates"
@@ -39,6 +37,13 @@ type User struct {
 
 // AuthorizeUser implements the PKCE OAuth2 flow.
 func AuthorizeUser() error {
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return fmt.Errorf("bind port for login redirect: %w", err)
+	}
+	port := listener.Addr().(*net.TCPAddr).Port
+	redirectURL := fmt.Sprintf("http://localhost:%d", port)
+
 	conf := &oauth2.Config{
 		ClientID: clientId,
 		Endpoint: oauth2.Endpoint{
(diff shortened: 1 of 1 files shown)


joaopalet:
Hi @gfelbing,

Thanks for the initiative!

We need to adjust the valid redirect URLs in the identity provider client configuration to allow the use of any local port. We are currently looking into it

joaopalet:
Hi @gfelbing,

The valid redirect URIs were adjusted to allow any port, I just tested this and it worked for me 👍

joaopalet (dismissed):

gfelbing:
@joaopalet thanks! I updated the PR to the latest main and re-tested: works :tada:

joaopalet:
Nice :)

The linter is still complaining as the type assertion must be checked (if it's `ok`)

internal/pkg/auth/user_login.go
```diff
@@ -39,6 +37,13 @@ type User struct {
 
 // AuthorizeUser implements the PKCE OAuth2 flow.
 func AuthorizeUser() error {
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return fmt.Errorf("bind port for login redirect: %w", err)
+	}
+	port := listener.Addr().(*net.TCPAddr).Port
```
joaopalet, line 44:
```suggestion
	port, ok := listener.Addr().(*net.TCPAddr).Port
        if !ok {
		return fmt.Errorf("asserting listener address type to TCP address")
	}
```

internal/pkg/auth/user_login.go
```diff
@@ -39,6 +37,16 @@ type User struct {
...
 func AuthorizeUser() error {
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return fmt.Errorf("bind port for login redirect: %w", err)
+	}
+	port, ok := listener.Addr().(*net.TCPAddr).Port
+        if !ok {
+		return fmt.Errorf("assert listener address type to TCP address")
+	}
+	redirectURL := fmt.Sprintf("http://localhost:%d", port)
```
joaopalet, line 48:
```suggestion
	address, ok := listener.Addr().(*net.TCPAddr)
	if !ok {
		return fmt.Errorf("assert listener address type to TCP address")
	}
	redirectURL := fmt.Sprintf("http://localhost:%d", address.Port)
```

joaopalet: Sorry, I had overlooked it

joaopalet (approved):

PR #125, opened 2024-03-07
Onboard Secrets Manager user delete, describe and update commands

DiogoFerrao:


internal/cmd/secrets-manager/user/describe/describe.go
```diff
@@ -0,0 +1,134 @@
...
+		table := tables.NewTable()
+		table.AddRow("ID", *user.Id)
+		table.AddSeparator()
+		table.AddRow("USERNAME", *user.Username)
+		table.AddSeparator()
+		table.AddRow("DESCRIPTION", *user.Description)
+		if *user.Password != "" {
+			table.AddSeparator()
+			table.AddRow("PASSWORD", *user.Password)
+		}
```
joaopalet, line 115: Even though the field is marked as required, since we already know it might come empty for some reason I think we could play safe here and check first if the field is present

```suggestion
		if user.Password != nil && *user.Password != "" {
			table.AddSeparator()
			table.AddRow("PASSWORD", *user.Password)
		}
```

internal/pkg/services/secrets-manager/utils/utils_test.go
```diff
@@ -32,6 +36,13 @@ func (s *secretsManagerClientMocked) GetInstanceExecute(_ context.Context, _, _
 	return s.getInstanceResp, nil
 }
 
+func (s *secretsManagerClientMocked) GetUserExecute(_ context.Context, _, _, _ string) (*secretsmanager.User, error) {
+	if s.getUserFails {
+		return nil, fmt.Errorf("could not get instance")
```
joaopalet, line 41:
```suggestion
		return nil, fmt.Errorf("could not get user")
```

internal/pkg/services/secrets-manager/utils/utils_test.go
```diff
@@ -79,3 +90,54 @@ func TestGetInstanceName(t *testing.T) {
...
+			description: "base",
+			GetUserResp: &secretsmanager.User{
+				Username:    utils.Ptr(testUserName),
+				Description: utils.Ptr(testDescription),
+			},
+			isValid:        true,
+			expectedOutput: [2]string{testUserName, testDescription},
+		},
+		{
+			description:  "get instance fails",
```
joaopalet, line 112:
```suggestion
			description:  "get user fails",
```

internal/pkg/services/secrets-manager/utils/utils_test.go
```diff
@@ -79,3 +90,54 @@ func TestGetInstanceName(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.description, func(t *testing.T) {
+			client := &secretsManagerClientMocked{
+				getUserFails: tt.getUserFails,
+				getUserResp:  tt.GetUserResp,
+			}
+
+			username, description, err := GetUserDetails(context.Background(), client, testProjectId, testInstanceId, testUserId)
+
+			output := [2]string{username, description}
```
joaopalet, line 127: Maybe we don't need the extra array, we could just define an `expectedUsername` and `expectedDescription` and compare both strings

internal/cmd/secrets-manager/user/update/update.go
```diff
@@ -0,0 +1,135 @@
...
+
+			instanceLabel, err := secretsManagerUtils.GetInstanceName(ctx, apiClient, model.ProjectId, model.InstanceId)
+			if err != nil {
+				instanceLabel = model.InstanceId
+			}
+
+			userLabel, userDescription, err := secretsManagerUtils.GetUserDetails(ctx, apiClient, model.ProjectId, model.InstanceId, model.UserId)
+			if err != nil {
+				userLabel = model.UserId
+			}
```
joaopalet, line 73: The idea of this `...Label` where we recover from the error is that we have a backup in case the API call to get the details fails (instead of making the command return an error for this nice-to-have).

Which means that the `userLabel` should be either `<user name> (<description>)` (nice-to-have) or `<user ID>` (in case the API call fails)

internal/cmd/secrets-manager/user/update/update.go
```diff
@@ -0,0 +1,135 @@
...
+
+			if !model.AssumeYes {
+				prompt := fmt.Sprintf("Are you sure you want to update user %q (%q) of instance %q?", userLabel, userDescription, instanceLabel)
```
joaopalet, line 76: Because currently if the API call above fails we would get `Are you sure you want to update user <user ID> () of instance ...`

internal/cmd/secrets-manager/user/update/update.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Var(flags.UUIDFlag(), instanceIdFlag, "ID of the instance")
+	cmd.Flags().Bool(enableWriteFlag, false, "Set the user to have write access to the secrets engine.")
+	cmd.Flags().Bool(disableWriteFlag, false, "Set the user to have read-only access to the secrets engine.")
```
joaopalet, line 102:
```suggestion
	cmd.Flags().Bool(enableWriteFlag, false, "Set the user to have write access")
	cmd.Flags().Bool(disableWriteFlag, false, "Set the user to have read-only access")
```

internal/cmd/secrets-manager/user/update/update.go
```diff
@@ -0,0 +1,135 @@
...
+
+	err := flags.MarkFlagsRequired(cmd, instanceIdFlag)
+
+	cmd.MarkFlagsMutuallyExclusive(enableWriteFlag, disableWriteFlag)
+	cmd.MarkFlagsOneRequired(enableWriteFlag, disableWriteFlag)
```
joaopalet, line 107: Instead of `MarkFlagsOneRequired` we could check if both are empty in `parseInput` and return an `EmptyUpdateError` in that case (you can check the implementation in other update commands)

internal/cmd/secrets-manager/user/update/update.go
```diff
@@ -0,0 +1,149 @@
...
+				instanceLabel = model.InstanceId
+			}
+
+			var userLabel string
+
+			userName, userDescription, err := secretsManagerUtils.GetUserDetails(ctx, apiClient, model.ProjectId, model.InstanceId, model.UserId)
+			if err != nil {
+				userLabel = fmt.Sprintf("%q", model.UserId)
+			} else {
+				userLabel = fmt.Sprintf("%q (%q)", userName, userDescription)
```
joaopalet, line 76: Probably the description doesn't need to be quoted, right? Not sure

joaopalet (approved):

PR #133, opened 2024-03-11
Onboard Argus (instance): create command and utils

DiogoFerrao:


internal/cmd/argus/instance/create/create.go
```diff
@@ -0,0 +1,191 @@
...
+	if planId == nil && (planName == "") {
+		return nil, &cliErr.DSAInputPlanError{
+			Cmd: cmd,
+		}
+	}
+	if planId != nil && (planName != "") {
+		return nil, &cliErr.DSAInputPlanError{
+			Cmd: cmd,
+		}
+	}
```
joaopalet, line 143: In here we should create a new error type `ArgusInputPlanError`

`DSAInputPlanError` is used for DSA services (`redis`, `mariadb`, `logme`, `rabbitmq`, and `opensearch`) where the plan is given as a combination of plan name and version. You can still take this implementation as reference

internal/cmd/argus/instance/create/create.go
```diff
@@ -0,0 +1,191 @@
...
+		return req, fmt.Errorf("get Argus plans: %w", err)
+	}
+
+	if model.PlanId == nil {
+		planId, err = argusUtils.LoadPlanId(model.PlanName, plans)
+		if err != nil {
+			var dsaInvalidPlanError *cliErr.DSAInvalidPlanError
+			if !errors.As(err, &dsaInvalidPlanError) {
+				return req, fmt.Errorf("load plan ID: %w", err)
+			}
```
joaopalet, line 175: Same here, we should create a `ArgusInvalidPlanError` for when the plan ID cannot be found for the given plan name

internal/pkg/services/argus/utils/utils.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/stackitcloud/stackit-sdk-go/services/argus"
+)
+
+const (
+	service = "argus"
+)
+
+func ValidatePlanId(planId string, resp *argus.PlansResponse) error {
+	if resp == nil {
+		// Should not happen, check is done before calling this function
```
joaopalet, line 19: I think we don't need this comment here

internal/pkg/services/argus/utils/utils.go
```diff
@@ -0,0 +1,71 @@
...
+	return &errors.DSAInvalidPlanError{
+		Service: service,
+		Details: fmt.Sprintf("You provided plan ID %q, which is invalid.", planId),
+	}
+}
+
+func LoadPlanId(planName string, resp *argus.PlansResponse) (*string, error) {
+	availablePlanNames := ""
+	if resp == nil {
+		// Should not happen, check is done before calling this function
```
joaopalet, line 39: Same here

internal/pkg/services/argus/utils/utils.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	for i := range *resp.Plans {
+		plan := (*resp.Plans)[i]
+		if plan.Id != nil && strings.EqualFold(*plan.Id, planId) {
+			return nil
+		}
+	}
+
+	return &errors.DSAInvalidPlanError{
```
joaopalet, line 30: Shoud use the new `ArgusInvalidPlanError`

internal/pkg/services/argus/utils/utils.go
```diff
@@ -0,0 +1,71 @@
...
+		if plan.Name == nil {
+			continue
+		}
+		if strings.EqualFold(*plan.Name, planName) && plan.Id != nil {
+			return plan.Id, nil
+		}
+		availablePlanNames = fmt.Sprintf("%s\n- %s", availablePlanNames, *plan.Name)
+	}
+
+	details := fmt.Sprintf("You provided plan_name %q, which is invalid. Available plan names are: %s", planName, availablePlanNames)
```
joaopalet, line 54:
```suggestion
	details := fmt.Sprintf("You provided plan name %q, which is invalid. Available plan names are: %s", planName, availablePlanNames)
```

internal/pkg/services/argus/utils/utils.go
```diff
@@ -0,0 +1,71 @@
...
+	return nil, &errors.DSAInvalidPlanError{
```
joaopalet, line 55: `ArgusInvalidPlanError`

internal/pkg/services/argus/utils/utils.go
```diff
@@ -0,0 +1,71 @@
...
+		// Should not happen, check is done before calling this function
+		return nil, fmt.Errorf("no Argus plans provided")
+	}
+
+	for i := range *resp.Plans {
+		plan := (*resp.Plans)[i]
+		if plan.Name == nil {
+			continue
+		}
+		if strings.EqualFold(*plan.Name, planName) && plan.Id != nil {
```
joaopalet, line 48: This is already checked in the condition above

internal/cmd/argus/instance/create/create_test.go
```diff
@@ -0,0 +1,315 @@
...
+			expectedRequest: fixtureRequest(),
+			getPlansReponse: &argus.PlansResponse{
+				Plans: &[]argus.Plan{
+					{
+						Name: utils.Ptr("example-plan-name"),
+						Id:   utils.Ptr(testPlanId),
+					},
+				},
+			},
+		},
```
joaopalet, line 241: These test cases do not `isValid` defined (it defaults to false). Are they suppose to be invalid? Either way I think it's better to explicitly set the field

internal/cmd/argus/instance/create/create_test.go
```diff
@@ -0,0 +1,315 @@
...
+			isValid: false,
+		},
+		{
+			description: "required fields only",
+			model: &inputModel{
+				GlobalFlagModel: &globalflags.GlobalFlagModel{
+					ProjectId: testProjectId,
+				},
+				PlanId: utils.Ptr(testPlanId),
+			},
```
joaopalet, line 278: Isn't this the same as the `fixtureInputModel`?

internal/pkg/services/argus/utils/utils.go
DiogoFerrao: I think the conditions are different. Maybe there are too many checks, but the one above makes sure that plan name is not null so that the comparison is safe, then the second one checks if the id is not null so that we can return it. I used the structure from mariadb.

joaopalet: Ah right, I misread it

internal/pkg/services/argus/utils/utils.go
```diff
@@ -51,8 +49,8 @@ func LoadPlanId(planName string, resp *argus.PlansResponse) (*string, error) {
 		availablePlanNames = fmt.Sprintf("%s\n- %s", availablePlanNames, *plan.Name)
 	}
 
-	details := fmt.Sprintf("You provided plan_name %q, which is invalid. Available plan names are: %s", planName, availablePlanNames)
-	return nil, &errors.DSAInvalidPlanError{
+	details := fmt.Sprintf("you provided plan name %q, which is invalid. available plan names are: %s", planName, availablePlanNames)
```
joaopalet, line 52: This can remain capitalized, it will be shown in a new line after the `Error: ...` message

joaopalet (approved):

PR #149, opened 2024-03-12
Onboard Argus (instance): update command

DiogoFerrao:


internal/cmd/argus/instance/update/update.go
```diff
@@ -0,0 +1,217 @@
...
+		PlanId:          planId,
+		PlanName:        planName,
+		InstanceName:    instanceName,
+	}, nil
+}
+
+type argusClient interface {
+	UpdateInstance(ctx context.Context, instanceId, projectId string) argus.ApiUpdateInstanceRequest
+	ListPlansExecute(ctx context.Context, projectId string) (*argus.PlansResponse, error)
+	GetInstanceExecute(ctx context.Context, instanceId, projectId string) (*argus.GetInstanceResponse, error)
```
joaopalet, line 167: Currently this is not used, right? But might be if we decide to call this directly instead of `GetInstanceDetails`

internal/cmd/argus/instance/update/update_test.go
```diff
@@ -0,0 +1,451 @@
...
+			model: fixtureInputModel(
+				func(model *inputModel) {
+					model.InstanceName = nil
+				},
+			),
+			getPlansResponse:    fixturePlansResponse(),
+			getInstanceResponse: fixtureGetInstanceResponse(),
+			expectedRequest:     fixtureRequest(),
+			isValid:             true,
+		},
```
joaopalet, line 376: In these tests we should test updating the plan to a new plan, to test that the built request contains the `NewPlanId` and not the current `TestPlanId`

Currently it is only testing that the request contains *a* plan ID (which should always be true)

internal/cmd/argus/instance/update/update.go
```diff
@@ -0,0 +1,215 @@
...
+		return nil, &cliErr.ArgusInputPlanError{
+			Cmd: cmd,
+		}
+	}
+
+	if planId == nil && planName == "" {
+		if instanceName == nil {
+			return nil, &cliErr.EmptyUpdateError{}
+		}
+	}
```
joaopalet, line 151: Can now be simplified to 
```suggestion
	if planId == nil && planName == "" && instanceName == nil {
        return nil, &cliErr.EmptyUpdateError{}
	}
```

joaopalet (approved):

joaopalet (approved):

PR #150, opened 2024-03-12
Onboard PostgreSQL Flex instance clone command

GokceGK:
Following commands have been added:

instance 
* clone


internal/cmd/postgresflex/backups/backups.go
joaopalet, line 1: According to the story, this command should be a child of `instance`, right?

`$ stackit mongodbflex instance backup {...}`

internal/cmd/postgresflex/backups/update-schedule/update_schedule.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/examples"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/flags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/services/postgresflex/client"
+	postgresflexUtils "github.com/stackitcloud/stackit-cli/internal/pkg/services/postgresflex/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/postgresflex"
+)
+
+const (
+	instanceIdArg = "INSTANCE_ID"
```
joaopalet, line 22: Same as omment above regarding argument, in this case it should also be a flag

internal/cmd/postgresflex/backups/backups.go
joaopalet: Not saying it's the best option, just wanted to confirm it was intentional. We could discuss it in a round

internal/cmd/postgresflex/backups/update-schedule/update_schedule.go
```diff
@@ -0,0 +1,116 @@
...
+			cmd.Printf("Updated backup schedule of instance %q\n", instanceLabel)
+			return nil
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().String(backupScheduleFlag, "", "Backup schedule")
```
joaopalet, line 89: We could add more info here regarding the expected format, e.g. comething like `Should follow the cron scheduling system format (e.g. "0 0 * * *")`

internal/cmd/postgresflex/backups/update-schedule/update_schedule_test.go
```diff
@@ -0,0 +1,257 @@
...
+			flagValues:  map[string]string{},
+			isValid:     false,
+		},
+		{
+			description:   "with defaults",
+			argValues:     fixtureArgValues(),
+			flagValues:    fixtureFlagValues(),
+			isValid:       true,
+			expectedModel: fixtureInputModel(),
+		},
```
joaopalet, line 120: This test case is repeated (same as `base`)

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("clone %s", instanceIdArg),
+		Short: "Clones a PostgreSQL Flex instance",
+		Long:  "Clones a PostgreSQL Flex instance from a selected point in time.",
+		Example: examples.Build(
+			examples.NewExample(
+				`Clone a PostgreSQL Flex instance with ID "xxx" . The recovery timestamp should be specified in UTC time following the format provided in the example.`,
```
joaopalet, line 48: This is nice info but I think it should be specified in the flag description instead

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -0,0 +1,154 @@
...
+				`$ stackit postgresflex instance clone xxx --recovery-timestamp 2023-04-17T09:28:00+00:00`),
+			examples.NewExample(
+				`Clone a PostgreSQL Flex instance with ID "xxx" from a selected recovery timestamp and specify storage class. If not specified, storage class from the existing instance will be used.`,
```
joaopalet, line 51: Same here

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -0,0 +1,154 @@
...
+				`$ stackit postgresflex instance clone xxx --recovery-timestamp 2023-04-17T09:28:00+00:00 --storage-class premium-perf6-stackit`),
+			examples.NewExample(
+				`Clone a PostgreSQL Flex instance with ID "xxx" from a selected recovery timestamp and specify storage size. If not specified, storage size from the existing instance will be used.`,
```
joaopalet, line 54: And here

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -0,0 +1,154 @@
...
+	StorageClass *string
+	StorageSize  *int64
+	RecoveryDate *string
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("clone %s", instanceIdArg),
+		Short: "Clones a PostgreSQL Flex instance",
+		Long:  "Clones a PostgreSQL Flex instance from a selected point in time.",
```
joaopalet, line 45: I think here would also be a good place to eleborate a bit more, saying the clone will be an independent instance with the same settings as the original instance unless the flags are specified etc

joaopalet: What do you think?

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -0,0 +1,154 @@
...
+			cmd.Printf("%s instance from instance %q. New Instance ID: %s\n", operationState, instanceLabel, instanceId)
+			return nil
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().String(recoveryTimestampFlag, "", "Recovery timestamp for the instance, in a date-time with the layout format, e.g. 2024-03-12T09:28:00+00:00")
```
joaopalet, line 121: Is `layout format` the known name for this? Or maybe I misendurstood this sentence

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/stackitcloud/stackit-sdk-go/services/postgresflex/wait"
+)
+
+const (
+	instanceIdArg = "INSTANCE_ID"
+
+	storageClassFlag      = "storage-class"
+	storageSizeFlag       = "storage-size"
+	recoveryTimestampFlag = "recovery-timestamp"
+	recoveryDateFormat    = "2023-04-17T09:28:00+00:00"
```
joaopalet, line 29: This is unused, right? Where is the timestamp format being validated?

internal/cmd/postgresflex/instance/clone/clone_test.go
```diff
@@ -0,0 +1,325 @@
...
+				flagValues[recoveryTimestampFlag] = ""
+			}),
+			isValid: false,
+		},
+		{
+			description: "recovery timestamp is invalid",
+			flagValues: fixtureRequiredFlagValues(func(flagValues map[string]string) {
+				flagValues[recoveryTimestampFlag] = "test"
+			}),
+			isValid: false,
```
joaopalet, line 235: Because this test seems to be passing, meaning the invalid timestamp is catched somewhere right?

internal/cmd/postgresflex/instance/clone/clone_test.go
```diff
@@ -0,0 +1,325 @@
...
+			}
+			diff := cmp.Diff(model, tt.expectedModel)
+			if diff != "" {
+				t.Fatalf("Data does not match: %s", diff)
+			}
+		})
+	}
+}
+
+func TestBuildRequest(t *testing.T) {
```
joaopalet, line 299: This one should have a couple more test cases, testing `nil` storage class and size

internal/cmd/postgresflex/instance/clone/clone_test.go
GokceGK: timestamp validation has been added

internal/cmd/postgresflex/instance/clone/clone.go
GokceGK: timestamp validation has been added

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -0,0 +1,190 @@
...
+
+		storages, err = apiClient.ListStoragesExecute(ctx, model.ProjectId, *validationFlavorId)
+		if err != nil {
+			return req, fmt.Errorf("get PostgreSQL Flex storages: %w", err)
+		}
+		err = postgresflexUtils.ValidateStorage(model.StorageClass, model.StorageSize, storages, *validationFlavorId)
+		if err != nil {
+			return req, err
+		}
+	}
```
joaopalet, line 182: In here, it should be possible for the user to only specify the size (or only the class), right?

In the case the user only wants to change the storage class e.g., we should validate that the current instance storage size together with the new storage class are valid

joaopalet: What you are doing currently is always taking `model.StorageClass` and `model.StorageSize` to `ValidateStorage`, even if one is null. Meaning this call will always fail in that case

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/stackitcloud/stackit-cli
 go 1.21
 
 require (
+	github.com/depp/bytesize v1.1.0
```
joaopalet, line 6: This is not used, right?

joaopalet: You can run `go mod tidy` in the repo root to clear unused dependencies from the `go.mod` file

internal/cmd/postgresflex/instance/clone/clone.go
GokceGK: updated validation and adapted unit test

joaopalet (approved):

PR #155, opened 2024-03-13
Onboard Argus (instance): list and describe commands

DiogoFerrao:


internal/cmd/argus/instance/list/list.go
```diff
@@ -0,0 +1,142 @@
...
+		details, err := json.MarshalIndent(instances, "", "  ")
+		if err != nil {
+			return fmt.Errorf("marshal Argus instance list: %w", err)
+		}
+		cmd.Println(string(details))
+
+		return nil
+	default:
+		table := tables.NewTable()
+		table.SetHeader("ID", "PLAN", "NAME", "STATUS")
```
joaopalet, line 130: Maybe swap the order of plan with name? Or did you take this from some other instance that has it like this?

joaopalet (commented):
Remember to also update the docs

joaopalet (approved):

PR #157, opened 2024-03-15
Add SKE login command

Kumm-Kai:
This PR adds a new command (`login`) below the `ske` `credentials` command group.
This command can only be used together with `kubectl` and a "special" kubeconfig.

To test this PR against prod you can adjust an existing kubeconfig so that it looks like this:
<details>

```yaml
kind: Config
apiVersion: v1
preferences: {}
clusters:
  - name: NAME
    cluster:
      server: URL
      certificate-authority-data: CACERT
      extensions:
        - name: client.authentication.k8s.io/exec
          extension:
            clusterName: NAME
            stackitProjectID: UUID
users:
  - name: NAME
    user:
      exec:
        command: stackit
        args:
          - ske
          - credentials
          - kubeconfig
          - login
        apiVersion: client.authentication.k8s.io/v1
        provideClusterInfo: true
        interactiveMode: Never
contexts:
  - name: NAME
    context:
      cluster: NAME
      user: NAME
current-context: NAME
```
</details>

Additionally, we will soon add a new endpoint that generates a kubeconfig for use with the login command.

CC: @dergeberl


vicentepinto98 (commented):
Currently there are two use cases related to `kubeconfig`. One is the login for `kubectl` and the other one is creating a `kubeconfig` (which we will introduce soon)

When we had the discussion offline to have the command under the `credentials` command group, I thought we could name the commands something like `kubectl-login` and then `get-kubeconfig`. This would be so that we don't introduce another command sub-group. 

For reference, the guideline for CLI commands is:
`stackit <GROUP> <SUB-GROUP> <COMMAND> <ARGUMENT> <PARAMETER FLAGS> [OPTION FLAGS]`

So we should to go with this approach:

`stackit ske kubeconfig login` and `stackit ske kubeconfig create` (I prefer this one)

internal/cmd/ske/credentials/kubeconfig/login/login.go
```diff
@@ -0,0 +1,223 @@
...
+	ProjectId   string
+	ClusterName string
+	CacheKey    string
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "login",
+		Short: "login plugin for kubectl",
+		Long:  "login plugin for kubectl to create a short-lived kubeconfig to authenticate against a STACKIT Kubernetes Engine (SKE) cluster. To get a kubeconfig to use with the login command use the 'kubeconfig create' command",
```
vicentepinto98, line 38: > To get a kubeconfig to use with the login command use the 'kubeconfig create' command

I don't get this sentence, is that something that the user needs to do or something that is done in the background in this command?

The `kubeconfig create` command is not part of the CLI yet

internal/cmd/ske/credentials/kubeconfig/login/login.go
```diff
@@ -0,0 +1,223 @@
...
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				"login to a SKE cluster specified in the kubeconfig",
+				"$ kubectl get pod"),
```
vicentepinto98, line 43: The examples should include the usage of the CLI command. If there are other specific follow-up commands to other CLIs that are relevant for the example, such as kubectl, it's fine to have it but I would expect the CLI command to also be present

So this one at least seems incomplete

internal/cmd/ske/credentials/kubeconfig/login/login.go
```diff
@@ -0,0 +1,223 @@
...
+	return &inputModel{
+		ClusterName: config.ClusterName,
+		ProjectId:   config.STACKITProjectID,
+		CacheKey:    fmt.Sprintf("ske-login-%x", sha256.Sum256([]byte(execCredential.Spec.Cluster.Server))),
+	}, nil
+}
+
+func buildRequest(ctx context.Context, apiClient *ske.APIClient, model *inputModel) ske.ApiCreateKubeconfigRequest {
+	req := apiClient.CreateKubeconfig(ctx, model.ProjectId, model.ClusterName)
+	expirationSeconds := "1800" // 30 min
```
vicentepinto98, line 132: you could extract this to a const instead of hardcoding, for better maitainability

internal/cmd/ske/credentials/kubeconfig/login/login.go
```diff
@@ -0,0 +1,223 @@
...
+			ClientCertificateData: string(kubeconfig.CertData),
+			ClientKeyData:         string(kubeconfig.KeyData),
+		},
+	}
+	return &outputExecCredential, nil
+}
+
+func getCachedKubeConfig(key string) *rest.Config {
+	cachedKubeconfig, err := cache.GetObject(key)
+	if err != nil {
```
vicentepinto98, line 164: we will introduce debug logs soon, at that point we should log these errors

internal/cmd/ske/credentials/kubeconfig/login/login.go
```diff
@@ -0,0 +1,223 @@
...
+
+	return restConfig
+}
+
+type SKEClusterConfig struct {
+	STACKITProjectID string `json:"stackitProjectId"`
+	ClusterName      string `json:"clusterName"`
+}
+
+func getCacheAndOutputKubeconfig(ctx context.Context, cmd *cobra.Command, apiClient *ske.APIClient, model *inputModel, refresh bool, oldKubeconfig *rest.Config) error {
```
vicentepinto98, line 181: I find the argument name `refresh` confusing. Since actually when `refresh` is true, you are returning the old kube config. 
So I would say it should be rather something like `getCachedKubeconfig`

internal/pkg/cache/cache.go
```diff
@@ -0,0 +1,86 @@
+package cache
```
vicentepinto98, line 1: unit tests should be added to the cache pkg

internal/cmd/ske/credentials/kubeconfig/login/login.go
vicentepinto98: The name of the method is also misleading because from the first look I thought we are getting the cached kubeconfig, but actually we are getting, then caching, then output

internal/cmd/ske/credentials/kubeconfig/login/login.go
```diff
@@ -0,0 +1,223 @@
...
+			if err != nil {
+				_ = cache.DeleteObject(model.CacheKey)
+				return getCacheAndOutputKubeconfig(ctx, cmd, apiClient, model, false, nil)
+			}
+
+			if time.Now().After(certificate.NotAfter.UTC()) {
+				// cert expired, request new
+				_ = cache.DeleteObject(model.CacheKey)
+				return getCacheAndOutputKubeconfig(ctx, cmd, apiClient, model, false, nil)
+			} else if time.Now().Add(time.Minute * 15).After(certificate.NotAfter.UTC()) {
```
vicentepinto98, line 81: could also be moved to a const

internal/cmd/ske/credentials/kubeconfig/login/login.go
Kumm-Kai: On SKE side, we will add a new endpoint, which will be used to create these "login kubeconfigs". That's something the user would need to do once, afterward kubectl will call `stackit ske kubeconfig login` directly to get new short-lived credentials for the cluster.

internal/cmd/ske/credentials/kubeconfig/login/login.go
Kumm-Kai: Then I think removing the example would be better, because this command can only be used together with the "login kubeconfig" and kubectl.

internal/cmd/ske/credentials/kubeconfig/login/login.go
Kumm-Kai: I've renamed the function and the `refresh` parameter. Let me know if it's better now.

internal/cmd/ske/credentials/kubeconfig/login/login.go
```diff
@@ -0,0 +1,223 @@
...
+	req := apiClient.CreateKubeconfig(ctx, model.ProjectId, model.ClusterName)
+	expirationSeconds := "1800" // 30 min
+
+	return req.CreateKubeconfigPayload(ske.CreateKubeconfigPayload{ExpirationSeconds: &expirationSeconds})
+}
+
+func parseKubeConfigToExecCredential(kubeconfig *rest.Config) (*clientauthenticationv1.ExecCredential, error) {
+	certPem, _ := pem.Decode(kubeconfig.CertData)
+	if certPem == nil {
+		return nil, fmt.Errorf("login SKE kubeconfig")
```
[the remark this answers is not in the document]
Kumm-Kai: Should be better now :)

internal/cmd/ske/credentials/kubeconfig/login/login.go
Kumm-Kai: We could add them, but because this command will only be invoked by `kubectl` users won't ever see them. In this case, a log file would be the better approach,

internal/cmd/ske/credentials/kubeconfig/login/login.go
Kumm-Kai: `ske kubeconfig create` command extended to allow creation of "login kubeconfigs"

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+const (
+	expirationSeconds     = 30 * 60          // 30 min
+	refreshBeforeDuration = 15 * time.Minute // 15 min
+)
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "login",
```
vicentepinto98, line 37:
```suggestion
		Use:   "login",
		Hidden: true,
```

Since this command is only to be used inside the "special" kubeconfig, I suggest that we hide it from the CLI documentation.

We are planning to introduce CLI plugins in the near future, when we have that in place we could move this command there, it would be more appropriate

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+		Short: "Login plugin for kubernetes clients",
+		Long: fmt.Sprintf("%s\n%s",
+			"Login plugin for kubernetes clients, that creates short-lived credentials to authenticate against a STACKIT Kubernetes Engine (SKE) cluster.",
```
vicentepinto98, line 40: Can you extend this description with the complete flow when using these two commands together and the use case with `kubectl`.

It might not be clear for most users

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+			"To obtain a kubeconfig for use with the login command, run 'kubeconfig create --login'.",
+		),
+		Args: args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Get a login kubeconfig for the SKE cluster with name "my-cluster". `+
+					"This kubeconfig does not contain any credentials and instead obtains valid credentials via the STACKIT CLI.",
+				"$ stackit ske kubeconfig create my-cluster --login"),
+			examples.NewExample(
```
vicentepinto98, line 49: these two examples should be merged, they are part of the same flow

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+			if err := output(cmd, model.CacheKey, cachedKubeconfig); err != nil {
+				return err
+			}
+			return nil
+		},
+	}
+	return cmd
+}
+
+type inputModel struct {
```
vicentepinto98, line 106: I think this `inputModel` is not needed and it doesnt match the ones that we have for the other commands as this one is not storing any user input. So we can just have the `SKEClusterConfig` struct with the `CacheKey` also defined there

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+	ClusterName string
+	CacheKey    string
+}
+
+type SKEClusterConfig struct {
+	STACKITProjectID string `json:"stackitProjectId"`
+	ClusterName      string `json:"clusterName"`
+}
+
+func parseInput() (*inputModel, error) {
```
vicentepinto98, line 117:
```suggestion
func parseSKEClusterConfig() (*SKEClusterConfig, error) {
```

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+	return cmd
+}
+
+type inputModel struct {
+	ProjectId   string
+	ClusterName string
+	CacheKey    string
+}
+
+type SKEClusterConfig struct {
```
vicentepinto98, line 112: this type doesn't need to be public right?

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+		if fallbackToCache {
+			return output(cmd, model.CacheKey, cachedKubeconfig)
+		}
+		return fmt.Errorf("cache kubeconfig: %w", err)
+	}
+
+	return output(cmd, model.CacheKey, kubeconfig)
+}
+
+func buildRequest(ctx context.Context, apiClient *ske.APIClient, model *inputModel) ske.ApiCreateKubeconfigRequest {
```
vicentepinto98, line 193: consider adding unit test for this function

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		_ = cache.DeleteObject(cacheKey)
+		return fmt.Errorf("marshal ExecCredential: %w", err)
+	}
+
+	cmd.Print(string(output))
+	return nil
+}
+
+func parseKubeConfigToExecCredential(kubeconfig *rest.Config) (*clientauthenticationv1.ExecCredential, error) {
```
vicentepinto98, line 217: consider adding unit test for this function

internal/pkg/cache/cache.go
```diff
@@ -0,0 +1,62 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"regexp"
+
+	"github.com/adrg/xdg"
+)
+
+var (
+	cacheFolderPath = xdg.CacheHome + "/stackit"
```
vicentepinto98, line 13: you can use the std library: https://pkg.go.dev/os#UserCacheDir
then we can remove the `xdg` dependency

internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -48,6 +50,10 @@ func NewCmd(p *print.Printer) *cobra.Command {
 			examples.NewExample(
 				`Create a kubeconfig for the SKE cluster with name "my-cluster"`,
 				"$ stackit ske kubeconfig create my-cluster"),
+			examples.NewExample(
+				`Get a login kubeconfig for the SKE cluster with name "my-cluster". `+
+					"This kubeconfig does not contain any credentials and instead obtains valid credentials via the STACKIT CLI.",
```
vicentepinto98, line 55:
```suggestion
					"This kubeconfig does not contain any credentials and instead obtains valid credentials via the `stackit ske kubeconfig login` command.",
```

internal/cmd/ske/kubeconfig/create/create_test.go
```diff
@@ -202,7 +212,7 @@ func TestParseInput(t *testing.T) {
 	}
 }
 
-func TestBuildRequest(t *testing.T) {
+func TestBuildRequestCreate(t *testing.T) {
```
vicentepinto98, line 215: please add unit test for `BuildRequestLogin`

internal/cmd/ske/credentials/kubeconfig/login/login.go
vicentepinto98: The log file is something that we could have across the CLI, so it is not specific for this command. We can add it later when that is implemented

internal/cmd/ske/credentials/kubeconfig/login/login.go
vicentepinto98: It's much nicer now :)

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -0,0 +1,240 @@
...
+		return fmt.Errorf("convert to ExecCredential: %w", err)
+	}
+
+	output, err := json.Marshal(outputExecCredential)
+	if err != nil {
+		_ = cache.DeleteObject(cacheKey)
+		return fmt.Errorf("marshal ExecCredential: %w", err)
+	}
+
+	cmd.Print(string(output))
```
vicentepinto98, line 213:
```suggestion
	p.Outputf(string(output), nil)
```

`p` is `p *print.Printer` which is a command argument

vicentepinto98:
Run `make generate-docs`

internal/cmd/ske/kubeconfig/create/create_test.go
Kumm-Kai: `buildRequestLogin` only wraps the `apiClient.GetLoginKubeconfig` func.
I can remove the `buildRequestLogin` and use `apiClient.GetLoginKubeconfig` if you want.

internal/cmd/ske/kubeconfig/login/login.go
Kumm-Kai: > // Print an output using Printf to the defined output (falling back to Stderr if not set).
> // If output format is set to none, it does nothing

Where does the output come from? I need stdout not stderr.

internal/cmd/ske/kubeconfig/login/login.go
Kumm-Kai: How can I merge them? There is only one flow in general. I don't think that separate examples mean different flows...

internal/cmd/ske/kubeconfig/login/login.go
Kumm-Kai: As discussed: we won't hide this command to allow easier discovery of the login feature.

internal/cmd/ske/kubeconfig/login/login.go
Kumm-Kai: What should I test there? The function basically only puts part of the kubeconfig into the expected struct.

internal/cmd/ske/kubeconfig/login/login.go
Kumm-Kai: The function contains no logic only assignments

internal/cmd/ske/kubeconfig/login/login.go
Kumm-Kai: Is it better now?

internal/cmd/ske/kubeconfig/login/login.go
vicentepinto98: Yeah in that case it's ok like this. I meant that usually we have different examples for different use cases but here they are the same

internal/cmd/ske/kubeconfig/login/login.go
vicentepinto98: Yes, looks good

internal/cmd/ske/kubeconfig/login/login.go
vicentepinto98: You can test that the returned struct is like expected

internal/cmd/ske/kubeconfig/login/login.go
vicentepinto98: `p.Outputf` is just a wrapper around `cmd.Printf` that is part of our `print` pkg that is used for all command output.

It prints to the defined output of the cobra command, which is `stdout`

The only difference is that it takes into account the configured output format on the CLI.

internal/pkg/cache/cache.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+var (
+	cacheFolderPath string
+
+	identifierRegex             = regexp.MustCompile("^[a-zA-Z0-9-]+$")
+	ErrorInvalidCacheIdentifier = fmt.Errorf("invalid cache identifier")
+)
+
+func init() {
```
vicentepinto98, line 18: Can you move this to a `Init` function and call it on the `login` command? Then you can pass the `cacheFolderPath` as argument to the other functions

internal/cmd/ske/kubeconfig/login/login.go
vicentepinto98: We have a unit test for this `buildRequest` function in all CLI commands, it's quite easy to write and in this case can test that the request includes a payload with the `expirationSeconds`.

For uniformity with the rest of the codebase, it should be added

internal/cmd/ske/kubeconfig/create/create_test.go
vicentepinto98: Again, mainly for uniformity with the rest of the codebase as we usually always test these `buildRequest` methods. But in this case it's really just a wrapper so you can skip this test if you want

internal/cmd/ske/kubeconfig/login/login.go
Kumm-Kai: Done! If I got it correctly, the configured output format would only apply if I would add something like:
```go
  switch model.OutputFormat {
  case print.JSONOutputFormat:
```
correct? Just want to be 100% sure that users aren't able to mess with the output format here.

vicentepinto98: Yes, that's correct for other output formats such as `json` or `yaml`
However, if the users pass the `--output-format` to `none`, no output would be shown, because that is part of our global verbosity handling. 

I think this is not a problem, it wouldn't make sense to pass that option when using this particular command

PR #163, opened 2024-03-18
Onboard Secrets Manager (ACLs): create command

DiogoFerrao:


internal/cmd/secrets-manager/instance/create/create.go
```diff
@@ -65,14 +71,25 @@ func NewCmd() *cobra.Command {
...
-			req := buildRequest(ctx, model, apiClient)
+			// Call API to create instance
+			req := buildCreateInstanceRequest(ctx, model, apiClient)
 			resp, err := req.Execute()
 			if err != nil {
 				return fmt.Errorf("create Secrets Manager instance: %w", err)
 			}
 			instanceId := *resp.Id
 
+			// Call API to create ACls for instance, if ACL is provided
```
joaopalet, line 82:
```suggestion
			// Call API to create ACLs for instance, if ACL is provided
```

internal/cmd/secrets-manager/instance/create/create.go
```diff
@@ -97,10 +115,26 @@ func parseInput(cmd *cobra.Command) (*inputModel, error) {
...
+		req := apiClient.CreateACL(ctx, model.ProjectId, instanceId)
+
+		req = req.CreateACLPayload(secretsmanager.CreateACLPayload{
+			Cidr: utils.Ptr(acl),
+		})
+		requests = append(requests, req)
+	}
+
+	return requests
+}
```
joaopalet, line 135: Is it possible to use the PUT endpoint to create all ACLs at once?

joaopalet: This would save us from making one request per ACL

internal/cmd/secrets-manager/instance/create/create.go
```diff
@@ -65,14 +71,25 @@ func NewCmd() *cobra.Command {
...
 
+			// Call API to create ACls for instance, if ACL is provided
+			if model.Acls != nil && len(*model.Acls) > 0 {
+				requests := buildCreateACLRequests(ctx, model, instanceId, apiClient)
+				for _, req := range requests {
+					_, err := req.Execute()
+					if err != nil {
+						return fmt.Errorf("create ACL for Secrets Manager instance: %w", err)
+					}
+				}
```
joaopalet, line 90: And would simplify a bit the error handling here

joaopalet: As it is, it might not be entirely clear to the user in which state he ends up if e.g. the second call to create an ACL fails

joaopalet: Nevertheless, we should still signal the user that the instance was created even if the call to create/update the ACLs failed.

The output could something like:
```shell
Error: create ACLs for Secrets Manager instance.

The creation of the ACLs for the Secrets Manager instance failed, but the instance was correctly created.
If you want to retry configuring the ACLs, you can do it via:
  $ stackit secrets-manager instance update (...)
```

joaopalet: Alternatively, we could delete the newly created instance. What do you think @vicentepinto98?

internal/cmd/secrets-manager/instance/create/create.go
```diff
@@ -109,3 +129,17 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *secretsmana
 
 	return req
 }
+
+func buildUpdateACLsRequest(ctx context.Context, model *inputModel, instanceId string, apiClient *secretsmanager.APIClient) secretsmanager.ApiUpdateACLsRequest {
+	req := apiClient.UpdateACLs(ctx, model.ProjectId, instanceId)
+
+	cidrs := []secretsmanager.AclUpdate{}
```
joaopalet, line 136: *Small nitpick:* In cases where we already know the size of the slice, we can use `make([]secretsmanager.AclUpdate{}, LENGTH)` to define the length instead of allocating more memory each time we want to append an item

In this case we would then do `cidrs[i] = secretsmanager.AclUpdate{Cidr: utils.Ptr(acl)}` to set each element

joaopalet (dismissed):

joaopalet (approved):

PR #167, opened 2024-03-19
Onboard Secrets Manager (ACL): update command

DiogoFerrao:


internal/cmd/secrets-manager/instance/update/update.go
```diff
@@ -0,0 +1,124 @@
...
+			cmd.Printf("Updated instance %q\n", instanceLabel)
+			return nil
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Var(flags.CIDRSliceFlag(), aclFlag, "List of IP networks in CIDR notation which are allowed to access this instance")
```
vicentepinto98, line 88: I was considering if it would make sense to even mark the --acl flag as required since it is the only one, but for uniformity with the other commands, your approach is good

internal/cmd/secrets-manager/instance/update/update_test.go
```diff
@@ -0,0 +1,304 @@
...
+			isValid:     false,
+		},
+		{
+			description: "required flags only (no values to update)",
+			argValues:   fixtureArgValues(),
+			flagValues: map[string]string{
+				projectIdFlag: testProjectId,
+			},
+			isValid: false,
+			expectedModel: &inputModel{
```
vicentepinto98, line 122: remove expected model for invalid test cases, makes it more readable

internal/cmd/secrets-manager/instance/instance.go
```diff
@@ -5,6 +5,7 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/cmd/secrets-manager/instance/delete"
```
vicentepinto98, line 5: Missing docs generation

vicentepinto98 (approved):

PR #170, opened 2024-03-20
Allow specifying empty list for CIDRSliceFlag

joaopalet:
Allow users to specify an empty CIDR slice by setting the flag to `""`.
E.g. `stackit <service> instance update <instance ID> --acl ""` sets the ACLs of the instance to the empty list.

This makes the custom flag `CIDRSliceFlag` have the same behaviour for empty values as the more generic and default `StringSlice` flag type

### DIFF
--- a/internal/pkg/flags/cidrslice.go
+++ b/internal/pkg/flags/cidrslice.go
@@ -24,7 +24,11 @@ func (f *cidrSliceFlag) String() string {
 
 func (f *cidrSliceFlag) Set(value string) error {
 	if value == "" {
-		f.value = []string{}
+		// If it's the first value to be set to the flag, we set it to an empty list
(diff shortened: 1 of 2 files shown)


vicentepinto98 (approved):

vicentepinto98:
Are you sure this will work for all services that have the --acl flag? I think some of them do not accept empty list

vicentepinto98 (changes requested):
Are you sure this will work for all services that have the --acl flag? I think some of them do not accept empty list

joaopalet:
As discussed, the flag logic should always allow setting an empty slice

I tested with all services that use this flag and the ones that do not accept an empty list validate it in the API and the error response is clear enough, so I think we don't need to validate it in the CLI

vicentepinto98 (approved):

PR #181, opened 2024-04-01
SKE Credentials Rotation: complete-rotation command 

DiogoFerrao:
Documentation follows the portal, shown here:
image

### DIFF
--- a/internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
+++ b/internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
@@ -30,10 +30,16 @@ func NewCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   fmt.Sprintf("complete-rotation %s", clusterNameArg),
 		Short: "Completes the rotation of the credentials associated to a SKE cluster",
-		Long: fmt.Sprintf("%s\n%s\n%s",
+		Long: fmt.Sprintf("%s\n%s\n%s\n%s\n%s\n\n%s\n%s\n%s\n%s",
 			"Completes the rotation of the credentials associated to a STACKIT Kubernetes Engine (SKE) cluster.",
-			"This is step 2 of a two-step process, if you haven't, please start the process by running:",
-			"  $ stackit ske credentials start-rotation my-cluster"),
+			"To ensure continued access to the Kubernetes cluster, please update your kubeconfig service account to the newly created account.",
+			"This is step 1 of a 2-step process to rotate all SKE cluster credentials. Tasks accomplished in this phase include:",
+			"  - The old certification authority will be dropped from the package.",
+			"  - The old signing key for the service account will be dropped from the bundle.",
+			"If you haven't, please start the process by running:",
+			"  $ stackit ske credentials start-rotation my-cluster",
+			"Generate a new kubeconfig file by running:",
+			"  $ stackit ske kubeconfig create my-cluster"),
 		Args: args.SingleArg(clusterNameArg, nil),
 		Example: examples.Build(
 			examples.NewExample(
@@ -84,6 +90,7 @@ func NewCmd() *cobra.Command {
 				operationState = "Triggered completion of credentials rotation"
 			}
 			cmd.Printf("%s for cluster %q\n", operationState, model.ClusterName)
+			cmd.Printf("Consider updating your kubeconfig with the new credentials, create a new kubeconfig by running:\n  $ stackit ske kubeconfig create %s\n", model.ClusterName)
 			return nil
 		},
 	}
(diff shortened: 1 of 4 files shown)


internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("complete-rotation %s", clusterNameArg),
+		Short: "Completes the rotation of the credentials associated to a SKE cluster",
+		Long: fmt.Sprintf("%s\n%s\n%s\n%s\n%s\n\n%s\n%s\n%s\n%s",
+			"Completes the rotation of the credentials associated to a STACKIT Kubernetes Engine (SKE) cluster.",
+			"To ensure continued access to the Kubernetes cluster, please update your kubeconfig service account to the newly created account.",
+			"This is step 1 of a 2-step process to rotate all SKE cluster credentials. Tasks accomplished in this phase include:",
```
joaopalet, line 36: Copy paste leftover?

joaopalet: This is step 2 and completes the process

internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
```diff
@@ -0,0 +1,117 @@
...
+			"  - The old certification authority will be dropped from the package.",
+			"  - The old signing key for the service account will be dropped from the bundle.",
+			"If you haven't, please start the process by running:",
+			"  $ stackit ske credentials start-rotation my-cluster",
+			"Generate a new kubeconfig file by running:",
+			"  $ stackit ske kubeconfig create my-cluster"),
```
joaopalet, line 42: Is this meant like a follow-up tip? Maybe we could emphasize that this can be done after the complete rotation, e.g. "After completing the rotation, you can generate ...`

internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
```diff
@@ -0,0 +1,117 @@
...
+		Args: args.SingleArg(clusterNameArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Complete the rotation of the credentials associated to the SKE cluster with name "my-cluster"`,
+				"$ stackit ske credentials complete-rotation my-cluster"),
+		),
```
joaopalet, line 48: I am wondering if it's worth to extend the example (or add a new one) with the complete flow (start rotation, complete rotation and create a new kubeconfig). What do you think @vicentepinto98?

joaopalet: Same for the other commands

internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
```diff
@@ -0,0 +1,117 @@
...
+				}
+				s.Stop()
+			}
+
+			operationState := "Rotation of credentials is completed"
+			if model.Async {
+				operationState = "Triggered completion of credentials rotation"
+			}
+			cmd.Printf("%s for cluster %q\n", operationState, model.ClusterName)
+			cmd.Printf("Consider updating your kubeconfig with the new credentials, create a new kubeconfig by running:\n  $ stackit ske kubeconfig create %s\n", model.ClusterName)
```
joaopalet, line 93: At this point the user should have already updated their kubeconfigs, right? Because the old credentials are no longer working after the complete rotation

joaopalet: So I would say we could either phrase it like "If you haven't yet, ...` or maybe remove this altogether

joaopalet: If we go with the first one, this would be another candidate for the (future) "warning" verbosity print

internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
vicentepinto98: I agree, an example with the complete flow would be nice and make it clearer

internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
vicentepinto98: This would be an important warning to have, probably before prompting for confirmation, similar to the portal where you see that warning before pressing "Continue"

vicentepinto98:
In general, getting some feedback from the SKE team regarding the commands help would be important

joaopalet:
> In general, getting some feedback from the SKE team regarding the commands help would be important

Good idea, but since this is a feature branch we could ask them after merging

internal/cmd/ske/credentials/complete-rotation/complete_rotation.go
DiogoFerrao: I don't think this is a big concern for a CLI user, they can always get the new kubeconfig with the CLI command. It isn't a "No going back" situation, access can always be provided at any point. My thinking with making it a suggestion at the end of the completion step is to serve as a reminder more than necessarily a warning, specially since there are legitimate use cases where the user may not even want to have kubectl access to the cluster.

vicentepinto98: Ok it makes sense, then let's keep it at the end and make it a warning once the new verbosity pkg is in place

joaopalet (approved):

PR #184, opened 2024-04-01
SKE Credentials Rotation: kubeconfig create command

DiogoFerrao:
## Some notes on implementation

This command is creating a kubeconfig and writing it to a file. By default, this location is ~/.kube/config (which is default for kubernetes) but can be changed via flag.
This choice is informed by behavior in GCP and AWS:
- https://docs.aws.amazon.com/eks/latest/userguide/create-kubeconfig.html#create-kubeconfig-automatically
- https://cloud.google.com/sdk/gcloud/reference/container/clusters/get-credentials

When setting the expiration time, I created a helper function to save the user from writing large timeouts (eg. 3 months) in seconds (the unit the API ingests)

### DIFF
--- a/internal/cmd/ske/kubeconfig/create/create.go
+++ b/internal/cmd/ske/kubeconfig/create/create.go
@@ -92,14 +92,14 @@ func NewCmd() *cobra.Command {
 					return fmt.Errorf("could not get user home directory: %w", err)
 				}
 
-				err = os.MkdirAll(fmt.Sprintf("%s/.kube", userHome), 0700)
+				err = os.MkdirAll(fmt.Sprintf("%s/.kube", userHome), 0o700)
 				if err != nil {
 					return fmt.Errorf("could not create kube directory: %w", err)
 				}
 				configPath = utils.Ptr(fmt.Sprintf("%s/.kube", userHome))
 			}
 
-			err = os.WriteFile(fmt.Sprintf("%s/config", *configPath), []byte(*resp.Kubeconfig), 0600)
+			err = os.WriteFile(fmt.Sprintf("%s/config", *configPath), []byte(*resp.Kubeconfig), 0o600)
 			if err != nil {
 				return fmt.Errorf("could not write kubeconfig file: %w", err)
 			}
--- a/internal/cmd/ske/kubeconfig/create/create.go
+++ b/internal/cmd/ske/kubeconfig/create/create.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strconv"
 
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/confirm"
@@ -13,6 +12,7 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/pkg/flags"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/services/ske/client"
+	skeUtils "github.com/stackitcloud/stackit-cli/internal/pkg/services/ske/utils"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
 
 	"github.com/spf13/cobra"
@@ -29,7 +29,7 @@ const (
 type inputModel struct {
 	*globalflags.GlobalFlagModel
 	ClusterName    string
-	KubeconfigPath *string
+	Location       *string
 	ExpirationTime *string
 }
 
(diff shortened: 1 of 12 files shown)


internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return &inputModel{
+		GlobalFlagModel: globalFlags,
+		ClusterName:     clusterName,
+		KubeconfigPath:  flags.FlagToStringPointer(cmd, locationFlag),
+		ExpirationTime:  expirationTime,
+	}, nil
+}
+
+func convertToSeconds(timeStr string) *string {
```
joaopalet, line 146: We should write unit tests for this function (instead of just testing it as part of the `parseInput`)

joaopalet: Maybe this function could even go to a `utils` file of the service in `pkg/`

internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -0,0 +1,198 @@
...
+		expirationTime = convertToSeconds(*expirationTimeInput)
+		if expirationTime == nil {
+			return nil, fmt.Errorf("invalid expiration time: %s", *expirationTimeInput)
+		}
+	}
+
+	return &inputModel{
+		GlobalFlagModel: globalFlags,
+		ClusterName:     clusterName,
+		KubeconfigPath:  flags.FlagToStringPointer(cmd, locationFlag),
```
joaopalet, line 141: I think usually we always keep the `inputModel` field with the same name as the exposed flag, any specific reason to have it dirrently here?

internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -0,0 +1,198 @@
...
+				}
+				configPath = utils.Ptr(fmt.Sprintf("%s/.kube", userHome))
+			}
+
+			err = os.WriteFile(fmt.Sprintf("%s/config", *configPath), []byte(*resp.Kubeconfig), 0o600)
+			if err != nil {
+				return fmt.Errorf("could not write kubeconfig file: %w", err)
+			}
+
+			fmt.Printf("Created kubeconfig file for cluster %s with expiration date %v.\n", model.ClusterName, *resp.ExpirationTimestamp)
```
joaopalet, line 107: Do you think it would be relevant to also indicate where the kubeconfig was written to?

internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -0,0 +1,198 @@
...
+	unit := timeStr[len(timeStr)-1:]
+	if _, err := strconv.Atoi(unit); err == nil {
+		// If the last character is a digit, assume the whole string is a number of seconds
+		return utils.Ptr(timeStr)
+	}
+
+	valueStr := timeStr[:len(timeStr)-1]
+	value, err := strconv.ParseUint(valueStr, 10, 64)
+	if err != nil {
+		return nil
```
joaopalet, line 160: This function should return an `error` too

internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -0,0 +1,198 @@
...
+
+	globalFlags := globalflags.Parse(cmd)
+	if globalFlags.ProjectId == "" {
+		return nil, &errors.ProjectIdError{}
+	}
+
+	expirationTimeInput := flags.FlagToStringPointer(cmd, expirationFlag)
+	var expirationTime *string
+	if expirationTimeInput != nil {
+		expirationTime = convertToSeconds(*expirationTimeInput)
```
joaopalet, line 132: Is there any other `parseInput` where we are adding extra logic besided mapping the flag to the model? If not I think we could separate these two and convert to seconds before/when building the request

joaopalet (commented):
Well spotted in noticing the usual behaviour was to write the file to a specific default location

internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -0,0 +1,150 @@
...
+				return fmt.Errorf("no kubeconfig returned from the API")
+			}
+
+			if model.Location == nil {
+				kubeconfigPath, err := skeUtils.GetDefaultKubeconfigLocation()
+				if err != nil {
+					return fmt.Errorf("get default kubeconfig location: %w", err)
+				}
+				model.Location = &kubeconfigPath
+			}
```
joaopalet, line 98: Could you write this to another variable? Just to avoi modifying the input model

Something like
```go
configLocation := model.Location
if configLocation == nil {
    kubeconfigPath, err := skeUtils.GetDefaultKubeconfigLocation()
    if err != nil {
        return fmt.Errorf("get default kubeconfig location: %w", err)
    }
    configLocation = &kubeconfigPath
}
```

internal/cmd/ske/kubeconfig/create/create.go
```diff
@@ -0,0 +1,150 @@
...
+
+func buildRequest(ctx context.Context, model *inputModel, apiClient *ske.APIClient) (ske.ApiCreateKubeconfigRequest, error) {
+	req := apiClient.CreateKubeconfig(ctx, model.ProjectId, model.ClusterName)
+
+	payload := ske.CreateKubeconfigPayload{}
+
+	if model.ExpirationTime != nil {
+		expirationTime, err := skeUtils.ConvertToSeconds(*model.ExpirationTime)
+		if err != nil {
+			return req, fmt.Errorf("parsing expiration time: %w", err)
```
joaopalet, line 143:
```suggestion
			return req, fmt.Errorf("parse expiration time: %w", err)
```

internal/pkg/services/ske/utils/utils.go
```diff
@@ -191,3 +194,74 @@ func getDefaultPayloadNodepool(resp *ske.ProviderOptions) (*ske.Nodepool, error)
...
+// WriteConfigFile writes the given data to the given path.
+// The directory is created if it does not exist.
+func WriteConfigFile(configPath, data string) error {
+	if data == "" {
+		return fmt.Errorf("no data to write")
+	}
+
+	dir := filepath.Dir(configPath)
+
+	err := os.MkdirAll(dir, 0o700)
```
joaopalet, line 247: Could this also be `0o600`?

internal/pkg/services/ske/utils/utils_test.go
```diff
@@ -441,3 +442,162 @@ func TestGetDefaultPayload(t *testing.T) {
...
+	tests := []struct {
+		description string
+		location    string
+		kubeconfig  string
+		isValid     bool
+		expectedErr string
+	}{
+		{
+			description: "base",
+			location:    "test_data/base/config",
```
joaopalet, line 546: Could be moved to a variable as discussed offline

internal/pkg/services/ske/utils/utils_test.go
```diff
@@ -553,29 +553,29 @@ func TestWriteConfigFile(t *testing.T) {
...
 			kubeconfig:  "",
 			isValid:     false,
 		},
 	}
 
+	baseTestDir := "test_data/"
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
-			err := WriteConfigFile(tt.location, tt.kubeconfig)
+			testLocation := baseTestDir + tt.location
```
joaopalet, line 573: Will this work on Windows? Probably we need to `filepath.Join`

joaopalet (approved):

PR #201, opened 2024-04-08
Move confirmation prompt to Print package

DiogoFerrao:


internal/pkg/print/print.go
```diff
@@ -85,3 +91,27 @@ func (p *Printer) Error(msg string) {
...
 	return p.Cmd.OutOrStdout()
 }
+
+// Prompts the user for confirmation.
+//
+// Returns nil only if the user (explicitly) answers positive.
+// Returns ErrAborted if the user answers negative.
+func (p *Printer) PromptForConfirmation(cmd *cobra.Command, prompt string) error {
+	question := fmt.Sprintf("%s [y/N] ", prompt)
+	reader := bufio.NewReader(cmd.InOrStdin())
```
DiogoFerrao, line 101: Would it make sense to have a Print.InOrStdin() for this? Don't really see a use case for this now, but could be useful in the future

internal/pkg/print/print.go
```diff
@@ -85,3 +91,27 @@ func (p *Printer) Error(msg string) {
 func (p *Printer) OutOrStdout() io.Writer {
 	return p.Cmd.OutOrStdout()
 }
+
+// Prompts the user for confirmation.
+//
+// Returns nil only if the user (explicitly) answers positive.
+// Returns ErrAborted if the user answers negative.
+func (p *Printer) PromptForConfirmation(cmd *cobra.Command, prompt string) error {
```
joaopalet, line 99: The printer already has a field for the command (`p.Cmd`), so we could probably remove it from the function arguments

internal/pkg/print/print.go
joaopalet: Maybe, or we could also just call `p.Cmd.InOrStdin()` (and do the same for `p.Cmd.InOrStdin()`, removing `(p *Printer) OutOrStdout()`)

internal/pkg/print/print.go
```diff
@@ -85,3 +91,27 @@ func (p *Printer) Error(msg string) {
...
+	question := fmt.Sprintf("%s [y/N] ", prompt)
+	reader := bufio.NewReader(cmd.InOrStdin())
+	for i := 0; i < 3; i++ {
+		cmd.PrintErr(question)
```
joaopalet, line 103: This one still calls the command's `PrintErr` directly, as it does not depend on the verbosity level, do we want to keep it?

DiogoFerrao: I don't think I understood, what do we still want to keep? 
I think its ok that the confirmation prompt does not depend on verbosity, since we would want to have it at every verbosity level right?

internal/pkg/print/print.go
vicentepinto98: I think you meant `p.Cmd.InOrStdout` the second time @joaopalet, but I agree, it seems unnecessary to have those methods defined in the `print` pkg

DiogoFerrao: Yes, calling `p.Cmd.InOrStdin()` makes sense. 
However, I don't understand the removing the `(p *Printer) OutOrStdout()`.

internal/pkg/print/print.go
vicentepinto98: Yes, the prompts for confirmation don't depend on the verbosity level
We should call `p.cmd.PrintErr` here

joaopalet: I was just wondering if we could create a new method in the `print` package that would not depend on the verbosity level and would always run `p.Cmd.PrintErr`

But I think this would be the only use case for it now so I agree we should keep it like this

internal/pkg/print/print.go
DiogoFerrao: But removing `p.OutOrStdout()` would mean that the `pager.Display` method, from the pager package, would have to receive a cmd in arguments. 
I think it makes more sense to keep this method than change the Display signature

joaopalet: But in `pager.Display` we could also do `p.Cmd.OutOrStdout`, right?

joaopalet (approved):

PR #203, opened 2024-04-08
Logging: change spinner to print package, update docs

DiogoFerrao:


internal/pkg/spinner/spinner.go
```diff
@@ -3,21 +3,21 @@ package spinner
...
 	delay     time.Duration
 	done      chan bool
 }
 
-func New(cmd *cobra.Command) *Spinner {
+func New(p *print.Printer) *Spinner {
 	return &Spinner{
-		cmd:       cmd,
+		printer:   p,
 		states:    []string{"|", "/", "-", "\\"},
```
joaopalet, line 21: You should also need to replace this with `printer: p`

internal/pkg/spinner/spinner.go
```diff
@@ -33,7 +33,7 @@ func (s *Spinner) Start(message string) {
 func (s *Spinner) Stop() {
 	s.done <- true
 	close(s.done)
-	s.cmd.Printf("\r%s ✓ \n", s.message)
+	s.printer.Info("\r%s ✓ \n", s.message)
 }
```
joaopalet, line 37: And this could also be replaced

joaopalet (changes requested):

internal/pkg/spinner/spinner.go
joaopalet: Hm, looks like I was seeing an older version

joaopalet (dismissed):

joaopalet (approved):

PR #209, opened 2024-04-08
Logging: move warnings to verbosity package

DiogoFerrao:
This PR changes a lot of files because there is a warning in the pkg, namely in the auth package. So every client and command that calls configureClient needs to be updated.

Also created a new method printer.Warnf() to allow the usage of PrintF syntax in warnings.


internal/pkg/print/print.go
```diff
@@ -81,6 +81,16 @@ func (p *Printer) Warn(msg string) {
 	p.Cmd.PrintErrf("Warning: %s\n", msg)
 }
 
+// Print a Warn level output to the defined Err output (falling back to Stderr if not set).
+// If the verbosity level is not Debug, Info, or Warn, it does nothing.
+func (p *Printer) Warnf(msg string, args ...any) {
```
vicentepinto98, line 86: I think this method is not necessary. For uniformity in the pkg, I would rather change the signature of the `Warn` method to `Warn(msg string, args ...any)`, same as `Info`.

Then it would make sense to also change the signature of the `Error` method to take the same arguments

vicentepinto98: Even though for `Error` the args won't really be used as all errors are passed on to the parent command and in the end printed in the root file

Issue #153, opened 2024-03-12
Config file is being saved on every invocation of the CLI


Kumm-Kai:
Currently, the STACKIT CLI calls the [`config.InitConfig`](https://github.com/stackitcloud/stackit-cli/blob/main/internal/pkg/config/config.go#L69) function on every invocation.

On the very first call of the binary (no `.stackit` directory exists), it creates the directory and a config file (with default values).
On every following binary call the config file is being read **and** directly written to disk again.

This is probably only problematic in some small edge cases, like calling `stackit` multiple times in one command using `process substiution` (e.g., `<(stackit xyz)`). In this case the CLI fails because `viper.WriteConfigAs` is not concurrency-safe.

This can be solved by either:
1. Only creating the directory & config file if it is actually needed, e.g., when calling `stackit config set/unset`
2. Only writing the config file when necessary (e.g., file on disk differs from config in memory)

From my point of view the first solution would be the best one because in my opinion there is no need to create a config file if the user didn't change the defaults.

joaopalet:
Hi @Kumm-Kai,
Thanks for the suggestion! This was addressed in the PR linked above and will be included in the next release.

Kumm-Kai:
Great, thanks for looking into it @joaopalet ❤️

solved by 81b75942e3c381edccd93277db3144ab6910eb4b

Kumm-Kai:
#185 doesn't completely solve this issue. A empty file is still being created on the very first invocation.
This still results in failures when using process substiution.
I would propose only creating the file **and** directory if something should actually be written.

If you want to I can open a PR for it.

joaopalet:
Ah good point, I though the issue was only in `viper.WriteConfigAs`.

If you want to raise the PR making the fix I would be happy to review and support, otherwise I will try to do it within the next few days.

joaopalet:
This was addressed in #214 and will be included in the next release.

Thanks @Kumm-Kai for the contribution 💪

PR #214, opened 2024-04-11
Change config handling: only create config file & directory if needed

Kumm-Kai:
This PR tries to solve #153 by only creating the config file & directory on commands that actually need to write the config, e.g., `set` & `unset`.

Every command that needs to write the config should now use the `config.Write` function instead of directly calling `viper.WriteConfig`. The `config.Write` function ensures that the config directory exists and creates it if necessary. Oddly, `viper` doesn't support this out of the box...

I additionally found that `viper.WriteConfig` also writes the default values (set with `viper.SetDefault`) to the config. This means that the defaults are not being used after the config file is first written. I don't think that's something we can/should easily change, but we should definitely keep it in mind that future default changes might not reach the users if they already have a config file.

### DIFF
--- a/internal/cmd/config/set/set.go
+++ b/internal/cmd/config/set/set.go
@@ -82,7 +82,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 
 			err = config.Write()
 			if err != nil {
-				return fmt.Errorf("writing config file: %w", err)
+				return fmt.Errorf("write config to file: %w", err)
 			}
 			return nil
 		},
--- a/internal/cmd/config/set/set.go
+++ b/internal/cmd/config/set/set.go
@@ -80,9 +80,9 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				viper.Set(config.ProjectNameKey, "")
 			}
 
-			err = viper.WriteConfig()
+			err = config.Write()
 			if err != nil {
-				return fmt.Errorf("write new config to file: %w", err)
+				return fmt.Errorf("writing config file: %w", err)
 			}
 			return nil
 		},
--- a/internal/cmd/config/unset/unset.go
+++ b/internal/cmd/config/unset/unset.go
@@ -146,7 +146,7 @@ func NewCmd() *cobra.Command {
 
 			err := config.Write()
 			if err != nil {
-				return fmt.Errorf("writing config file: %w", err)
+				return fmt.Errorf("write config to file: %w", err)
 			}
 			return nil
 		},
--- a/internal/cmd/config/unset/unset.go
+++ b/internal/cmd/config/unset/unset.go
@@ -144,9 +144,9 @@ func NewCmd() *cobra.Command {
 				viper.Set(config.SKECustomEndpointKey, "")
 			}
 
-			err := viper.WriteConfig()
+			err := config.Write()
 			if err != nil {
-				return fmt.Errorf("write updated config to file: %w", err)
+				return fmt.Errorf("writing config file: %w", err)
 			}
 			return nil
 		},
--- a/internal/pkg/auth/storage.go
+++ b/internal/pkg/auth/storage.go
@@ -136,7 +136,7 @@ func GetAuthField(key authFieldKey) (string, error) {
 		var errFallback error
 		value, errFallback = getAuthFieldFromEncodedTextFile(key)
 		if errFallback != nil {
-			return "", fmt.Errorf("read from keyring failed (%w), read from encoded file also failed: %w", err, errFallback)
+			return "", fmt.Errorf("read from keyring: %w, read from encoded file as fallback: %w", err, errFallback)
 		}
 	}
 	return value, nil
--- a/internal/pkg/auth/storage.go
+++ b/internal/pkg/auth/storage.go
@@ -136,7 +136,7 @@ func GetAuthField(key authFieldKey) (string, error) {
 		var errFallback error
 		value, errFallback = getAuthFieldFromEncodedTextFile(key)
 		if errFallback != nil {
-			return "", fmt.Errorf("write to keyring failed (%w), tried write to encoded text file: %w", err, errFallback)
+			return "", fmt.Errorf("read from keyring failed (%w), read from encoded file also failed: %w", err, errFallback)
 		}
 	}
 	return value, nil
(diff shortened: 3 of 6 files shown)


internal/pkg/auth/storage.go
```diff
@@ -136,7 +136,7 @@ func GetAuthField(key authFieldKey) (string, error) {
 		var errFallback error
 		value, errFallback = getAuthFieldFromEncodedTextFile(key)
 		if errFallback != nil {
-			return "", fmt.Errorf("write to keyring failed (%w), tried write to encoded text file: %w", err, errFallback)
+			return "", fmt.Errorf("read from keyring failed (%w), read from encoded file also failed: %w", err, errFallback)
```
joaopalet, line 139: Thanks for trying to improve the error handling here, but was there any error message coming from here that was unclear?

The practise we are following for our Go error wrapping is that wrapper strings do not need to mention that something failed if that is already clear in the innermost error. E.g.:
- This: `send message: establish connection: connection refused`
- Instead of this: `sending message: establishing connection: connection refused`

internal/pkg/config/config.go
```diff
@@ -70,6 +71,8 @@ var ConfigKeys = []string{
 	SKECustomEndpointKey,
 }
 
+var FolderPath string
```
joaopalet, line 74: I think this `var` could be private:

```suggestion
var folderPath string
```

internal/pkg/config/config.go
```diff
@@ -80,23 +83,28 @@ func InitConfig() {
...
+
 	setConfigDefaults()
 
 	viper.AutomaticEnv()
 	viper.SetEnvPrefix("stackit")
 }
 
-func createFolderIfNotExists(folderPath string) error {
-	_, err := os.Stat(folderPath)
+func CreateFolderIfNotExists() error {
```
joaopalet, line 104: Also this method I think it doesn't need to be public:

```suggestion
func createFolderIfNotExists() error {
```

internal/cmd/config/set/set.go
```diff
@@ -80,9 +80,9 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				viper.Set(config.ProjectNameKey, "")
 			}
 
-			err = viper.WriteConfig()
+			err = config.Write()
 			if err != nil {
-				return fmt.Errorf("write new config to file: %w", err)
+				return fmt.Errorf("writing config file: %w", err)
```
joaopalet, line 85: Related to the comment above about error wrapping, we generally also use the imperative forms of verbs or no verbs at all to help to keep the message concise


```suggestion
				return fmt.Errorf("write config to file: %w", err)
```

internal/cmd/config/unset/unset.go
```diff
@@ -144,9 +144,9 @@ func NewCmd() *cobra.Command {
 				viper.Set(config.SKECustomEndpointKey, "")
 			}
 
-			err := viper.WriteConfig()
+			err := config.Write()
 			if err != nil {
-				return fmt.Errorf("write updated config to file: %w", err)
+				return fmt.Errorf("writing config file: %w", err)
```
joaopalet, line 149:
```suggestion
				return fmt.Errorf("write config to file: %w", err)
```

joaopalet:
Thanks for the contribution @Kumm-Kai!

Left a few comments but the logic looks good and it works as expected 🙂


> I additionally found that viper.WriteConfig also writes the default values (set with viper.SetDefault) to the config. This means that the defaults are not being used after the config file is first written. I don't think that's something we can/should easily change, but we should definitely keep it in mind that future default changes might not reach the users if they already have a config file.

I'm not sure I understood this right, but `viper.SetDefault` is meant to set the default values only if the user hasn't set any values himself for that field in the config file. So as if we add a new field with a default value it should be written to the config file

internal/pkg/config/config.go
```diff
@@ -80,23 +83,28 @@ func InitConfig() {
 	viper.SetConfigType(configFileExtension)
 	viper.AddConfigPath(configFolderPath)
```
joaopalet, line 84: One more thing, I think these are not needed anymore if we are using `SetConfigFile`, right? As this will override those values

internal/pkg/auth/storage.go
joaopalet: On a second though, this is kind of a special case with two inner errors and I agree with your suggested improvement :+1:

vicentepinto98: We could also write
`fmt.Errorf("read from keyring: %w, read from encoded file as fallback: %w", err, errFallback)`

Kumm-Kai:
> I'm not sure I understood this right, but `viper.SetDefault` is meant to set the default values only if the user hasn't set any values himself for that field in the config file. So as if we add a new field with a default value it should be written to the config file

New fields should be fine. I was thinking about changing the default of an already existing fields. But admittedly, there aren't many cases where we could need this.

internal/pkg/auth/storage.go
Kumm-Kai: I changed the error because pervously it said "write to keyring failed" but this is the `GetAuthField` function which reads and not writes :)

Kumm-Kai:
@joaopalet thanks for the review 😁
I've implemented all requested changes.

internal/pkg/auth/storage.go
joaopalet: Ah didn't even notice that, even better than. Thanks!

internal/pkg/config/config.go
```diff
@@ -70,30 +70,36 @@ var ConfigKeys = []string{
...
+	// see https://github.com/spf13/viper/issues/851#issuecomment-789393451
+	viper.SetConfigFile(configFilePath)
+
+	f, err := os.Open(configFilePath)
+	if !os.IsNotExist(err) {
+		if err := viper.ReadConfig(f); err != nil {
+			cobra.CheckErr(err)
+		}
+	}
+	defer f.Close() //nolint:errcheck // file close errors are hard to handle
```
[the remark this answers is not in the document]
Kumm-Kai: Now `cobra.CheckErr(err)` returns `invalid argument`?

joaopalet (approved):

PR #215, opened 2024-04-15
Add repology packaging status badge to INSTALLATION.md

Kumm-Kai:
This PR adds the [repology](https://repology.org/) [packaging status badge](https://repology.org/project/stackit-cli/badges) for the STACKIT CLI to the `INSTALLATION.md`.

This allows future users to quickly see which (native/official) repositories have the STACKIT CLI packaged and enables a fast installation using their most favorite package manager.

In the future, we can increase the `columns` parameter once the list gets too long.

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -2,6 +2,8 @@
 
 ## Package managers
 
+[![Packaging status](https://repology.org/badge/vertical-allrepos/stackit-cli.svg?columns=3)](https://repology.org/project/stackit-cli/versions)
+
 ### macOS
 
 The STACKIT CLI can be installed through the [Homebrew](https://brew.sh/) package manager.


README.md
```diff
@@ -10,7 +10,9 @@ Your feedback is appreciated!
 
 ## Installation
 
-Please refer to our [installation guide](./INSTALLATION.md) for instructions on how to install and get started using the STACKIT CLI.
+[![Packaging status](https://repology.org/badge/vertical-allrepos/stackit-cli.svg?columns=1)](https://repology.org/project/stackit-cli/versions)
```
joaopalet, line 13: Nice! Thanks for the contribution

Do you know think this fits better here in the README than in the installation guide?

joaopalet: The only concern I see is that those are not maintained by us, so we cannot guarantee they will always be up to date and working. So I'm not sure about mentioning them before and with more highlight than out official installation methods

vicentepinto98: In my opinion we should recommend users for our installation methods first, we can have the badge after the link to the Installation guide

Kumm-Kai: > The only concern I see is that those are not maintained by us, so we cannot guarantee they will always be up to date and working. So I'm not sure about mentioning them before and with more highlight than out officially supported installation methods

Fair point, the badge will mark outdated packages in red, so these will be visible to the user.

And you're correct about the fact that they are not maintained by you. I'm a bit split about this because users almost always prefer an installation via their native package manager & repository (so without any manual steps, like adding additional repos, just `apt install xyz`). On the other hand, we should definitely be more clear which methods are officially supported to reduce support ticket.

In the future, we might also maintain official packages in upstream package repos, then the repology list does not only contain "inofficial" repos.

joaopalet: Ok I see. I agree that it's nice to have this table tracking package versions across different known package repositories, but probably it also belongs to the `INSTALLATION.md` file, together with other installation instructions

Could you move itto the top of that file then?

Kumm-Kai: I can do that, but I again want to emphasize, that "we should show them directly where they can install the CLI from".

I've done it similar to https://github.com/eza-community/eza?tab=readme-ov-file#installation 

Let me know where you want to have it.

joaopalet: We've discussed within the team and we decided we will have it in the installation guide _for now_

We will reconsider this once we start distributing in channels watched by Repology, then we can even also have the exact Repology entry in each section of the installation instructions (like e.g. in https://github.com/eza-community/eza/blob/main/INSTALL.md)

PR #217, opened 2024-04-16
Add debug error logs

GokceGK:


internal/cmd/argus/instance/create/create.go
```diff
@@ -64,6 +64,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 
 			projectLabel, err := projectname.GetProjectName(ctx, cmd, p)
 			if err != nil {
+				p.Debug(print.ErrorLevel, "get project name: %v", err)
```
joaopalet, line 67: Not exactly sure about best practises here, but couldn't some of these logs be done in the function itself before returning the error instead of in all caller functions?

GokceGK: also thought about this but was not sure if that would be okay. But I can change it

vicentepinto98: I think that's breaking the error handling structure that we have, it would require some check before returning in the function which is not the standard.

Also, in that case we can't be sure that error won't be logged twice in the caller function

internal/pkg/services/argus/client/client.go
```diff
@@ -18,6 +18,7 @@ func ConfigureClient(p *print.Printer) (*argus.APIClient, error) {
 
 	authCfgOption, err := auth.AuthenticationConfig(p, auth.AuthorizeUser)
 	if err != nil {
+		p.Debug(print.ErrorLevel, "auth err: %v", err)
```
vicentepinto98, line 21:
```suggestion
		p.Debug(print.ErrorLevel, "configure authentication: %v", err)
```

internal/pkg/services/argus/client/client.go
```diff
@@ -30,6 +31,7 @@ func ConfigureClient(p *print.Printer) (*argus.APIClient, error) {
 
 	apiClient, err = argus.NewAPIClient(cfgOptions...)
 	if err != nil {
+		p.Debug(print.ErrorLevel, "auth err: %v", err)
```
vicentepinto98, line 34:
```suggestion
		p.Debug(print.ErrorLevel, "create new API client: %v", err)
```

internal/cmd/argus/instance/create/create.go
vicentepinto98: In general, logging the error should happen on the most top-level possible. We are logging here because this error is being "silenced" by the CLI

internal/pkg/flags/flag_to_value.go
```diff
@@ -5,13 +5,15 @@ import (
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
 )
 
 // Returns the flag's value as a string.
 // Returns "" if the flag is not set, if its value can not be converted to string, or if the flag does not exist.
-func FlagToStringValue(cmd *cobra.Command, flag string) string {
+func FlagToStringValue(cmd *cobra.Command, flag string, p *print.Printer) string {
 	value, err := cmd.Flags().GetString(flag)
 	if err != nil {
+		p.Debug(print.ErrorLevel, "convert flag to string value err: %v", err)
```
vicentepinto98, line 16:
```suggestion
		p.Debug(print.ErrorLevel, "convert flag to string value: %v", err)
```
Err in this case is redundant

internal/pkg/flags/flag_to_value.go
```diff
@@ -5,13 +5,15 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
 )
 
 // Returns the flag's value as a string.
 // Returns "" if the flag is not set, if its value can not be converted to string, or if the flag does not exist.
-func FlagToStringValue(cmd *cobra.Command, flag string) string {
+func FlagToStringValue(cmd *cobra.Command, flag string, p *print.Printer) string {
```
vicentepinto98, line 13: nitpick: pass the printer as the first argument (similar to the context, which is usually passed everywhere and is always the first argument, when present)

vicentepinto98: this comment applies to the rest of the PR

internal/pkg/flags/flag_to_value.go
vicentepinto98: this comment applies to the rest of the PR

internal/pkg/flags/flag_to_value.go
```diff
@@ -22,19 +24,21 @@ func FlagToStringValue(cmd *cobra.Command, flag string) string {
 
 // Returns the flag's value as a bool.
 // Returns false if its value can not be converted to bool, or if the flag does not exist.
-func FlagToBoolValue(cmd *cobra.Command, flag string) bool {
+func FlagToBoolValue(cmd *cobra.Command, flag string, p *print.Printer) bool {
 	value, err := cmd.Flags().GetBool(flag)
 	if err != nil {
+		p.Debug(print.ErrorLevel, "convert flag to bool value err: %v", err)
```
vicentepinto98, line 30: The flag string value could be added to the error message, but **only** if it is not shown in `err` already

internal/pkg/services/argus/client/client.go
```diff
@@ -18,6 +18,7 @@ func ConfigureClient(p *print.Printer) (*argus.APIClient, error) {
 
 	authCfgOption, err := auth.AuthenticationConfig(p, auth.AuthorizeUser)
 	if err != nil {
+		p.Debug(print.ErrorLevel, "configure authentication: %v", err)
```
vicentepinto98, line 21: This change should be done for all of the other clients

vicentepinto98 (commented):
Missing debug Error log on `getBearerToken` function right before returning `AuthError`

internal/pkg/flags/flag_to_value.go
GokceGK: made printer first argument in all files.

internal/pkg/services/argus/client/client.go
GokceGK: Edited debug messages in all files

internal/cmd/root.go
```diff
@@ -131,6 +131,7 @@ func Execute(version, date string) {
 	if err != nil {
 		err := beautifyUnknownAndMissingCommandsError(cmd, err)
 		p.Error(err.Error())
+		p.Debug(print.ErrorLevel, "execute err: %v", err)
```
GokceGK, line 134: This seems redundant to me as error message printed one line above.

vicentepinto98: But they serve different purposes: 
- `p.Error` is a normal print and is always shown to the user, no matter the `verbosity`
- `p.Debug` is a structured log and is only shown if verbosity is set to `debug` and includes more info such as the timestamp and the call trace

I would say for now we keep both and then if we get feedback that it is redundant (or we have the debug logs too cluttered) we can remove it

internal/pkg/auth/user_token_flow.go
```diff
@@ -53,6 +53,8 @@ func (utf *userTokenFlow) RoundTrip(req *http.Request) (*http.Response, error) {
 		err = refreshTokens(utf)
 		if err == nil {
 			accessTokenValid = true
+		} else {
+			utf.printer.Debug(print.ErrorLevel, "refresh access token err: %v", err)
```
vicentepinto98, line 57:
```suggestion
			utf.printer.Debug(print.ErrorLevel, "refresh access token: %v", err)
```

internal/cmd/root.go
vicentepinto98, line 134:
```suggestion
		p.Debug(print.ErrorLevel, "execute command: %v", err)
```

vicentepinto98 (dismissed):

PR #221, opened 2024-04-16
Add `argus grafana describe` command

joaopalet:


internal/cmd/argus/grafana/describe/describe.go
```diff
@@ -0,0 +1,129 @@
...
+		Use:   "describe",
+		Short: "Shows details of the Grafana configuration of an Argus instance",
+		Long:  "Shows details of the Grafana configuration of an Argus instance.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Get details of the Grafana configuration of an Argus instance with ID "xxx"`,
+				"$ stackit argus credentials describe --instance-id xxx"),
+			examples.NewExample(
+				`Get details of of the Grafana configuration of an Argus instance with ID "xxx" in a table format`,
```
DiogoFerrao, line 41:
```suggestion
				`Get details of the Grafana configuration of an Argus instance with ID "xxx" in a table format`,
```

internal/cmd/argus/grafana/describe/describe.go
```diff
@@ -0,0 +1,129 @@
...
+		InstanceId:      flags.FlagToStringValue(cmd, instanceIdFlag),
+	}, nil
+}
+
+func buildGrafanaConfigRequest(ctx context.Context, model *inputModel, apiClient *argus.APIClient) argus.ApiGetGrafanaConfigsRequest {
+	req := apiClient.GetGrafanaConfigs(ctx, model.InstanceId, model.ProjectId)
+	return req
+}
+
+func buildInstanceRequest(ctx context.Context, model *inputModel, apiClient *argus.APIClient) argus.ApiGetInstanceRequest {
```
DiogoFerrao, line 99:
```suggestion
func buildGetInstanceRequest(ctx context.Context, model *inputModel, apiClient *argus.APIClient) argus.ApiGetInstanceRequest {
```

internal/cmd/argus/grafana/describe/describe.go
```diff
@@ -0,0 +1,129 @@
...
+func outputResult(p *print.Printer, outputFormat string, grafanaConfigs *argus.GrafanaConfigs, instance *argus.GetInstanceResponse) error {
+	switch outputFormat {
+	case globalflags.PrettyOutputFormat:
+
+		table := tables.NewTable()
+		table.AddRow("GRAFANA DASHBOARD", *instance.Instance.GrafanaUrl)
+		table.AddSeparator()
+		table.AddRow("PUBLIC READ ACCESS", *grafanaConfigs.PublicReadAccess)
+		table.AddSeparator()
+		table.AddRow("SINGLE SIGN-ON", *grafanaConfigs.UseStackitSso)
```
DiogoFerrao, line 113: In the Grafana tab, the portal shows the initial Admin login to Grafana. We should probably do something similar in the CLI, right? 

In the instance describe this is not shown, so maybe here is the place if we are going to do it.

internal/cmd/argus/grafana/describe/describe_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+func fixtureGrafanaConfigsRequest(mods ...func(request *argus.ApiGetGrafanaConfigsRequest)) argus.ApiGetGrafanaConfigsRequest {
+	request := testClient.GetGrafanaConfigs(testCtx, testInstanceId, testProjectId)
+	for _, mod := range mods {
+		mod(&request)
+	}
+	return request
+}
+
+func fixtureInstanceRequest(mods ...func(request *argus.ApiGetInstanceRequest)) argus.ApiGetInstanceRequest {
```
DiogoFerrao, line 57:
```suggestion
func fixtureGetInstanceRequest(mods ...func(request *argus.ApiGetInstanceRequest)) argus.ApiGetInstanceRequest {
```

internal/cmd/argus/grafana/describe/describe_test.go
```diff
@@ -0,0 +1,253 @@
...
+				cmpopts.EquateComparable(testCtx),
+			)
+			if diff != "" {
+				t.Fatalf("Data does not match: %s", diff)
+			}
+		})
+	}
+}
+
+func TestBuildInstanceRequest(t *testing.T) {
```
DiogoFerrao, line 227:
```suggestion
func TestBuildGetInstanceRequest(t *testing.T) {
```

internal/cmd/argus/grafana/describe/describe.go
joaopalet: Great point, added those

internal/cmd/argus/grafana/describe/describe.go
```diff
@@ -0,0 +1,145 @@
...
+
+const (
+	instanceIdFlag   = "instance-id"
+	showPasswordFlag = "show-password"
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	InstanceId   string
+	ShowPassword bool
```
DiogoFerrao, line 29: In other commands we use the inverse, `hide-password` that is false by default. I think that we should keep the pattern for code uniformity. 

I believe there is a story to change this default, but we should probably keep false until that story is done.

internal/cmd/argus/grafana/describe/describe_test.go
```diff
@@ -0,0 +1,263 @@
...
+			isValid:       true,
+			expectedModel: fixtureInputModel(),
+		},
+		{
+			description: "show password",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[showPasswordFlag] = "true"
+			}),
+			isValid: true,
+			expectedModel: fixtureInputModel(func(model *inputModel) {
```
DiogoFerrao, line 85: You could also test for an invalid case and the explicitly "false" case

DiogoFerrao: Nevermind

internal/cmd/argus/grafana/describe/describe.go
joaopalet: I did it for similarity with the Portal, where the password is also hidden by default, but yes I'll leave it with `hide-password` for consistency within the CLI then

And yes, we have a story to change other `hide-password` to `show-password`

DiogoFerrao (approved):

PR #228, opened 2024-04-17
Add `argus grafana single-sign-on` commands

joaopalet:


internal/cmd/argus/grafana/single-sign-on/disable/disable.go
joaopalet, line 1: The `disable` command is almost similar to `enable`

Slight changes are mainly on:
- Command usage, description and examples
- Confirmation prompt and info prints
- In `buildRequest`, in `disable` we always set SSO to `false` where in `enable` we set it to `true`

internal/cmd/argus/grafana/single-sign-on/disable/disable_test.go
```diff
@@ -0,0 +1,265 @@
...
+		model                  *inputModel
+		getGrafanaConfigsFails bool
+		getGrafanaConfigsResp  *argus.GrafanaConfigs
+		isValid                bool
+		expectedRequest        argus.ApiUpdateGrafanaConfigsRequest
+	}{
+		{
+			description:            "base",
+			model:                  fixtureInputModel(),
+			getGrafanaConfigsFails: false,
```
DiogoFerrao, line 202: I think these can be ommited for clarity when false

internal/cmd/argus/grafana/single-sign-on/disable/disable_test.go
```diff
@@ -0,0 +1,265 @@
...
+			getGrafanaConfigsResp:  fixtureGrafanaConfigs(),
+			isValid:                true,
+			expectedRequest: fixtureRequest().UpdateGrafanaConfigsPayload(
+				argus.UpdateGrafanaConfigsPayload{
+					GenericOauth:     argusUtils.ToPayloadGenericOAuth(fixtureGrafanaConfigs().GenericOauth),
+					PublicReadAccess: fixtureGrafanaConfigs().PublicReadAccess,
+					UseStackitSso:    utils.Ptr(false),
+				}),
```
DiogoFerrao, line 210: Shouldn't this be the fixture request? That way the base case is simpler

DiogoFerrao: I'll leave the comment to get your thoughts on it, but thinking about it maybe this is better here because of the calls to argus utils.

internal/pkg/services/argus/utils/utils_test.go
```diff
@@ -461,7 +412,7 @@ func TestGetPartialUpdateGrafanaConfigsPayload(t *testing.T) {
 			description:           "no grafana configs",
 			singleSignOn:          utils.Ptr(true),
 			publicReadAccess:      utils.Ptr(true),
-			getGrafanaConfigsResp: &argus.GrafanaConfigs{},
+			getGrafanaConfigsResp: nil,
 			isValid:               false,
 		},
```
DiogoFerrao, line 417: Empty response could also be a good test

internal/cmd/argus/grafana/single-sign-on/disable/disable_test.go
joaopalet: Totally agree, good suggestion

internal/pkg/services/argus/utils/utils_test.go
```diff
@@ -300,61 +305,7 @@ func TestToPayloadGenericOAuth(t *testing.T) {
```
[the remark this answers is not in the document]
joaopalet: In this case the input is already the OAuth field, and that is already being tested. Adjusted the field name for clarity

DiogoFerrao (approved):

PR #231, opened 2024-04-18
Onboard Argus Scrape Configs: generate-payload command

DiogoFerrao:


internal/pkg/services/argus/utils/utils.go
```diff
@@ -56,6 +57,143 @@ func LoadPlanId(planName string, resp *argus.PlansResponse) (*string, error) {
...
+		configs := make([]argus.UpdateScrapeConfigPayloadStaticConfigsInner, 0)
+		for _, config := range *data.StaticConfigs {
+			newConfig, err := mapStaticConfig(config)
+			if err != nil {
+				return nil, fmt.Errorf("map static config: %w", err)
+			}
+			configs = append(configs, newConfig)
+		}
+		staticConfigs = &configs
+	}
```
vicentepinto98, line 80: I suggest some changes to this block (and the following ones) to make it more readable:

```
var staticConfigs []argus.UpdateScrapeConfigPayloadStaticConfigsInner
if data.StaticConfigs != nil {
    for _, config := range *data.StaticConfigs {
        newConfig, err := mapStaticConfig(config)
        if err != nil {
            return nil, fmt.Errorf("map static config: %w", err)
        }
        staticConfigs = append(staticConfigs, newConfig)
    }
}

(...)

return &argus.UpdateScrapeConfigPayload{
		BasicAuth:             basicAuth,
		BearerToken:           data.BearerToken,
		HonorLabels:           data.HonorLabels,
		HonorTimeStamps:       data.HonorTimeStamps,
		MetricsPath:           data.MetricsPath,
		(...)
		StaticConfigs:         &staticConfigs,
		(...)
	}, nil
```

Doing `var staticConfigs []argus.UpdateScrapeConfigPayloadStaticConfigsInner` or `make([]argus.UpdateScrapeConfigPayloadStaticConfigsInner, 0)` is the same thing and the first one is more readable. What you could do instead would be `make([]argus.UpdateScrapeConfigPayloadStaticConfigsInner, len(*data.StaticConfigs))` but then this would require the nil check first which would make it more complex without much added improvement on performance for most likely small arrays

internal/pkg/services/argus/utils/utils.go
```diff
@@ -56,6 +57,143 @@ func LoadPlanId(planName string, resp *argus.PlansResponse) (*string, error) {
...
+		SampleLimit:           convertIntToFloat64(data.SampleLimit),
+		Scheme:                data.Scheme,
+		ScrapeInterval:        data.ScrapeInterval,
+		ScrapeTimeout:         data.ScrapeTimeout,
+		StaticConfigs:         staticConfigs,
+		TlsConfig:             tlsConfig,
+	}, nil
+}
+
+func convertIntToFloat64(i *int64) *float64 {
```
vicentepinto98, line 120: Quite unfortunate that the API has different types in the response and in the payload.

I think this fits better in the general `utils` pkg, might be useful for other services in the future.

internal/pkg/services/argus/utils/utils.go
```diff
@@ -56,6 +57,143 @@ func LoadPlanId(planName string, resp *argus.PlansResponse) (*string, error) {
...
+	if tlsConfig == nil {
+		return nil
+	}
+
+	return &argus.CreateScrapeConfigPayloadHttpSdConfigsInnerOauth2TlsConfig{
+		InsecureSkipVerify: tlsConfig.InsecureSkipVerify,
+	}
+}
+
+func convertMapAnyToInterface(m interface{}) (map[string]interface{}, error) {
```
vicentepinto98, line 178: wouldn't it make more sense to have a function to map specifically `params` and `staticConfig.Labels` instead of this generic one? 
We should avoid receiving interface as an argument

internal/cmd/argus/scrape-configs/generate-payload/generate_payload.go
```diff
@@ -0,0 +1,136 @@
...
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "generate-payload",
+		Short: "Generates a payload to create/update Scrape Configurations for an Argus instance ",
+		Long: fmt.Sprintf("%s\n%s\n%s\n%s\n%s\n",
+			"Generates a JSON payload with values to be used as --payload input for Scrape Configurations creation or update.",
+			"This command can be used to generate a payload to update an existing Scrape Config job or to create a new Scrape Config job.",
+			"To update an existing Scrape Config job, provide the job name and the instance ID of the Argus instance.",
+			"To obtain a default payload to create a new Scrape Config job, run the command with no flags.",
```
vicentepinto98, line 39: We should make more clear in the description that there are fields that **must** be changed in the default payload, such as the **target url**, otherwise the scrape config creation won't work

internal/pkg/services/argus/utils/utils_test.go
```diff
@@ -220,3 +349,87 @@ func TestValidatePlanId(t *testing.T) {
 		})
 	}
 }
+
+func TestMapToUpdateScrapeConfigPayload(t *testing.T) {
```
vicentepinto98, line 353: The other smaller functions that are called from this one should also be tested (apart maybe from the ones that are just a simple assignment to another data structure)

internal/pkg/services/argus/utils/utils.go
vicentepinto98: also, add unit test for it

internal/pkg/services/argus/utils/utils_test.go
```diff
@@ -220,3 +349,87 @@ func TestValidatePlanId(t *testing.T) {
...
+			description: "nil data",
+			resp: &argus.GetScrapeConfigResponse{
+				Data: nil,
+			},
+			isValid: false,
+		},
+		{
+			description: "empty data",
+			resp: &argus.GetScrapeConfigResponse{
+				Data: &argus.Job{},
```
vicentepinto98, line 381: This use case is really weird and probably not happenning, but it makes me think that we could maybe check in the generate-payload command if the updateScrapeConfigPayload is empty and return error, because that's not usable at all (it doesn't even have the fields in the resulting json)

internal/pkg/services/argus/utils/utils_test.go
```diff
@@ -220,3 +349,87 @@ func TestValidatePlanId(t *testing.T) {
...
+				cmp.AllowUnexported(*tt.expectedPayload),
+			)
+			if diff != "" {
+				t.Fatalf("Data does not match: %s", diff)
+			}
+		})
+	}
+}
+
+func TestGetDefaultCreateScrapeConfigPayload(t *testing.T) {
```
vicentepinto98, line 412: This test is not testing much and will add maintenance effort because if the defaults are updated then they also need to be updated here. The function is like a getter, I would remove this test.

Probably what make more sense is to make the `DefaultCreateScrapeConfig` a const instead of a function

internal/pkg/services/argus/utils/utils.go
DiogoFerrao: I actually was doing it this way, the problem with this is that the field in the API Request is optional so it should be set to nil if not used right? 
Because this way we always set a pointer to something in the API request, even if it is a pointer to an empty array. But if setting to nil is not a requirement, then this is much cleaner!

internal/pkg/services/argus/utils/utils.go
DiogoFerrao: The only reason I did it this way is that the conversion code is the same, but I agree that the code is not the clearest. Updated it.

internal/cmd/argus/scrape-configs/generate-payload/generate_payload.go
DiogoFerrao: That's an important topic, but I'm not sure how to do it. The payload provided will work without errors, but with the default target url it is not very useful. However, that is also the case with the jobName and metrics path. Should I document these cases?
Please comment on my suggested change.

internal/pkg/services/argus/utils/utils.go
DiogoFerrao: Yeah, I agree. The original thinking was that the way you suggest I would always return a pointer to a list instead of a nil pointer if it is not set in the initial payload.
I refactored the code overall, should also fix this problem.

internal/cmd/argus/scrape-configs/generate-payload/generate_payload.go
DiogoFerrao: The creation works with the current default payload but isn't very useful with the default target url. I updated the help

internal/pkg/services/argus/utils/utils.go
vicentepinto98: The zero value of a slice is `nil`.
So if there is nothing on the initial payload, we will have a pointer to `nil`, which is the same as a `nil` pointer

internal/pkg/services/argus/utils/utils.go
```diff
@@ -56,6 +75,123 @@ func LoadPlanId(planName string, resp *argus.PlansResponse) (*string, error) {
...
+		mappedConfigs = append(mappedConfigs, mappedConfig)
+	}
+	return &mappedConfigs
+}
+
+func mapStaticConfig(staticConfigs *[]argus.StaticConfigs) *[]argus.UpdateScrapeConfigPayloadStaticConfigsInner {
+	if staticConfigs == nil {
+		return nil
+	}
+	mappedConfigs := make([]argus.UpdateScrapeConfigPayloadStaticConfigsInner, 0)
```
vicentepinto98, line 142: Even with these new functions, I still think it would be more readable to have it `var mappedConfigs []argus.UpdateScrapeConfigPayloadStaticConfigsInner`, and it's the same thing

vicentepinto98 (approved):

PR #234, opened 2024-04-22
Output format none

vicentepinto98:


internal/pkg/print/print.go
```diff
@@ -18,6 +21,8 @@ const (
 	InfoLevel    Level = "info"
 	WarningLevel Level = "warning"
 	ErrorLevel   Level = "error"
+
+	NoneOutputFormat = "none"
```
joaopalet, line 25: This const seems to belong together with `JSONOutputFormat` and `PrettyOutputFormat`, but we can't import from there in this file

joaopalet: But maybe it makes even more sense if the 3 of them are defined here in the `print` package

internal/pkg/print/print_test.go
```diff
@@ -55,16 +64,26 @@ func TestOutputf(t *testing.T) {
 				Verbosity: tt.verbosity,
 			}
 
+			if tt.outputFormatNone {
+				viper.Set(config.OutputFormatKey, NoneOutputFormat)
+			} else {
+				viper.Set(config.OutputFormatKey, "") // needed to prevent conflict with other tests
```
joaopalet, line 70: I think if we use the [`viper.New`](https://pkg.go.dev/github.com/dvln/viper#New) method this won't be needed

internal/pkg/print/print.go
```diff
@@ -112,3 +127,21 @@ func (p *Printer) PromptForConfirmation(prompt string) error {
 	}
 	return fmt.Errorf("max number of wrong inputs")
 }
+
+// Shows the content in the command's stdout using the "less" command
+// If output format is set to none, it does nothing
+func (p *Printer) PagerDisplay(content string) error {
```
joaopalet, line 133: Could we have a simple test for this function also? Similar to the ones for the `Output...` functions?

internal/pkg/pager/pager.go
joaopalet, line 1: Nice, fits well in the `print` package 👍

joaopalet (approved):

PR #235, opened 2024-04-22
Onboard Argus Scrape Config: delete command

DiogoFerrao:


internal/cmd/argus/scrape-configs/delete/delete.go
```diff
@@ -0,0 +1,126 @@
...
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Var(flags.UUIDFlag(), instanceIdFlag, "Instance ID")
+
+	err := flags.MarkFlagsRequired(cmd, instanceIdFlag)
+	cobra.CheckErr(err)
+}
+
+func parseInput(cmd *cobra.Command, inputArgs []string) (*inputModel, error) {
+	clusterName := inputArgs[0]
```
vicentepinto98, line 109:
```suggestion
	jobName := inputArgs[0]
```

internal/cmd/argus/scrape-configs/delete/delete.go
```diff
@@ -0,0 +1,126 @@
...
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	JobName    string
+	InstanceId string
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("delete %s", jobNameArg),
+		Short: "Deletes a Scrape Config job from an Argus instance",
```
vicentepinto98, line 37: This is valid for all scrape config commands: I think it's not needed to start with capital letter and we should have the full word "configuration". Also, not sure if we really need to specify job because also in the Portal it is just identified as scrape configuration.

So I would suggest that we just have "Deletes a scrape configuration from an Argus instance"

docs/stackit_argus_scrape-configs_delete.md
```diff
@@ -0,0 +1,40 @@
+## stackit argus scrape-configs delete
```
vicentepinto98, line 1: the command should be named `stackit argus scrape-config`, in the singular form

vicentepinto98 (approved):

vicentepinto98 (approved):

PR #236, opened 2024-04-22
Onboard Argus Scrape Configs: update command

DiogoFerrao:


internal/cmd/argus/scrape-config/update/update.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("update %s", jobNameArg),
+		Short: "Updates a scrape configuration of an Argus instance",
+		Long: fmt.Sprintf("%s\n%s\n%s",
+			"Updates a scrape configuration of an Argus instance.",
+			"The payload can be provided as a JSON string or a file path prefixed with \"@\".",
+			"See https://docs.api.stackit.cloud/documentation/argus/version/v1#tag/scrape-config/operation/v1_projects_instances_scrapeconfigs_partial_update for information regarding the payload structure.",
```
joaopalet, line 41: I think you meant
```suggestion
			"See https://docs.api.stackit.cloud/documentation/argus/version/v1#tag/scrape-config/operation/v1_projects_instances_scrapeconfigs_update for information regarding the payload structure.",
```
right?

internal/cmd/argus/scrape-config/update/update.go
```diff
@@ -0,0 +1,130 @@
...
+		),
+		Args: args.SingleArg(jobNameArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Update a scrape configuration from Argus instance "xxx", using an API payload sourced from the file "./payload.json"`,
```
joaopalet, line 46: Do you think it's relevant to add `Update a scrape configuration with job name "my-config" ...`?

internal/cmd/argus/scrape-config/update/update.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			_, err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("update scrape config: %w", err)
+			}
+
+			// The API has no status to wait on, so async mode is default
```
joaopalet, line 85: Does the Portal also indicate the update as done immediatelly?

internal/cmd/argus/scrape-config/update/update_test.go
```diff
@@ -0,0 +1,294 @@
...
+		{
+			description: "instance id invalid 2",
+			argValues:   fixtureArgValues(),
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[instanceIdFlag] = "invalid-uuid"
+			}),
+			isValid: false,
+		},
+		{
+			description: "invalid json",
```
joaopalet, line 205: Missing a test case for just the `payload` flag missing

internal/cmd/argus/scrape-config/update/update.go
DiogoFerrao: This is something that was also commented on the Terraform provider, but I looked into it now and the endpoint seems to be synchronous. Removed the comment

joaopalet: If we know the endpoint is synchronous then we can remove the comment, if it's just fast I think we could keep it

DiogoFerrao: The response code is a 202, which would indicate asynchronous. So maybe it's just fast, although the message in the reply states "Successfully updated scrape config".

DiogoFerrao: I'll add the comment back

PR #240, opened 2024-04-23
Onboard Argus scrape config: list and describe commands

DiogoFerrao:


DiogoFerrao:
Describe currently looks like this:
image

internal/cmd/argus/scrape-config/describe/describe.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func outputResult(p *print.Printer, outputFormat string, config *argus.Job) error {
+	switch outputFormat {
+	case globalflags.PrettyOutputFormat:
+
+		saml2Enabled := "Enabled"
+		if config.Params != nil {
+			saml2 := (*config.Params)["saml2"]
+			if len(saml2) > 0 && saml2[0] == "disabled" {
```
vicentepinto98, line 107: what are the other items in the list used for? Is this consistent with the Portal?

internal/cmd/argus/scrape-config/describe/describe.go
```diff
@@ -0,0 +1,181 @@
...
+				saml2Enabled = "Disabled"
+			}
+		}
+
+		targets := []string{}
+		for _, target := range *config.StaticConfigs {
+			targetFmt := ""
+			if target.Labels != nil {
+				// make map prettier
+				for k, v := range *target.Labels {
```
vicentepinto98, line 117: This seems too complex. I would make it simpler, something like this:
```
var targetLabels []string
targetLabelsString = "N/A"
for k, v := range *target.Labels {
    targetLabels = append(targetLabels, fmt.Sprintf("%s:%s", k, v))
}
if targetLabels != nil {
   targetLabels = strings.Join(targetLabels, ",")
}
targetUrls := "N/A"
if target.Targets != nil {
  targetUrls = *target.Targets
}

(...)

targets := fmt.Sprintf("labels: %s\nurls:%s", targetLabelsString, targetUrls)

```

Check what the Portal shows when labels or urls are empty (I only checked for labels)

vicentepinto98 (changes requested):

internal/cmd/argus/scrape-config/describe/describe.go
DiogoFerrao: This is consistent with the Portal, its how I got this information in the first place. I am not aware of any other items using this Params field, but the API also doesn't specify this

internal/cmd/argus/scrape-config/describe/describe.go
DiogoFerrao: For labels it shows the "N/A" you gave, for the urls, at least one is required so it is never shown. (I added an initial value for the string of "N/A" instead of "" for uniformity.
Looks like this now:
image

DiogoFerrao: I am also unsure about the TARGET 1, TARGET 2, etc. since targets are not numbered or named. 
Alternative would be to just print TARGET, TARGET, etc.
Also just print TARGETS in the first one and then have the first column be empty.
Any suggestions?

vicentepinto98: For now we could make it TARGET #1, TARGET #2

Then we can implement a way to merge rows only on the first column, keeping the separator on the others

PR #249, opened 2024-04-25
Implement user confirmation for re-authenticating

GokceGK:


internal/pkg/auth/user_login.go
```diff
@@ -36,7 +38,14 @@ type User struct {
 }
 
 // AuthorizeUser implements the PKCE OAuth2 flow.
-func AuthorizeUser() error {
+func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
+	if isReauthentication {
+		err := p.PromptForConfirmation("Your session has expired! Do you want to login again? ")
```
joaopalet, line 43: The acceptance criteria in this case would be that the user only needs to press Enter, so the prompt would be something like `Your session has expired, press Enter to login again...`

So we would need to implement a new function, e.g. `PromptForEnter`

joaopalet (commented):
All the rest looks great

vicentepinto98:
I don't think it's worth adding this dependency for this use case. The original implementation was fairly simple

joaopalet (approved):

PR #250, opened 2024-04-25
Onboard Load Balancer generate-payload command

GokceGK:


internal/cmd/load-balancer/generate-payload/generate_payload.go
```diff
@@ -0,0 +1,124 @@
...
+		),
+		Args: args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Generate a payload, and adapt it with custom values for the different configuration options`,
+				`$ stackit load-balancer generate-payload > ./payload.json`,
+				`<Modify payload in file, if needed>`,
+				`$ stackit load-balancer create --payload @./payload.json`),
+			examples.NewExample(
+				`Generate a payload with values of a load balancer, and adapt it with custom values for the different configuration options`,
```
vicentepinto98, line 46:
```suggestion
				`Generate a payload with values of an existing load balancer, and adapt it with custom values for the different configuration options`,
```

internal/cmd/load-balancer/generate-payload/generate_payload.go
```diff
@@ -0,0 +1,124 @@
...
+			var payload *loadbalancer.CreateLoadBalancerPayload
+			if model.InstanceName == nil {
+				payload = lbUtils.GetDefaultPayload()
+			} else {
+				req := buildRequest(ctx, model, apiClient)
+				resp, err := req.Execute()
+				if err != nil {
+					return fmt.Errorf("read load balancer: %w", err)
+				}
+				payload = &loadbalancer.CreateLoadBalancerPayload{
```
vicentepinto98, line 73: As discussed before, it would be cleaner to create here an `UpdateLoadBalancerPayload`. And then create two separate functions: `outputCreateResult`, `outputUpdateResult`

internal/cmd/load-balancer/generate-payload/generate_payload.go
```diff
@@ -0,0 +1,124 @@
...
+
+			return outputResult(p, payload)
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().StringP(instanceNameFlag, "n", "", "If set, generates the payload with the current values of the given instance. If unset, generates the payload with empty values")
```
vicentepinto98, line 93:
```suggestion
	cmd.Flags().StringP(instanceNameFlag, "n", "", "If set, generates the payload with the current values of the given load balancer. If unset, generates the payload with empty values")
```

internal/pkg/services/load-balancer/utils/utils_test.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+		},
+	}
+	for _, mod := range mods {
+		mod(payload)
+	}
+	return payload
+}
+
+func TestGetDefaultPayload(t *testing.T) {
```
vicentepinto98, line 92: this unit test is basically not testing anything and it's just adding maintenance effort if some value is updated on the default payload. It makes more sense to make the default payload a const of the generate-payload command and remove this `utils` pkg entirely

internal/pkg/services/load-balancer/utils/utils.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/loadbalancer"
+)
+
+type LoadBalancerClient interface {
+}
+
+func GetDefaultPayload() *loadbalancer.CreateLoadBalancerPayload {
```
vicentepinto98, line 11: make this a const (or multiple ones) of the `generatepayload` pkg instead of these getters in a separate `utils` pkg. See the `argus scrape-config generate-payload` implementation for a similar case

internal/cmd/load-balancer/generate-payload/generate_payload.go
vicentepinto98: See the implementation of `argus scrape-config generate-payload` for a similar case

GokceGK: Splitted the payloads

internal/pkg/services/load-balancer/utils/utils.go
GokceGK: moved the getter results in a var in generate payload, utils have been removed

internal/cmd/load-balancer/generate-payload/generate_payload.go
```diff
@@ -0,0 +1,216 @@
...
+		InstanceName:    instanceName,
+	}, nil
+}
+
+func buildRequest(ctx context.Context, model *inputModel, apiClient *loadbalancer.APIClient) loadbalancer.ApiGetLoadBalancerRequest {
+	req := apiClient.GetLoadBalancer(ctx, model.ProjectId, *model.InstanceName)
+	return req
+}
+
+func outputCreateResult(p *print.Printer, payload *loadbalancer.CreateLoadBalancerPayload) error {
```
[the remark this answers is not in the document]
GokceGK: Linter throws error if I do not use pointer in these functions.

vicentepinto98 (approved):

PR #258, opened 2024-04-29
Add debug logs to the CLI

DiogoFerrao:


internal/pkg/config/config.go
```diff
@@ -145,3 +145,16 @@ func setConfigDefaults() {
 	viper.SetDefault(ServiceAccountCustomEndpointKey, "")
 	viper.SetDefault(SKECustomEndpointKey, "")
 }
+
+// GetConfigKeys returns a map of the config keys and their values
+// Ignores empty values
+func GetConfigKeys() map[string]string {
```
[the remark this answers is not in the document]
DiogoFerrao: I missed that function, thanks

internal/pkg/print/print.go
```diff
@@ -147,3 +147,8 @@ func (p *Printer) PagerDisplay(content string) error {
 	}
 	return nil
 }
+
+// Returns True if the verbosity level is set to Debug, False otherwise.
+func (p *Printer) IsVerbosityDebug() bool {
```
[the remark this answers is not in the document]
DiogoFerrao: Thanks, done 😁

internal/pkg/print/debug.go
```diff
@@ -0,0 +1,91 @@
...
+// The keys are ordered alphabetically to make the output deterministic.
+func BuildDebugStrFromMap(inputMap map[string]any) string {
+	// Sort the keys to make the output deterministic
+	keys := make([]string, 0, len(inputMap))
+	for key := range inputMap {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	var builder strings.Builder
```
vicentepinto98, line 41: I would do this in a simpler way by just using string formatting and string joining instead of the use of `strings.Builder`, the code is not the most readable

internal/cmd/root.go
```diff
@@ -45,6 +46,13 @@ func NewRootCmd(version, date string, p *print.Printer) *cobra.Command {
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			p.Cmd = cmd
 			p.Verbosity = print.Level(globalflags.Parse(p, cmd).Verbosity)
+
+			argsString := print.BuildDebugStrFromSlice(os.Args)
+			p.Debug(print.DebugLevel, "arguments entered to the CLI: %s", argsString)
```
vicentepinto98, line 51:
```suggestion
			p.Debug(print.DebugLevel, "arguments: %s", argsString)
```

To reduce noise on the logs

internal/cmd/argus/grafana/describe/describe.go
```diff
@@ -95,11 +95,22 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
 		HidePassword:    flags.FlagToBoolValue(p, cmd, hidePasswordFlag),
-	}, nil
+	}
+
+	if p.IsVerbosityDebug() {
+		modelStr, err := print.BuildDebugStrFromInputModel(model)
+		if err != nil {
+			p.Debug(print.ErrorLevel, "convert model to string for debugging: %v", err)
+		} else {
+			p.Debug(print.DebugLevel, "input model: %s", modelStr)
```
vicentepinto98, line 109: "input model" might not be the most obvious name for users. More clear would be something like "parsed input values"

internal/pkg/print/debug.go
```diff
@@ -38,8 +38,7 @@ func BuildDebugStrFromMap(inputMap map[string]any) string {
 	}
 	sort.Strings(keys)
 
-	var builder strings.Builder
-	builder.WriteString("[")
+	result := "["
```
vicentepinto98, line 41: In my opinion, you can make it even more readable like this:

```
var keyValues []string
for _, key := range keys {
		value := inputMap[key]
		if isEmpty(value) {
			continue
		}
		keyValues = append(keyValues,  fmt.Sprintf("%s: %v", key, value))
}

result := strings.Join(keyValues, ",")
return fmt.Sprintf("[%s]", result)
```

No `if...else` block and string concatenation needed

joaopalet (commented):
Could you also update the command code template in `CONTRIBUTION.md`?

internal/cmd/config/set/set.go
```diff
@@ -160,10 +160,21 @@ func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
...
 
-	return &inputModel{
+	model := inputModel{
 		SessionTimeLimit: sessionTimeLimit,
 		ProjectIdSet:     projectIdSet,
-	}, nil
+	}
+
+	if p.IsVerbosityDebug() {
+		modelStr, err := print.BuildDebugStrFromInputModel(model)
```
joaopalet, line 169: In this command all other flags are parsed through Viper, maybe we could handle it differently here by also showing the binded Viper values

DiogoFerrao: I think this is already shown when printing the config keys

joaopalet: True, the flags are already binded at that point

joaopalet (approved):

PR #260, opened 2024-04-29
Onboard first set of Load Balancer functionality

vicentepinto98:


joaopalet:
Do we still need this PR? We could merge all changes to main at once when all the stories are accepted, right?

vicentepinto98:
> Do we still need this PR? We could merge all changes to main at once when all the stories are accepted, right?

Yes but we always need to open a PR, so it's the same as having it open from the start no?

joaopalet:
> Yes but we always need to open a PR, so it's the same as having it open from the start no?

Ah ok, I was just checking if you were planning to merge this PR with some of the changes

joaopalet (approved):

PR #264, opened 2024-04-29
LoadBalancer observability-credentials add command

vicentepinto98:


internal/cmd/load-balancer/observability-credentials/delete/delete_test.go
```diff
@@ -0,0 +1,213 @@
...
+		{
+			description: "project id invalid 2",
+			argValues:   fixtureArgValues(),
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[projectIdFlag] = "invalid-uuid"
+			}),
+			isValid: false,
+		},
+		{
+			description: "credentials id invalid 1",
```
GokceGK, line 126:
```suggestion
			description: "credentials reference is invalid 1",
```

internal/cmd/load-balancer/observability-credentials/observability-credentials.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "observability-credentials",
+		Short:   "Provides functionality for Load Balancer observability credentials",
+		Long:    `Provides functionality for Load Balancer observability credentials. These commands can be used to store and update existing credentials, which are valid to be used for Load Balancer Observability. This means, e.g. when using Argus, that credentials first must be created for that Argus instance (by using "stackit argus credentials create") and then can be managed for a Load Balancer by using the commands in this group.`,
```
GokceGK, line 20:
```suggestion
		Long:    `Provides functionality for Load Balancer observability credentials. These commands can be used to store and update existing credentials, which are valid to be used for Load Balancer observability. This means, e.g. when using Argus, first of all these credentials must be created for that Argus instance (by using "stackit argus credentials create") and then can be managed for a Load Balancer by using the commands in this group.`,
```

internal/cmd/load-balancer/observability-credentials/delete/delete_test.go
GokceGK: For other unit tests, you used "credentials ref invalid 1", lets use it here also

internal/cmd/load-balancer/observability-credentials/observability-credentials.go
GokceGK: if you apply these changes, docs must be updated.

GokceGK (approved):

PR #265, opened 2024-04-30
Fixes to lb list and describe

vicentepinto98:


internal/cmd/load-balancer/describe/describe.go
```diff
@@ -160,37 +158,17 @@ func renderListeners(listeners []loadbalancer.Listener) string {
 	for i := range listeners {
 		listener := listeners[i]
 		table.AddRow(*listener.Name, *listener.Port, *listener.Protocol, *listener.TargetPool)
+		table.AddSeparator()
```
joaopalet, line 161: If these tables are horizontal, I don't think we need the separators

joaopalet: My point before was that when we have auto-merging on columns we usually have separator between rows that are not merged (e.g. see `stackit project member list`)

joaopalet: That raises another point, do these 2 last tables need to be horizontal? Because the first one for the LB is vertical

internal/cmd/load-balancer/list/list.go
```diff
@@ -133,6 +134,7 @@ func outputResult(p *print.Printer, outputFormat string, loadBalancers []loadbal
 		for i := range loadBalancers {
 			l := loadBalancers[i]
 			table.AddRow(*l.Name, *l.Status, *l.ExternalAddress, len(*l.Listeners), len(*l.TargetPools))
+			table.AddSeparator()
```
joaopalet, line 137: Same here, we don't usually have separators for horizontal tables

joaopalet: image

internal/cmd/load-balancer/describe/describe.go
joaopalet: Nevermind, those can actually have multiple rows

internal/cmd/load-balancer/list/list.go
joaopalet: Missing this one
```suggestion
```

joaopalet (approved):

PR #268, opened 2024-04-30
Onboard load balancer create & update commands

GokceGK:


internal/cmd/load-balancer/generate-payload/generate_payload.go
```diff
@@ -152,13 +150,37 @@ func NewCmd(p *print.Printer) *cobra.Command {
...
 				ExternalAddress: resp.ExternalAddress,
-				Listeners:       resp.Listeners,
+				Listeners:       &listeners,
 				Name:            resp.Name,
 				Networks:        resp.Networks,
 				Options:         resp.Options,
-				PrivateAddress:  resp.PrivateAddress,
 				TargetPools:     resp.TargetPools,
 				Version:         resp.Version,
 			}
```
joaopalet, line 186: I would abstract this logic in a function that can be unit tested

joaopalet: Also, can't we just set the invalid fields to `nil` instead of recreating each listener?

internal/cmd/load-balancer/create/create.go
```diff
@@ -0,0 +1,146 @@
...
+			examples.NewExample(
+				`Create a load balancer using an API payload sourced from the file "./payload.json"`,
+				"$ stackit load-balancer create --payload @./payload.json"),
+			examples.NewExample(
+				`Create a load balancer using an API payload provided as a JSON string`,
+				`$ stackit load-balancer create --payload "{...}"`),
+			examples.NewExample(
+				`Generate a payload with default values, and adapt it with custom values for the different configuration options`,
+				`$ stackit load-balancer generate-payload > ./payload.json`,
+				`<Modify payload in file, if needed>`,
```
joaopalet, line 58:
```suggestion
				`<Modify payload in file>`,
```

internal/cmd/load-balancer/create/create.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+
+			// Configure API client
+			apiClient, err := client.ConfigureClient(p)
+			if err != nil {
+				return err
+			}
+
+			if !model.AssumeYes {
+				prompt := fmt.Sprintf("Are you sure you want to create load balancer %q?", *model.Payload.Name)
```
joaopalet, line 75:
```suggestion
				prompt := fmt.Sprintf("Are you sure you want to create a load balancer for project %q?", projectLabel)
```

internal/cmd/load-balancer/generate-payload/generate_payload.go
joaopalet, line 1: The `instanceNameFlag = "instance-name"` could be renamed to `loadBalancerFlag = "lb-name"` right?

vicentepinto98:
Debug logs should be added, similar to what was done here https://github.com/stackitcloud/stackit-cli/pull/258

internal/cmd/load-balancer/generate-payload/generate_payload.go
GokceGK: changed to loadBalancerNameFlag

internal/cmd/load-balancer/generate-payload/generate_payload.go
GokceGK: Implemented setting listener.name to nil in a new function, added unit test.

internal/cmd/load-balancer/generate-payload/generate_payload_test.go
```diff
@@ -190,3 +192,68 @@ func TestBuildRequest(t *testing.T) {
 		})
 	}
 }
+
+func TestToModifyListeners(t *testing.T) {
```
joaopalet, line 196:
```suggestion
func TestModifyListeners(t *testing.T) {
```

joaopalet: Maybe the test could contain at least two listeners just to check that the name is being removed from all

GokceGK: changed function name and add one more listener

joaopalet (approved):

PR #271, opened 2024-05-02
Add HTTP Request and Response to debug logs

DiogoFerrao:


internal/pkg/print/debug.go
```diff
@@ -3,10 +3,17 @@ package print
...
+	"io"
+	"net/http"
+	"slices"
 	"sort"
 	"strings"
+
+	"github.com/stackitcloud/stackit-sdk-go/core/config"
 )
 
+var defaultDebugIncludeHeaders = []string{"Accept", "Content-Type", "Content-Length", "User-Agent", "Date", "Referrer-Policy"}
```
vicentepinto98, line 15:
```suggestion
var defaultHTTPHeaders = []string{"Accept", "Content-Type", "Content-Length", "User-Agent", "Date", "Referrer-Policy"}
```

internal/pkg/print/debug.go
```diff
@@ -44,6 +51,23 @@ func BuildDebugStrFromMap(inputMap map[string]any) string {
 		if isEmpty(value) {
 			continue
 		}
+
+		if valueMap, ok := value.(map[string]any); ok {
```
vicentepinto98, line 55: can't we make this a type switch? See https://go.dev/tour/methods/16

internal/pkg/services/argus/client/client.go
```diff
@@ -29,6 +29,12 @@ func ConfigureClient(p *print.Printer) (*argus.APIClient, error) {
 		cfgOptions = append(cfgOptions, sdkConfig.WithEndpoint(customEndpoint))
 	}
 
+	if p.IsVerbosityDebug() {
+		cfgOptions = append(cfgOptions,
+			sdkConfig.WithMiddleware(print.RequestResponseCapturer(p, []string{})),
```
vicentepinto98, line 34: I think you can pass `nil` here instead of `[]string{}`

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
 	return fmt.Sprintf("[%s]", sliceStr)
 }
 
+// buildHeaderMap converts a map to a user-friendly string representation.
+// This function also filters the headers based on the includeHeaders parameter.
+// If includeHeaders is empty, the default header filters are used.
+func buildHeaderMap(headers http.Header, includeHeaders []string) map[string]any {
```
vicentepinto98, line 87: why return `map[string]any` if we are always constructing a `map[string]string`?

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
...
+	headersMap := make(map[string]any)
+	for key, values := range headers {
+		headersMap[key] = strings.Join(values, ", ")
+	}
+
+	var headersToInclude []string
+
+	if len(includeHeaders) == 0 {
```
vicentepinto98, line 95: it's simpler like this
```
headersToInclude := defaultDebugIncludeHeaders
if len(includeHeaders) != 0 {
    headersToInclude = includeHeaders
}
```

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
...
+		headersToInclude = defaultDebugIncludeHeaders
+	} else {
+		headersToInclude = includeHeaders
+	}
+
+	for key := range headersMap {
```
vicentepinto98, line 101: simpler like this

```
for key := range headersMap {
		if !slices.Contains(headersToInclude, key) {
			delete(headersMap, key)
		}
	}
```

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
...
+	}
+	var bodyMap map[string]any
+	if len(body) != 0 {
+		if err := json.Unmarshal(body, &bodyMap); err != nil {
+			return []string{status, headers}, fmt.Errorf("unmarshal request body: %w", err)
+		}
+	}
+
+	// restore body
+	req.Body = io.NopCloser(strings.NewReader(string(body)))
```
vicentepinto98, line 142: is there no way of reading the body without emptying it? This seems dangerous.

Maybe you can make a copy of the request (or the body) and use that one

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
...
+func RequestResponseCapturer(p *Printer, includeHeaders []string) config.Middleware {
+	return func(rt http.RoundTripper) http.RoundTripper {
+		return &roundTripperWithCapture{rt, p, includeHeaders}
+	}
+}
+
+type roundTripperWithCapture struct {
+	transport    http.RoundTripper
+	p            *Printer
+	debugHeaders []string
```
vicentepinto98, line 199:
```suggestion
	debugHTTPHeaders []string
```

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
...
+}
+
+func (rt roundTripperWithCapture) RoundTrip(req *http.Request) (*http.Response, error) {
+	reqStr, err := BuildDebugStrFromHTTPRequest(req, rt.debugHeaders)
+	if err != nil {
+		rt.p.Debug(ErrorLevel, "printing request to debug logs: %v", err)
```
vicentepinto98, line 205:
```suggestion
		rt.p.Debug(ErrorLevel, "printing HTTP request to debug logs: %v", err)
```

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
...
+	}
+	for _, line := range reqStr {
+		rt.p.Debug(DebugLevel, line)
+	}
+	resp, err := rt.transport.RoundTrip(req)
+	defer func() {
+		if err == nil {
+			respStrSlice, err := BuildDebugStrFromHTTPResponse(resp, rt.debugHeaders)
+			if err != nil {
+				rt.p.Debug(ErrorLevel, "printing response to debug logs: %v", err)
```
vicentepinto98, line 215:
```suggestion
				rt.p.Debug(ErrorLevel, "printing HTTP response to debug logs: %v", err)
```

internal/pkg/print/debug.go
```diff
@@ -57,6 +81,147 @@ func BuildDebugStrFromSlice(inputSlice []string) string {
...
+	if err != nil {
+		rt.p.Debug(ErrorLevel, "printing request to debug logs: %v", err)
+	}
+	for _, line := range reqStr {
+		rt.p.Debug(DebugLevel, line)
+	}
+	resp, err := rt.transport.RoundTrip(req)
+	defer func() {
+		if err == nil {
+			respStrSlice, err := BuildDebugStrFromHTTPResponse(resp, rt.debugHeaders)
```
vicentepinto98, line 213:
```suggestion
			respStrSlice, tempErr := BuildDebugStrFromHTTPResponse(resp, rt.debugHeaders)
```
To not overwrite the main function error

internal/pkg/print/debug.go
DiogoFerrao: The point is I'm trying to use the generic `BuildDebugStrFromMap` as much as possible, and that receives a map[string]any. So returning that type here makes it so that I can use the return to call that function

internal/pkg/print/debug.go
DiogoFerrao: I don't think there is a way around it. I have refactored the implementation slightly to match the go std library's implementation, as described in the comments above the new functions.

internal/pkg/print/debug.go
vicentepinto98: But you can use `map[string]string` as an argument that is of type `map[string]any`

internal/pkg/print/debug.go
```diff
@@ -38,6 +40,9 @@ func BuildDebugStrFromInputModel(model any) (string, error) {
 // The string representation is in the format: [key1: value1, key2: value2, ...]
 // The keys are ordered alphabetically to make the output deterministic.
 func BuildDebugStrFromMap(inputMap map[string]any) string {
+	if inputMap == nil {
```
vicentepinto98, line 43: :+1:

internal/pkg/print/debug.go
```diff
@@ -162,25 +177,110 @@ func BuildDebugStrFromHTTPResponse(resp *http.Response, includeHeaders []string)
...
-	body, err := io.ReadAll(resp.Body)
+	body := fmt.Sprintf("response body: %s", BuildDebugStrFromMap(bodyMap))
+
+	return []string{status, headers, body}, nil
+}
+
+// dumpRespBody reads the response body and returns a string representation of it.
+// Based on code from httputil package
+// https://pkg.go.dev/net/http/httputil#DumpResponse
+func dumpRespBody(resp *http.Response) (map[string]any, error) {
```
vicentepinto98, line 195: Why not use the `DumpRequest` and `DumpResponse` methods directly from the standard library `httputils` pkg? 
If needed, then make a wrapper around those, but I don't see why we need to copy the code here and maintain it ourselves

DiogoFerrao: The DumpResponse/Request methods return the entire Response/Request in a format that is not consistent with the other debug logs from the CLI. They also include a lot of formatting characters (\n,\r,...) that don't look good in slog and make it practically impossible to filter the noise (additional headers, etc).
The other problem is that their auxiliary functions (drainBody), which would be particularly useful here, are private.

internal/pkg/print/debug.go
DiogoFerrao: I don't think so
image

DiogoFerrao: Unless I am missing something

vicentepinto98: You're right, let's leave it like this

internal/pkg/print/debug.go
vicentepinto98: as discussed offline, let's use the `drainBody` function

internal/pkg/print/debug.go
vicentepinto98: as discussed offline, let's use the initial implementation with the `ReadAll` and `drainBody`

vicentepinto98: No need to have `DumpRequest` and `DumpResponse` methods copied here

vicentepinto98 (dismissed):

vicentepinto98 (approved):

PR #272, opened 2024-05-03
add debug logs to the authentication flow

DiogoFerrao:


internal/pkg/auth/user_login.go
```diff
@@ -168,9 +177,12 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 	if err != nil {
 		return fmt.Errorf("open browser to URL %s: %w", authorizationURL, err)
 	}
+	p.Debug(print.DebugLevel, "opening browser for authentication")
```
vicentepinto98, line 180: these two logs should go above the call to `openBrowser`, otherwise if there is an error we don't see them

internal/pkg/auth/user_login.go
```diff
@@ -98,13 +99,17 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
...
 		// Trade the authorization code and the code verifier for access and refresh tokens
 		accessToken, refreshToken, err := getUserAccessAndRefreshTokens(authDomain, clientId, codeVerifier, code, redirectURL)
 		if err != nil {
 			errServer = fmt.Errorf("retrieve tokens: %w", err)
 			return
 		}
 
+		p.Debug(print.DebugLevel, "received response from the authentication server")
+
 		sessionExpiresAtUnix, err := getStartingSessionExpiresAtUnix()
```
vicentepinto98, line 113: also add here a debug log  "current session will expire at..."

internal/pkg/auth/user_token_flow.go
```diff
@@ -59,6 +60,7 @@ func (utf *userTokenFlow) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	if !accessTokenValid {
+		utf.printer.Debug(print.DebugLevel, "user access token expired, reauthenticating...")
```
vicentepinto98, line 63:
```suggestion
		utf.printer.Debug(print.DebugLevel, "user access token is not valid, reauthenticating...")
```
might be something else other than expiration

vicentepinto98 (approved):

PR #276, opened 2024-05-03
Add tint package for colors in slog debug logs

DiogoFerrao:


DiogoFerrao:
image

GokceGK (approved):

vicentepinto98:
> Package tint implements a zero-dependency [slog.Handler](https://pkg.go.dev/log/slog#Handler)

nice :+1:

internal/pkg/print/print.go
```diff
@@ -37,7 +38,7 @@ type Printer struct {
 
 // Creates a new printer, including setting up the default logger.
 func NewPrinter() *Printer {
-	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{AddSource: false, Level: slog.LevelDebug}))
+	logger := slog.New(tint.NewHandler(os.Stderr, &tint.Options{AddSource: false, Level: slog.LevelDebug}))
```
vicentepinto98, line 41: Maybe we need to also check if Windows is being used and adapt like described here? https://github.com/lmittmann/tint?tab=readme-ov-file#windows-support

vicentepinto98: check the `openBrowser` method for a `switch...case` on OS

DiogoFerrao: Great catch, thank you! I investigated the [go-colorable](https://github.com/mattn/go-colorable) they suggest in the documentation and its fairly simple, should be ok to add.
There's no need for a `switch...case` though, the go-colorable package already handles other OSes by [not doing anything](https://github.com/mattn/go-colorable/blob/master/colorable_others.go) if it isn't Windows.

DiogoFerrao: It's also used by the GitHub cli, so we have more assurance there

vicentepinto98 (approved):

PR #283, opened 2024-05-06
Implement `lb target-pool remove-target` command

joaopalet:


internal/cmd/load-balancer/target-pool/remove-target/remove_target.go
```diff
@@ -0,0 +1,145 @@
...
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("remove-target %s", loadBalancerNameArg),
+		Short: "Removes a target from a target pool",
+		Long:  "Removes a target from a target pool.",
+		Args:  args.SingleArg(loadBalancerNameArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Remove target with IP 1.2.3.4 from target pool "my-target-pool" of load balancer with name "my-load-balancer"`,
+				"$ stackit load-balancer target-pool remove-target my-load-balancer --target-pool-name my-target-pool --ip 1.2.3.4"),
```
GokceGK, line 44: Generally, after remove/delete commands, the resource identifier is passed. Here we are passing the load balancer name. It is a bit confusing. 

It is also same for the add command. Would it not be more understandable if we give lb-name also as flag?

joaopalet: This question also came to my mind. I did not go deep into the args/flags of the commands when discussing with the team

What would you suggest? Have the target-pool name as argument or no argument at all?

vicentepinto98: Target-pool name as argument is the most uniform with the rest of the CLI, if it is the unique identifier of the resource

GokceGK (approved):

PR #284, opened 2024-05-06
Document config file location in command

DiogoFerrao:
Includes a `go mod tidy` that was missed in the previous PR

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -17,7 +17,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "config",
 		Short: "Provides functionality for CLI configuration options",
-		Long:  fmt.Sprintf("%s\n%s\n%s\n%s\n%s\n%s", "Provides functionality for CLI configuration options", "The configuration is stored in a file in the user's config directory, which is OS dependent.", "Windows: %APPDATA%\\stackit", "Linux: $XDG_CONFIG_HOME/stackit or $HOME/.config/stackit", "macOS: $HOME/Library/Application Support/stackit", "The configuration file is named `cli-config.json` and is created automatically when setting a configuration option."),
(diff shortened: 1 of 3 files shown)


internal/cmd/config/config.go
```diff
@@ -15,7 +17,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "config",
 		Short: "Provides functionality for CLI configuration options",
-		Long:  "Provides functionality for CLI configuration options.",
+		Long:  fmt.Sprintf("%s\n%s\n%s\n%s\n%s\n%s", "Provides functionality for CLI configuration options", "The configuration is stored in a file in the user's config directory, which is OS dependent.", "Windows: %APPDATA%\\stackit", "Linux: $XDG_CONFIG_HOME/stackit or $HOME/.config/stackit", "macOS: $HOME/Library/Application Support/stackit", "The configuration file is named `cli-config.json` and is created automatically when setting a configuration option."),
```
vicentepinto98, line 20:
```suggestion
		Long:  fmt.Sprintf("%s\n%s\n%s\n%s\n%s\n%s", "Provides functionality for CLI configuration options", "The configuration is stored in a file in the user's config directory, which is OS dependent.", "Windows: %APPDATA%\\stackit", "Linux: $XDG_CONFIG_HOME/stackit or $HOME/.config/stackit", "macOS: $HOME/Library/Application Support/stackit", "The configuration file is named `cli-config.json` and is created automatically in your first CLI run."),
```

vicentepinto98, line 20: > "Linux: $XDG_CONFIG_HOME/stackit or $HOME/.config/stackit"

Which one takes precedence? it's not clear

DiogoFerrao: By default they are the same, it was to clarify if someone does not know what `$XDG_CONFIG_HOME` is, but maybe it adds unnecessary complexity

DiogoFerrao: But the only thing used is the `$XDG_CONFIG_HOME`, so I'll remove the other one to avoid confusion

vicentepinto98 (approved):

PR #292, opened 2024-05-07
Onboard load-balancer target-pool describe

DiogoFerrao:


internal/pkg/services/load-balancer/utils/utils.go
```diff
@@ -34,13 +34,23 @@ func GetLoadBalancerTargetPool(ctx context.Context, apiClient LoadBalancerClient
...
 
-	for _, targetPool := range *resp.TargetPools {
+	targetPool := FindLoadBalancerTargetPoolByName(resp.TargetPools, targetPoolName)
+	if targetPool == nil {
+		return nil, fmt.Errorf("target pool not found")
+	}
+	return targetPool, nil
+}
+
+func FindLoadBalancerTargetPoolByName(targetPools *[]loadbalancer.TargetPool, targetPoolName string) *loadbalancer.TargetPool {
```
joaopalet, line 44: I think the function signature doesn't need to take a `*[]loadbalancer.TargetPool`, the argument could simply be a `[]loadbalancer.TargetPool`

internal/cmd/load-balancer/target-pool/describe/describe.go
```diff
@@ -0,0 +1,189 @@
...
+				return err
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("read load balancer: %w", err)
+			}
+
```
joaopalet, line 68: I would also call `FindLoadBalancerTargetPoolByName` here to get the target pool, and then pass it to output result only if it is not `nil`

internal/cmd/load-balancer/target-pool/describe/describe.go
```diff
@@ -0,0 +1,189 @@
...
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		return outputResultAsTable(p, model.TargetPoolName, loadBalancer)
+	}
+}
+
+func outputResultAsTable(p *print.Printer, targetPoolName string, loadBalancer *loadbalancer.LoadBalancer) error {
+	targetPool := utils.FindLoadBalancerTargetPoolByName(loadBalancer.TargetPools, targetPoolName)
```
joaopalet, line 130: Because currently this `targetPool` could be `nil` here, and in this case we shouldn't output anything and should return an error

internal/cmd/load-balancer/target-pool/describe/describe.go
```diff
@@ -0,0 +1,189 @@
...
+		if targetPool.ActiveHealthCheck.Timeout != nil {
+			healthCheckTimeout = *targetPool.ActiveHealthCheck.Timeout
+		}
+	}
+
+	targets := "-"
+	if targetPool.Targets != nil {
+		var targetsSlice []string
+		for _, target := range *targetPool.Targets {
+			targetStr := fmt.Sprintf("%s: %s", *target.DisplayName, *target.Ip)
```
joaopalet, line 160: The Portal shows target as
```suggestion
			targetStr := fmt.Sprintf("%s (%s)", *target.DisplayName, *target.Ip)
```

I think that might be a bit clearer

internal/cmd/load-balancer/target-pool/describe/describe.go
```diff
@@ -0,0 +1,189 @@
...
+	table.AddSeparator()
+	table.AddRow("HEALTH CHECK TIMEOUT", healthCheckTimeout)
+	table.AddSeparator()
+	table.AddRow("HEALTH CHECK UNHEALTHY THRESHOLD", healthCheckUnhealthyThreshold)
+	table.AddSeparator()
+	table.AddRow("HEALTH CHECK HEALTHY THRESHOLD", healthCheckHealthyThreshold)
+	table.AddSeparator()
+	table.AddRow("TARGET PORT", *targetPool.TargetPort)
+	table.AddSeparator()
+	table.AddRow("TARGETS", targets)
```
joaopalet, line 181: Also I would move this two to after the `NAME`, as in the Portal

internal/cmd/load-balancer/target-pool/describe/describe.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+}
+
+func outputResultAsTable(p *print.Printer, targetPoolName string, loadBalancer *loadbalancer.LoadBalancer) error {
+	targetPool := utils.FindLoadBalancerTargetPoolByName(loadBalancer.TargetPools, targetPoolName)
+
+	sessionPersistence := "None"
+	if targetPool.SessionPersistence != nil && targetPool.SessionPersistence.UseSourceIpAddress != nil && *targetPool.SessionPersistence.UseSourceIpAddress {
+		sessionPersistence = "Use Source IP"
+	}
```
joaopalet, line 135: :+1:

internal/cmd/load-balancer/target-pool/describe/describe.go
```diff
@@ -0,0 +1,189 @@
...
+	if targetPool.Targets != nil {
+		var targetsSlice []string
+		for _, target := range *targetPool.Targets {
+			targetStr := fmt.Sprintf("%s: %s", *target.DisplayName, *target.Ip)
+			targetsSlice = append(targetsSlice, targetStr)
+		}
+		targets = strings.Join(targetsSlice, "\n")
+	}
+
+	table := tables.NewTable()
```
joaopalet, line 166: The Portal also shows the listener that is attached to the target pool, that is in a different field of the LB (`listeners`, which has a `targetPool` field to reference the target pool name of the target pool it is attached to)

joaopalet: Since we need to fetch the whole LB already, we would get this for free so I would also add it, we would just need one more util function to extract it

internal/cmd/load-balancer/target-pool/describe/describe.go
DiogoFerrao: Great point, I generally agree but it's probably better to do it inside the outputResult, otherwise the arguments for the function are repetive, since we would have: `outputResult(printer, model, targetPool, listener, loadBalancer)`

But I am also open to this way

joaopalet: Hm true, and there's the question regarding the JSON response again

But we don't need to return the whole LB in the response, in this case we could return the target pool object with a listener field inside. I think you have already done something similar for some other command, right?

joaopalet: I also think it's more clear in this case than just returning the whole LB, which is not what the user is expecting if we just wants to check a single target pool we identified by name

DiogoFerrao: Also a good point. I agree, implemented it now

joaopalet (approved):

PR #298, opened 2024-05-08
Implement YAML output format

GokceGK:


joaopalet (commented):
Could you also update the template in CONTRIBUTION.md?

joaopalet (commented):
The YAML output is missing in 2 files I think:
- `project/role/list`
- `organization/role/list`

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/stackitcloud/stackit-cli
 go 1.22
 
 require (
+	github.com/ghodss/yaml v1.0.0
```
joaopalet, line 6: This one seems to work well, but did you try https://github.com/goccy/go-yaml?

They seem so have around the same number of stars and forks, but https://github.com/ghodss/yaml hasn't been received any update since 2017

GokceGK: Changed used library

joaopalet (approved):

Issue #154, opened 2024-03-12
Allow specifying custom pager


Kumm-Kai:
Currently, the `less` pager is assumed to exist in the users `PATH`. The STACKIT CLI should allow users to specify their preferred pager using the `PAGER` environment variable.

Additionally, the STACKIT CLI should fall back to not paging at all, if the `less` command is not found in the user's environment.

GokceGK:
Hi @Kumm-Kai, 

with the latest STACKIT CLI release [v0.6.0](https://github.com/stackitcloud/stackit-cli/releases/tag/v0.6.0), it is now possible to specify a custom pager.

PR #299, opened 2024-05-08
Add custom pager handling

GokceGK:
Related Github issue https://github.com/stackitcloud/stackit-cli/issues/154


README.md
```diff
@@ -129,6 +129,26 @@ stackit config list
...
 
+## Customization
+
+### Using **less** as a pager
+
+STACKIT CLI uses the `less` as default pager. 
+When using `less` as a pager, STACKIT CLI will automatically pass following options
+- -F, --quit-if-one-screen - Less will automatically exit if the entire file can be displayed on the first screen.
+- -S, --chop-long-lines - Lines longer than the screen width will be chopped rather than being folded.
+- -w, --hilite-unread - Temporarily highlights the first "new" line after a forward movement of a full page.
```
vicentepinto98, line 140: It's missing the newly introduced argument "-R" for the colors

README.md
```diff
@@ -129,6 +129,26 @@ stackit config list
 
 You can also edit the configuration file manually.
 
+## Customization
```
vicentepinto98, line 132: Re-structure this section to be more clear like this:
```
## Customization

### Pager

To specify a custom pager, use the `PAGER` environment variable.

If the variable is not set, STACKIT CLI uses the `less` as default pager.
When using `less` as a pager .....

[IMPORTANT] 
....
```

vicentepinto98:
Please link the GH issue that this PR solves

internal/pkg/print/print.go
```diff
@@ -165,9 +165,15 @@ func (p *Printer) PagerDisplay(content string) error {
 	}
 	pagerCmd := exec.Command("less", "-F", "-S", "-w")
 
+	pager, pagerExists := os.LookupEnv("PAGER")
```
vicentepinto98, line 168: It this a commonly known env var?

README.md
```diff
@@ -129,6 +129,26 @@ stackit config list
...
+STACKIT CLI uses the `less` as default pager. 
+When using `less` as a pager, STACKIT CLI will automatically pass following options
+- -F, --quit-if-one-screen - Less will automatically exit if the entire file can be displayed on the first screen.
+- -S, --chop-long-lines - Lines longer than the screen width will be chopped rather than being folded.
+- -w, --hilite-unread - Temporarily highlights the first "new" line after a forward movement of a full page.
+
+
+>[IMPORTANT]
+> These options will not be added if:
+> - The pager is not named `less`.
```
joaopalet, line 145: Can't the user pass in flags for their custom pager by setting `PAGER` to something like `less -x`?

joaopalet, line 145: Can't the user pass in flags for their custom pager by setting `PAGER` to something like `less -x`?

joaopalet:
```suggestion
> These options will not be added automatically if a custom pager is defined.
```

GokceGK: In the current state of implementation, no args can be passed. Should this be possible? 

If we want to support args, we need to split the env variable content as command and args.

internal/pkg/print/print.go
GokceGK: yes it is, a lot of tools uses PAGER env variable.

README.md
GokceGK: Added a line to explain the -R option.

README.md
GokceGK: Updated the section structure as suggestion

README.md
joaopalet: Even if the env variable is defined as `PAGER="less -x -y -z"`?

GokceGK: yes, even if we pass args like that

vicentepinto98: That's weird :thinking:

vicentepinto98: I would say it's not a must have, but also the other CLIs don't support this?

GokceGK: @vicentepinto98 the one I checked handles the given string differently to detect parameters and pass as args in the exec.command

GokceGK: @joaopalet extended README for the env variables to define parameters as discussed tomorrow :)

README.md
```diff
@@ -130,6 +130,31 @@ stackit config list
...
+
+When using `less` as a pager, STACKIT CLI will automatically pass following options
+- -F, --quit-if-one-screen - Less will automatically exit if the entire file can be displayed on the first screen.
+- -S, --chop-long-lines - Lines longer than the screen width will be chopped rather than being folded.
+- -w, --hilite-unread - Temporarily highlights the first "new" line after a forward movement of a full page.
+- -R, --RAW-CONTROL-CHARS - ANSI color and style sequences will be interpreted.
+
+> These options will not be added automatically if a custom pager is defined.
+> 
+> In that case, users can define the parameters by using the specific environment variable required by the PAGER (if supported by pager).
```
vicentepinto98, line 149:
```suggestion
> In that case, users can define the parameters by using the specific environment variable required by the `PAGER` (if supported).

```

README.md
```diff
@@ -130,6 +130,31 @@ stackit config list
...
+> 
+> For example, if user sets the PAGER environment variable to less and would like to pass some arguments, LESS environment variable must be used as following.
```
vicentepinto98, line 151:
```suggestion
> For example, if user sets the `PAGER` environment variable to `less` and would like to pass some arguments, `LESS` environment variable must be used as following:

```

vicentepinto98 (approved):

PR #303, opened 2024-05-08
Improve examples of LB observability credentials commands

joaopalet:


internal/cmd/load-balancer/observability-credentials/update/update.go
```diff
@@ -46,11 +46,8 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				`Update the password of observability credentials of Load Balancer with credentials reference "credentials-xxx". The password is entered using the terminal`,
 				"$ stackit load-balancer observability-credentials update credentials-xxx"),
 			examples.NewExample(
-				`Update the password of observability credentials of Load Balancer with credentials reference "credentials-xxx", by providing it in the flag`,
-				"$ stackit load-balancer observability-credentials update credentials-xxx --password new-pwd"),
-			examples.NewExample(
-				`Update the display name of observability credentials of Load Balancer with credentials reference "credentials-xxx".`,
-				"$ stackit load-balancer observability-credentials update credentials-xxx --display-name yyy"),
```
joaopalet, line 53: I think this third example was not needed as it's not possible to not update the password

DiogoFerrao (approved):

PR #305, opened 2024-05-08
Integrate WinGet distribution in release pipeline

joaopalet:


.goreleaser.yaml
```diff
@@ -101,7 +101,7 @@ brews:
       email: noreply@stackit.de
     homepage: "https://github.com/stackitcloud/stackit-cli"
     description: "A command-line interface to manage STACKIT resources.\nThis CLI is in a BETA state. More services and functionality will be supported soon."
-    folder: Formula
+    directory: Formula
```
joaopalet, line 104: The previous field name was deprecated: https://goreleaser.com/deprecations#brewsfolder

.goreleaser.yaml
```diff
@@ -125,3 +125,26 @@ snapcrafts:
...
+    # If set to auto, the release will not be uploaded to the homebrew tap repo
+    # if the tag has a prerelease indicator (e.g. v0.0.1-alpha1)
+    # Temporarily not skipping prereleases to test integration with Winget
+    # skip_upload: auto
+    repository:
+      owner: stackitcloud
+      name: winget-pkgs
+      pull_request:
+        enabled: true
+        draft: true
```
joaopalet, line 146: This means that the PR that will created to the Microsoft's repo will be a Draft PR. We can then manually set it to Ready for review if everything looks ok

After all the integration is done and tested we can then disable this

.goreleaser.yaml
```diff
@@ -125,3 +125,26 @@ snapcrafts:
     grade: devel
     # Whether to publish the Snap to the store
     publish: true
+
+winget:
+  - name: stackit
+    publisher: stackitcloud
+    short_description: A command-line interface to manage STACKIT resources.
+    license: Apache-2.0
+    publisher_support_url: "https://github.com/derailed/k9s/issues"
```
DiogoFerrao, line 134: Should point to ours

DiogoFerrao (approved):

PR #307, opened 2024-05-09
MongoDB backup list, describe, restore-jobs

vicentepinto98:


internal/cmd/mongodbflex/backup/describe/describe.go
```diff
@@ -0,0 +1,141 @@
...
+	default:
+		table := tables.NewTable()
+		table.AddRow("ID", *backup.Id)
+		table.AddSeparator()
+		table.AddRow("CREATED AT", *backup.StartTime)
+		table.AddSeparator()
+		table.AddRow("EXPIRES AT", *backup.EndTime)
+		table.AddSeparator()
+		table.AddRow("BACKUP SIZE", bytesize.New(float64(*backup.Size)))
+
```
DiogoFerrao, line 133: As discussed offline, this could include the restore status

vicentepinto98: done

internal/cmd/mongodbflex/backup/list/list.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		table := tables.NewTable()
+		table.SetHeader("ID", "CREATED AT", "EXPIRES AT", "BACKUP SIZE", "RESTORE STATUS")
+		for i := range backups {
+			backup := backups[i]
+			restoreStatus, err := mongodbflexUtils.GetRestoreStatus(ctx, apiClient, model.ProjectId, *model.InstanceId, *backup.Id)
```
vicentepinto98, line 156: It could lead to high response times on the command if there is a high volume of backups and restores, but if we want to show the restore status for each backup it seems like the only way.

I tried another alternative by making `GetRestoreStatus` receive a list of backups instead of a single one and returning a map, but it turned out much more complex and with no significant improve on the performance (we still had to iterate over two arrays.

The current implementation also forces changing the signature of the usual `outputResult`, unfortunately

internal/pkg/services/mongodbflex/utils/utils.go
```diff
@@ -157,3 +158,23 @@ func GetUserName(ctx context.Context, apiClient MongoDBFlexClient, projectId, in
...
+	if restoreJobs.Items == nil {
+		return state, nil
+	}
+
+	for _, restoreJob := range *restoreJobs.Items {
+		if *restoreJob.BackupID == backupId {
+			state = *restoreJob.Status
+			break
+		}
+	}
```
DiogoFerrao, line 178: What happens if the first restore job in the list (with backupID matching) is not the one being ran now?

DiogoFerrao: Not sure if this is something that can happen

internal/cmd/mongodbflex/backup/list/list.go
DiogoFerrao: Wouldn't it be possible to get all the restore jobs for the instance once and then have the utils function receive the list of jobs?
Because the largest overhead in our commands is usually the API call

internal/pkg/services/mongodbflex/utils/utils.go
vicentepinto98: Not sure if I understood what you meant by "the one being ran now". The method returns the restore status of a specific `backupId`, which is passed as an argument

internal/cmd/mongodbflex/backup/list/list.go
vicentepinto98: Thanks for the suggestion, I simplified the code a lot and saved quite some requests

internal/pkg/services/mongodbflex/utils/utils.go
DiogoFerrao: If there are multiple restore jobs, does the API guarantee that the first one in the list is the latest that was ran? That is my question. 
Because if the order in the list is arbitrary, it could happen that this `for loop` finds the first restore job for a backupID that has state `Restored` but in fact the latest restore job that was ran by the user could be further down the list and still have a `In Progress` state.

vicentepinto98: I got it, this is how the Portal is doing it, but nonetheless I will double check with the MongoDBFlex team

DiogoFerrao (approved):

DiogoFerrao (approved):

PR #308, opened 2024-05-09
Onboard extra functionality for load balancer credentials

DiogoFerrao:


internal/cmd/load-balancer/observability-credentials/list/list.go
```diff
@@ -39,6 +44,9 @@ func NewCmd(p *print.Printer) *cobra.Command {
 			examples.NewExample(
 				`List all observability credentials for Load Balancer`,
 				"$ stackit load-balancer observability-credentials list"),
+			examples.NewExample(
+				`List all used observability credentials for Load Balancer`,
```
joaopalet, line 48: I think we could make the example more clear by writting something like `List all observability credentials being used by a Load Balancer`

joaopalet: And maybe rewrite other examples as `List all Load Balancer observability credentials`

internal/cmd/load-balancer/observability-credentials/list/list.go
```diff
@@ -92,6 +105,10 @@ func NewCmd(p *print.Printer) *cobra.Command {
 
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().Int64(limitFlag, 0, "Maximum number of entries to list")
+	cmd.Flags().Bool(usedFlag, false, "List only used credentials")
```
joaopalet, line 108: Maybe also rephrase to something like `List only credentials being used by a Load Balancer`

internal/pkg/services/load-balancer/utils/utils.go
```diff
@@ -130,3 +133,100 @@ func GetTargetName(ctx context.Context, apiClient LoadBalancerClient, projectId,
...
+	if err != nil {
+		return nil, fmt.Errorf("get used observability credentials: %w", err)
+	}
+
+	if unused {
+		filteredCredentials = GetUnusedObsCredentials(filteredCredentials, credentials)
+	}
+
+	return filteredCredentials, nil
+}
```
joaopalet, line 232: Not sure if it's clearer, what do you think?

```suggestion
func FilterCredentials(ctx context.Context, client LoadBalancerClient, allCredentials []loadbalancer.CredentialsResponse, projectId string, used, unused bool) ([]loadbalancer.CredentialsResponse, error) {
	if !used && !unused {
		return allCredentials, nil
	}
	if used && unused {
		return nil, fmt.Errorf("used and unused flags are mutually exclusive")
	}
	usedCredentials, err := GetUsedObsCredentials(ctx, client, allCredentials, projectId)
	if err != nil {
		return nil, fmt.Errorf("get used observability credentials: %w", err)
	}

    if unused {
		return GetUnusedObsCredentials(filteredCredentials, allCredentials), nil
    }
    return usedCredentials, nil
}
```

internal/pkg/services/load-balancer/utils/utils_test.go
joaopalet, line 1: Nice tests :+1:

internal/pkg/services/load-balancer/utils/utils.go
joaopalet: I think the formatting is a bit off in my suggestion

DiogoFerrao: I agree, it's better

internal/pkg/services/load-balancer/utils/utils_test.go
```diff
@@ -793,3 +841,314 @@ func TestGetTargetName(t *testing.T) {
...
+			description:         "no credentials",
+			allCredentials:      []loadbalancer.CredentialsResponse{},
+			expectedCredentials: []loadbalancer.CredentialsResponse{},
+			isValid:             true,
+		},
+		{
+			description:            "list load balancers fails",
+			listLoadBalancersFails: true,
+			isValid:                false,
+		},
```
joaopalet, line 1125: Don't you need to pass a `filterOp` to these tests? Otherwise the error will be because of the invalid filter, and not the actual error of the use case we are aiming to test

internal/cmd/load-balancer/observability-credentials/list/list.go
```diff
@@ -22,23 +23,33 @@ import (
...
 		Use:   "list",
-		Short: "Lists all observability credentials for Load Balancer",
-		Long:  "Lists all observability credentials for Load Balancer.",
+		Short: "Lists observability credentials for Load Balancer",
+		Long:  "Lists observability credentials for Load Balancer.",
 		Args:  args.NoArgs,
 		Example: examples.Build(
 			examples.NewExample(
-				`List all observability credentials for Load Balancer`,
+				`List all Load Balancer observability credentials`,
```
joaopalet, line 45: Should also be done for the last 2 examples

joaopalet (approved):

PR #309, opened 2024-05-09
Implement `GetProfile`

joaopalet:


internal/pkg/config/file_utils.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	return nil
+}
+
+// readFileIfExists reads the contents of a file and returns it as a string, along with a boolean indicating if the file exists.
+// If the file does not exist, it returns an empty string and no error.
+// If the file exists but cannot be read, it returns an error.
+func readFileIfExists(filePath string) (contents string, exists bool, err error) {
+	_, err = os.Stat(filePath)
+	if os.IsNotExist(err) {
```
vicentepinto98, line 27: should this function continue if there is an error that is not of type `IsNotExist`? 
I guess the call to `ReadFile` will always fail if there was an error on `Stat`

internal/pkg/config/profiles.go
```diff
@@ -0,0 +1,51 @@
...
+
+	err := validateProfile(profile)
+	if err != nil {
+		return "", fmt.Errorf("validate profile from env var: %w", err)
+	}
+	return profile, nil
+}
+
+// validateProfile validates the profile name.
+// It can only use letters, numbers, or "-".
```
vicentepinto98, line 40:
```suggestion
// It can only use letters, numbers, or "-" and cannot be empty.
```

internal/pkg/config/file_utils.go
joaopalet: Good catch

vicentepinto98 (approved):

PR #311, opened 2024-05-10
Onboard `load-balancer observability-credentials cleanup`

DiogoFerrao:


internal/cmd/load-balancer/observability-credentials/cleanup/cleanup.go
```diff
@@ -0,0 +1,140 @@
...
+			if len(credentials) == 0 {
+				p.Info("No unused observability credentials found on project %q\n", projectLabel)
+				return nil
+			}
+
+			if !model.AssumeYes {
+				prompt := "Will delete the following unused observability credentials: \n"
+				for _, credential := range credentials {
+					if credential.DisplayName == nil || credential.Username == nil {
+						return fmt.Errorf("delete Load Balancer observability credentials: missing display name or username")
```
joaopalet, line 77:
```suggestion
						return fmt.Errorf("list unused Load Balancer observability credentials: credentials %q missing display name or username", *credential.CredentialsRef)
```

internal/cmd/load-balancer/observability-credentials/cleanup/cleanup.go
```diff
@@ -0,0 +1,140 @@
...
+				err = p.PromptForConfirmation(prompt)
+				if err != nil {
+					return err
+				}
+			}
+
+			for _, credential := range credentials {
+				if credential.CredentialsRef == nil {
+					return fmt.Errorf("delete Load Balancer observability credentials: missing credentials reference")
+				}
```
joaopalet, line 93: I think this will never happen

internal/cmd/load-balancer/observability-credentials/cleanup/cleanup.go
```diff
@@ -0,0 +1,140 @@
...
+				credentialsRef := *credential.CredentialsRef
+				// Call API
+				req := buildDeleteCredentialRequest(ctx, model, apiClient, credentialsRef)
+				_, err = req.Execute()
+				if err != nil {
+					return fmt.Errorf("delete Load Balancer observability credentials: %w", err)
+				}
+			}
+
+			p.Info("Deleted unused observability credentials on project %q\n", projectLabel)
```
joaopalet, line 103:
```suggestion
			p.Info("Deleted unused Load Balancer observability credentials on project %q\n", projectLabel)
```

internal/cmd/load-balancer/observability-credentials/cleanup/cleanup.go
DiogoFerrao: Should I not check for it then? I think it makes sense to avoid a segmentation fault being displayed to the user

joaopalet: Yeah it's fine, there are a lot of places where this could happen theoretically but the object is required in the API response

joaopalet (approved):

PR #312, opened 2024-05-10
Implement `stackit config profile` commands

joaopalet:


internal/pkg/errors/errors.go
```diff
@@ -113,6 +113,10 @@ For more details on the available storages for the configured flavor (%[3]s), ru
 
 	SUBCOMMAND_MISSING = `missing subcommand`
 
+	INVALID_PROFILE = `the profile name %q is invalid.
+
```
DiogoFerrao, line 117:
```suggestion
	INVALID_PROFILE_NAME = `the profile name %q is invalid.
	
```

DiogoFerrao:
I have some overall comments that maybe are not completely related to this PR:

- Would be cool to have more information on the `config list` so that the user can see what each configuration is. This could include the project name and the email of the user. GCP does this:
image
- We should probably add to the debug logs in the PersistentRunE in the root command stating which config profile is being used.
- The docs in the root `config.go` command could also be updated with information as to where the new multi configs are stored.

joaopalet:
> Would be cool to have more information on the config list so that the user can see what each configuration is. This could include the project name and the email of the user. Also, being able to tell which config is active is really useful,
I don't think we can currently check this, right?

Regarding this point:
- I've the active profile to the output of the `config list` command
- For other details of the profiles, we could have a new `config profile list` command that would show something like:
```
PROFILE NAME  │  IS ACTIVE  │      ACCOUNT       │ ...
──────────────┼─────────────┼────────────────────┼ ...
 profile-1    │ true        │ aaa@email.com      │ ...
 profile-2    │ false       │ bbb@email.com      │ ...
```

DiogoFerrao (approved):

PR #314, opened 2024-05-10
Configure table titles

vicentepinto98:
Example output

image


internal/pkg/print/print.go
```diff
@@ -171,7 +171,12 @@ func (p *Printer) PagerDisplay(content string) error {
 	if outputFormat == NoneOutputFormat {
 		return nil
 	}
-	pagerCmd := exec.Command("less", "-F", "-S", "-w")
+
+	// less arguments
+	// -F: exits if the entire file fits on the first screen
+	// -S: disables line wrapping
+	// -R: interprets ANSI color and style sequences
+	pagerCmd := exec.Command("less", "-F", "-S", "-R")
```
vicentepinto98, line 179: the `-w` argument was removed because it is used to wrap lines longer than the width of the terminal window, but we are disabling line wrapping with `-S` and that's the intended behavior

joaopalet: `-w` was used to highlight the first line after moving one full page down (with the `f` key)
image

internal/pkg/print/print.go
```diff
@@ -171,7 +171,12 @@ func (p *Printer) PagerDisplay(content string) error {
 	if outputFormat == NoneOutputFormat {
 		return nil
 	}
-	pagerCmd := exec.Command("less", "-F", "-S", "-w")
+
+	// less arguments
```
joaopalet, line 175: 👍

joaopalet (commented):
In the `mongodbflex` and `postgresflex` `options` commands, there is also the possibility that multiple tables are displayed if the user passes in multiple flags (e.g. `--versions --flavors`). We could also show table titles in that case I think

internal/pkg/print/print.go
vicentepinto98: I guess I read some wrong documentation then, re-added it

joaopalet (approved):

PR #315, opened 2024-05-13
Improvements to profiles functionality

joaopalet:
- Add table titles to `config set` command
- Add debug logs to `config profile set/unset`
- Fix setting a new profile when there is a currently active profile

Only two last commits are relevant, the rest is just bringing the latest changes from main

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -53,7 +53,7 @@ func NewRootCmd(version, date string, p *print.Printer) *cobra.Command {
 			p.Debug(print.DebugLevel, "arguments: %s", argsString)
 
 			configFilePath := viper.ConfigFileUsed()
-			p.Debug(print.DebugLevel, "will persist and read config from: %s", configFilePath)
+			p.Debug(print.DebugLevel, "configuration is persisted and read from: %s", configFilePath)
 
 			activeProfile, err := config.GetProfile()
 			if err != nil {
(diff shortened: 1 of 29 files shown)


internal/cmd/root.go
```diff
@@ -53,7 +53,7 @@ func NewRootCmd(version, date string, p *print.Printer) *cobra.Command {
 			p.Debug(print.DebugLevel, "arguments: %s", argsString)
 
 			configFilePath := viper.ConfigFileUsed()
-			p.Debug(print.DebugLevel, "using config file: %s", configFilePath)
+			p.Debug(print.DebugLevel, "will persist and read config from: %s", configFilePath)
```
vicentepinto98, line 56: `configuration is persisted and read from: `

This way it works for all cases

internal/pkg/config/profiles.go
```diff
@@ -38,27 +39,36 @@ func GetProfile() (string, error) {
...
 	}
+	p.Debug(print.DebugLevel, "persisted new active profile in: %s", profileFilePath)
+
+	configFolderPath = filepath.Join(defaultConfigFolderPath, profile)
+	err = createFolderIfNotExists(configFolderPath)
+	if err != nil {
+		return fmt.Errorf("create config folder: %w", err)
+	}
+	p.Debug(print.DebugLevel, "created folder for the new profile: %s", configFolderPath)
+
```
vicentepinto98, line 60: maybe another debug log here explicitly saying that the new profile is now active

vicentepinto98 (approved):

PR #326, opened 2024-05-17
Implement profiles for auth settings

GokceGK:


internal/pkg/auth/storage.go
```diff
@@ -69,6 +71,15 @@ func SetAuthField(key authFieldKey, value string) error {
 }
 
 func setAuthFieldInKeyring(key authFieldKey, value string) error {
+	activeProfile, err := config.GetProfile()
+	if err != nil {
+		return fmt.Errorf("get profile: %w", err)
+	}
```
joaopalet, line 77: Would it make sense to move these to `SetAuthFeld` and pass the profile as an argument? Same for the getters

joaopalet: We can discuss this once you're back as it would have no influence in the behaviour

internal/pkg/auth/storage.go
```diff
@@ -69,6 +71,15 @@ func SetAuthField(key authFieldKey, value string) error {
...
+
+	if activeProfile != "" {
+		activeProfileKeyring := fmt.Sprintf("%s%s%s", keyringService, "/", activeProfile)
```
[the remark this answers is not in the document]
joaopalet: Using `filepath.Join`

DiogoFerrao (approved):

PR #327, opened 2024-05-20
Add warnings when set/unset profile with no auth

DiogoFerrao:


internal/cmd/config/profile/set/set.go
```diff
@@ -49,6 +50,14 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				return fmt.Errorf("set profile: %w", err)
 			}
 
+			flow, err := auth.GetAuthFlow()
+			if err != nil {
+				p.Debug(print.WarningLevel, "both keyring and text file storage failed to find a valid authentication flow for the active profile")
+				p.Warn("Failed to find a valid authentication flow for the active profile. Please login using the 'stackit auth login' command.\n")
+			} else {
+				p.Debug(print.InfoLevel, "found valid authentication flow for active profile: %s", auth.GetPrettyAuthFlow(flow))
```
vicentepinto98, line 58: do we have other INFO logs or is this the first one? We could consider no having INFO at all, just DEBUG, WARN or ERROR for simplicity

internal/cmd/config/profile/set/set.go
```diff
@@ -49,6 +50,14 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				return fmt.Errorf("set profile: %w", err)
 			}
 
+			flow, err := auth.GetAuthFlow()
+			if err != nil {
+				p.Debug(print.WarningLevel, "both keyring and text file storage failed to find a valid authentication flow for the active profile")
+				p.Warn("Failed to find a valid authentication flow for the active profile. Please login using the 'stackit auth login' command.\n")
```
vicentepinto98, line 56: Make this message more user friendly, "authentication flow" is not something that is user-facing.
It can be just "The active profile "profile-name" is not authenticated, please login ...."

internal/cmd/config/profile/unset/unset.go
```diff
@@ -31,6 +32,14 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				return fmt.Errorf("unset profile: %w", err)
 			}
 
+			flow, err := auth.GetAuthFlow()
+			if err != nil {
+				p.Debug(print.WarningLevel, "both keyring and text file storage failed to find a valid authentication flow for the active profile")
+				p.Warn("Failed to find a valid authentication flow for the active profile. Please login using the 'stackit auth login' command.\n")
```
vicentepinto98, line 38: The unset command always sets the profile to the default one right? We could also make it more clear in this message that the **default** profile is not authenticated

internal/pkg/auth/storage.go
```diff
@@ -260,3 +260,16 @@ func createEncodedTextFile() error {
 
 	return nil
 }
+
+func GetPrettyAuthFlow(flow AuthFlow) string {
```
vicentepinto98, line 264: do we really need this if it's just used in debug logs? :thinking:

DiogoFerrao: Yes, maybe we don't. I'll just print the raw string, its very readable

internal/cmd/config/profile/set/set.go
DiogoFerrao: That's true, we don't have it anywhere else. I'll simplify it then. But maybe we should at some point consider having this diversity to make the debug logs more readable

PR #329, opened 2024-05-21
Rework active profile handling in auth

GokceGK:


internal/pkg/auth/storage_test.go
```diff
@@ -308,8 +318,13 @@ func TestSetGetAuthFieldEncodedTextFile(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
+			activeProfile, err := config.GetProfile()
+			if err != nil {
+				t.Errorf("get profile: %v", err)
+			}
```
joaopalet, line 324: For this one we could make the `activeProfile` a field if the test, which would allow us to test these functions for empty profile and active profile cases

internal/pkg/auth/storage_test.go
```diff
@@ -219,8 +224,13 @@ func TestSetGetAuthFieldKeyring(t *testing.T) {
 		t.Run(tt.description, func(t *testing.T) {
 			keyring.MockInit()
 
+			activeProfile, err := config.GetProfile()
+			if err != nil {
+				t.Errorf("get profile: %v", err)
+			}
```
joaopalet, line 230: Same here

internal/pkg/auth/storage_test.go
```diff
@@ -219,8 +363,21 @@ func TestSetGetAuthFieldKeyring(t *testing.T) {
...
+				err := config.SetProfile(p, tt.activeProfile)
+				if err != nil {
+					t.Errorf("set profile: %v", err)
+				}
+			} else {
+				err := config.UnsetProfile(p)
+				if err != nil {
+					t.Errorf("unset profile: %v", err)
+				}
+			}
```
joaopalet, line 377: Do we really need to set and unset profiles in tests? I think it's not needed for this function and for `TestSetGetAuthFieldEncodedTextFile`

joaopalet: But in any case, these are actually creacting files in the user config dir and actually setting and unsetting profiles for the user, without cleaning up after themselves

GokceGK: removed the set/unset from all test cases

joaopalet (approved):

PR #332, opened 2024-05-22
Increase test coverage

DiogoFerrao:


internal/cmd/auth/activate-service-account/activate_service_account_test.go
```diff
@@ -120,3 +122,62 @@ func TestParseInput(t *testing.T) {
...
+				if err == nil {
+					t.Fatalf("did not fail on invalid input")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("store flags: %v", err)
+			}
+
+			value, err := auth.GetAuthField(auth.TOKEN_CUSTOM_ENDPOINT)
```
vicentepinto98, line 166: why only check these two fields and not all of the ones present in the `inputModel`?

internal/cmd/root_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	setupCmd()
+	for _, tt := range tests {
+		t.Run(tt.description, func(t *testing.T) {
+			actualError := beautifyUnknownAndMissingCommandsError(cmd, tt.inputError)
+
+			if tt.isNotUnknownFlagError {
+				if actualError.Error() != tt.expectedMsg {
+					t.Fatalf("expected error to be %s, got %s", tt.expectedMsg, actualError.Error())
```
vicentepinto98, line 64:
```suggestion
					t.Fatalf("expected error message to be %s, got %s", tt.expectedMsg, actualError.Error())
```

internal/cmd/root_test.go
```diff
@@ -0,0 +1,75 @@
...
+				}
+				return
+			}
+
+			appendedErr := pkgErrors.AppendUsageTip(errors.New(tt.expectedMsg), cmd)
+			if actualError.Error() != appendedErr.Error() {
+				t.Fatalf("expected error to be %s, got %s", appendedErr.Error(), actualError.Error())
```
vicentepinto98, line 71:
```suggestion
				t.Fatalf("expected error message to be %s, got %s", appendedErr.Error(), actualError.Error())
```

internal/pkg/config/config_test.go
```diff
@@ -0,0 +1,63 @@
...
+				t.Fatalf("expected error to be nil, got %v", err)
+			}
+
+			// Check if the file was created
+			_, err = os.Stat(configPath)
+			if os.IsNotExist(err) {
+				t.Fatalf("expected file to exist, got %v", err)
+			}
+
+			// Delete the file
```
vicentepinto98, line 56: Shouldn't we also remove the `tempDir` that was created?

internal/cmd/auth/activate-service-account/activate_service_account_test.go
DiogoFerrao: Because those are the only two fields being set in the `storeFlags` method. I could make sure that the other known AuthFields are empty but since the keys are just strings, it doesn't fully test unwanted behavior, so I just tested the positive case.

PR #341, opened 2024-05-23
Implement file path flag for generate-payload

GokceGK:


internal/pkg/print/print.go
```diff
@@ -197,6 +197,27 @@ func (p *Printer) PagerDisplay(content string) error {
...
+		}
+	}()
+
+	_, err = fo.WriteString(content)
+	if err != nil {
+		return fmt.Errorf("write content to output file: %w", err)
+	}
+
+	return nil
+}
```
DiogoFerrao, line 219: I don't think this should be part of the `print` package since it doesn't use printer functionality. We should have a `fileutils` package to do things like this. In the `feature/multi-config` branch, we have something like this implemented, although inside the `config` package for some reason:
https://github.com/stackitcloud/stackit-cli/blob/807d6887c422d96ece8d412b5f7491aac96b7f60/internal/pkg/config/file_utils.go#L11

I think you should extract this function to `pkg/file_utils/file_utils.go` and then I'll move the code in the `multi-config` branch to there as well

internal/cmd/argus/scrape-config/generate-payload/generate_payload.go
```diff
@@ -92,6 +94,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().Var(flags.UUIDFlag(), instanceIdFlag, "Instance ID")
 	cmd.Flags().StringP(jobNameFlag, "n", "", "If set, generates an update payload with the current state of the given scrape config. If unset, generates a create payload with default values")
+	cmd.Flags().StringP(filePathFlag, "f", "", "If set, writes the payload in the given file. If unset, writes the payload to the standard output")
 }
```
DiogoFerrao, line 98:
```suggestion
	cmd.Flags().StringP(filePathFlag, "f", "", "If set, writes the payload to the given file. If unset, writes the payload to the standard output")
}
```

internal/cmd/argus/scrape-config/generate-payload/generate_payload.go
```diff
@@ -116,22 +120,38 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *argus.APICl
 	return req
 }
 
-func outputCreateResult(p *print.Printer, payload *argus.CreateScrapeConfigPayload) error {
+func outputCreateResult(p *print.Printer, model *inputModel, payload *argus.CreateScrapeConfigPayload) error {
```
DiogoFerrao, line 123: I think you can just pass `filePath` here, we don't need the entire model

internal/cmd/argus/scrape-config/generate-payload/generate_payload.go
```diff
@@ -44,12 +46,12 @@ func NewCmd(p *print.Printer) *cobra.Command {
 		Example: examples.Build(
 			examples.NewExample(
 				`Generate a Create payload with default values, and adapt it with custom values for the different configuration options`,
-				`$ stackit argus scrape-config generate-payload > ./payload.json`,
+				`$ stackit argus scrape-config generate-payload --file-path ./payload.json`,
 				`<Modify payload in file, if needed>`,
 				`$ stackit argus scrape-config create my-config --payload @./payload.json`),
```
DiogoFerrao, line 51: Since we are keeping the option to output to StdOut I think we should keep an example without the `--file-path` flag and say something like:

*Generate an Update payload with the values of an existing configuration (...) and preview it in the terminal*

DiogoFerrao:
The comments about `generate_payload.go` apply to the three commands

internal/pkg/print/print.go
joaopalet: This function as is will never return that error as it just assigns it to a variable that is not used afterwards.

In this case if we also want to capture the error of closing the file we would need to use a named parameter ad do something like:

```suggestion
func (p *Printer) FileOutput(outputFileName, content string) (err error) { // Notice the named variable here
	fo, err := os.Create(outputFileName)
	if err != nil {
		return fmt.Errorf("create output file: %w", err)
	}

	defer func() {
		tempErr := fo.Close()
		if tempErr != nil {
            if err != nil {
				err = fmt.Errorf("%v; close output file: %w", err, tmpErr)
			} else {
				err = fmt.Errorf("close output file: %w", tmpErr)
			}
		}
	}()

	_, err = fo.WriteString(content)
	if err != nil {
		return fmt.Errorf("write content to output file: %w", err)
	}

	return err
}
```

joaopalet: Regarding your point @DiogoFerrao: also tend to agree it would fir better in another package, as I see it no as an output of the CLI but as part of the functionality of the command to write something to a file. For this reason also would rename it to something like `WriteToFile`

DiogoFerrao (approved):

PR #344, opened 2024-05-24
Add `Traceparent` header to the raw HTTP debug logs

DiogoFerrao:


internal/pkg/print/debug.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/stackitcloud/stackit-sdk-go/core/config"
 )
 
-var defaultHTTPHeaders = []string{"Accept", "Content-Type", "Content-Length", "User-Agent", "Date", "Referrer-Policy"}
+var defaultHTTPHeaders = []string{"Accept", "Content-Type", "Content-Length", "User-Agent", "Date", "Referrer-Policy", "Traceparent"}
```
vicentepinto98, line 16: this is not case sensitive right? The header is just `traceparent`

DiogoFerrao: It is but I tested and the header is `Traceparent` with a capital T.
image

DiogoFerrao: Without capitalization nothing appears

vicentepinto98: I was checking in Postman, so I guess Postman is making it everything lower-case

vicentepinto98: it's fine like this then for the scope of this PR

PR #345, opened 2024-05-24
Add `sqlserverflex` client and configuration logic (and `instance list` command)

joaopalet:


internal/cmd/sqlserverflex/instance/list/list.go
```diff
@@ -0,0 +1,164 @@
...
+		details, err := yaml.MarshalWithOptions(instances, yaml.IndentSequence(true))
+		if err != nil {
+			return fmt.Errorf("marshal SQLServer Flex instance list: %w", err)
+		}
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		table := tables.NewTable()
+		table.SetHeader("ID", "NAME", "STATUS")
```
vicentepinto98, line 152:
```suggestion
		table.SetHeader("ID", "NAME", "STATE")
```

for uniformity with the other services

joaopalet: For the other Flex databases we also have "status", as that's the field name in the JSON response. Same for Argus instance for example

vicentepinto98: Alright, I checked and on the Portal it's also "Status"

vicentepinto98 (approved):

PR #350, opened 2024-05-27
Add `beta` to `sqlserverflex` command examples

joaopalet:


internal/pkg/errors/errors.go
```diff
@@ -194,21 +194,23 @@ func (e *DSAInvalidPlanError) Error() string {
...
 func (e *DatabaseInputFlavorError) Error() string {
 	fullCommandPath := e.Cmd.CommandPath()
 	if len(e.Args) > 0 {
 		fullCommandPath = fmt.Sprintf("%s %s", fullCommandPath, strings.Join(e.Args, " "))
 	}
-	// Assumes a structure of the form "stackit <service> <resource> <operation>"
-	service := e.Cmd.Parent().Parent().Use
 
-	return fmt.Sprintf(DATABASE_INVALID_INPUT_FLAVOR, fullCommandPath, service)
+	if e.Service == "" {
```
vicentepinto98, line 208: :+1:

internal/pkg/errors/errors.go
```diff
@@ -194,21 +194,23 @@ func (e *DSAInvalidPlanError) Error() string {
 }
 
 type DatabaseInputFlavorError struct {
-	Service   string
-	Operation string
-	Cmd       *cobra.Command
-	Args      []string
+	Service string
+	Cmd     *cobra.Command
```
vicentepinto98, line 198: The `Operation` field was not used for anything?

joaopalet: No, I think we forgot to cleanup there. The `Service` field was also not used (expect in the test)

vicentepinto98 (approved):

PR #351, opened 2024-05-27
Add `stackit profile create` command and refactor `set` command

DiogoFerrao:


internal/pkg/fileutils/file_utils.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 )
 
+// WriteToFile writes the given content to a file.
```
joaopalet, line 8: 👍

internal/cmd/config/profile/set/set.go
```diff
@@ -33,7 +34,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 			"A new profile is created automatically if it does not exist.",
```
joaopalet, line 34: This can now be removed

internal/cmd/config/profile/set/set.go
```diff
@@ -67,9 +68,21 @@ func NewCmd(p *print.Printer) *cobra.Command {
...
+	// Read profile name
+	if len(inputArgs) > 0 {
+		profile = inputArgs[0]
+	} else {
+		var profileSet bool
+		profile, profileSet = config.GetProfileFromEnv()
+		if !profileSet {
+			return nil, &errors.ProfileNameNotProvided{}
+		}
+	}
```
joaopalet, line 83: What is the use case of setting a profile using an env var?

Using the env var or any command (e.g. `STACKIT_CLI_PROFILE=alt-profile stackit dns zone list`) will execute that command using the `alt-profile`, overriding any profile that might be set in the file, right?

internal/cmd/config/profile/create/create.go
```diff
@@ -0,0 +1,129 @@
...
+			"Creates a CLI configuration profile based on the currently active profile and sets it as active.",
+			`The profile name can be provided via the STACKIT_CLI_PROFILE environment variable or as an argument in this command.`,
+			"The environment variable takes precedence over the argument.",
+			"If you do not want to set the profile as active, use the --no-set flag.",
+			"If you want to create the new profile with the initial default configurations, use the --empty flag.",
+		),
+		Args: args.SingleOptionalArg(profileArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a new configuration profile "my-profile" as the active profile`,
```
joaopalet, line 46:
```suggestion
				`Create a new configuration profile "my-profile" with the current configuration, setting it as the active profile`,
```

internal/pkg/config/profiles.go
```diff
@@ -20,7 +22,7 @@ import (
 //
 // If the profile is not valid, it returns an error.
 func GetProfile() (string, error) {
-	profile, profileSet := os.LookupEnv("STACKIT_CLI_PROFILE")
+	profile, profileSet := os.LookupEnv(ProfileEnvVar)
```
joaopalet, line 25:
```suggestion
	profile, profileSet := GetProfileFromEnv(ProfileEnvVar)
```

internal/pkg/config/profiles.go
```diff
@@ -39,6 +41,98 @@ func GetProfile() (string, error) {
...
+// If the profile is not valid, it returns an error.
+func GetProfileFromEnv() (string, bool) {
+	return os.LookupEnv(ProfileEnvVar)
+}
+
+// CreateProfile creates a new profile.
+// It creates a new folder for the profile and copies the config file from the current profile to the new profile.
+// If the fromDefault flag is set, it creates an empty profile.
+// If the setProfile flag is set, it sets the new profile as the active profile.
+// If the profile already exists, it returns an error.
```
joaopalet, line 55:
```suggestion
// CreateProfile creates a new profile.
// If emptyProfile is true, it creates an empty profile. Otherwise, copies the config from the current profile to the new profile.
// If setProfile is true, it sets the new profile as the active profile.
// If the profile already exists, it returns an error.
```

internal/cmd/config/profile/set/set.go
DiogoFerrao: Good point, its not very good functionality. Removed

internal/cmd/config/profile/set/set.go
joaopalet, line 1: Onde more thing I think it's missing: according to the story, the `set` command is supposed to error if the profile does not yet exist, right? And prompt the user to create it using the `create` command

Currently it still directly creates a new profile and activates it. Or were you planning on addressing this in a next PR?

joaopalet: If we really want that, it should also be the behaviour for when a command is run with the `STACKIT_CLI_PROFILE` env var set to a non-existent profile (command would fail and the error would indicate the profile needs to be created)

joaopalet: This would probably mean changing the `GetProfile` to check if the folder exists and erroring otherwise. And calling it before setting the profile in the `set` command

DiogoFerrao: Great point, added that functionality.

internal/pkg/errors/errors.go
```diff
@@ -38,6 +38,10 @@ Please double check if they are correctly configured.
 For more details run:
   $ stackit auth activate-service-account -h`
 
+	SET_INEXISTENT_PROFILE = `the configuration profile %q does not exist.
+Please create it by running:
+  $ stackit config profile create %q`
```
joaopalet, line 43:
```suggestion

To create it, run:
  $ stackit config profile create %q`
```

internal/pkg/errors/errors.go
```diff
@@ -147,6 +151,14 @@ func (e *ActivateServiceAccountError) Error() string {
 	return FAILED_SERVICE_ACCOUNT_ACTIVATION
 }
 
+type SetInexistentProfile struct {
+	Profile string
+}
+
+func (e *SetInexistentProfile) Error() string {
+	return fmt.Sprintf(SET_INEXISTENT_PROFILE, e.Profile, e.Profile)
```
joaopalet, line 159: You don't need to pass the argument here twice, you can pass it once and use `%[1]s` (you can see some examples in the other errors)

joaopalet (approved):

PR #359, opened 2024-05-29
Add `stackit config profile list/delete` commands

DiogoFerrao:


internal/cmd/config/profile/delete/delete.go
```diff
@@ -0,0 +1,91 @@
...
+			"Delete a CLI configuration profile.",
+			"If the deleted profile is the active profile, the default profile will be set to active.",
+		),
+		Args: args.SingleArg(profileArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Delete the configuration profile "my-profile"`,
+				"$ stackit config profile delete my-profile"),
+		),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
joaopalet, line 39: We should also prompt for confirmation here, right?

internal/pkg/auth/storage_test.go
```diff
@@ -452,16 +606,191 @@ func TestSetGetAuthFieldEncodedTextFile(t *testing.T) {
...
+			tt.activeProfile = makeProfileNameUnique(tt.activeProfile)
+
+			// Make sure profile name is valid
+			err := config.ValidateProfile(tt.activeProfile)
+			if err != nil {
+				t.Fatalf("Profile name \"%s\" is invalid: %v", tt.activeProfile, err)
+			}
+
+			oldUserEmail, _ := getAuthFieldWithProfile(tt.activeProfile, USER_EMAIL)
+			oldServiceAccEmail, _ := getAuthFieldWithProfile(tt.activeProfile, SERVICE_ACCOUNT_EMAIL)
```
joaopalet, line 738: Is this logic to reset previous values still needed now that we are using a mocked keyring?

internal/pkg/auth/storage.go
```diff
@@ -214,3 +226,28 @@ func createEncodedTextFile(activeProfile string) error {
 
 	return nil
 }
+
+// GetProfileEmail returns the email of the user associated with the given profile.
+// If the profile is not authenticated, it returns an empty string.
```
joaopalet, line 231:
```suggestion
// GetProfileEmail returns the email of the user or service account associated with the given profile.
// If the profile is not authenticated, it returns an empty string.
// If the email cannot be obtained, an empty string is returned.
```

internal/pkg/auth/storage.go
```diff
@@ -214,3 +226,28 @@ func createEncodedTextFile(activeProfile string) error {
...
+func GetProfileEmail(profile string) string {
+	authFlow, err := getAuthFlowWithProfile(profile)
```
joaopalet, line 233: I think it would be easier to read of this was inlined instead of calling a small new function that is just used here

joaopalet: But it's also fine this way

DiogoFerrao: I agree, the original intention was to abstract the fetching of the authflow, but I don't think it will change and it is also done in this package

internal/pkg/auth/storage_test.go
DiogoFerrao: No, nice catch

joaopalet (approved):

internal/pkg/auth/storage.go
```diff
@@ -44,6 +44,23 @@ const (
...
+		SERVICE_ACCOUNT_TOKEN,
+		SERVICE_ACCOUNT_EMAIL,
+		USER_EMAIL,
+		SERVICE_ACCOUNT_KEY,
+		PRIVATE_KEY,
+		TOKEN_CUSTOM_ENDPOINT,
+		JWKS_CUSTOM_ENDPOINT,
+		authFlowType,
+	}
+}
```
joaopalet, line 62: For consistency with the array of config keys in `config.go`


```suggestion
var authFieldKeys = []authFieldKey{
		SESSION_EXPIRES_AT_UNIX,
		ACCESS_TOKEN,
		REFRESH_TOKEN,
		SERVICE_ACCOUNT_TOKEN,
		SERVICE_ACCOUNT_EMAIL,
		USER_EMAIL,
		SERVICE_ACCOUNT_KEY,
		PRIVATE_KEY,
		TOKEN_CUSTOM_ENDPOINT,
		JWKS_CUSTOM_ENDPOINT,
		authFlowType,
}
```

internal/pkg/auth/storage.go
```diff
@@ -214,3 +247,46 @@ func createEncodedTextFile(activeProfile string) error {
...
+	if err != nil {
+		return fmt.Errorf("validate profile: %w", err)
+	}
+
+	for _, key := range allKeys {
+		err := deleteAuthFieldInKeyring(profile, key)
+		if err != nil {
+			// If the key doesn't exist, continue
+			continue
+		}
```
joaopalet, line 288: Is not existing the only error that can happen here?

joaopalet (dismissed):
Otherwise LGTM

joaopalet (approved):

PR #367, opened 2024-06-03
Choose Snap distribution channel based on tag and remove beta notices

joaopalet:
- Should only be merged after WinGet distribution is finalized


bahkauv70:
closed to inactivity

PR #392, opened 2024-06-27
Add env value check to project name detection

GokceGK:


internal/pkg/projectname/project_name.go
```diff
@@ -77,3 +79,13 @@ func isProjectIdSetInFlags(p *print.Printer, cmd *cobra.Command) bool {
 	}
 	return projectIdSetInFlag
 }
+
+func isProjectIdSetInEnvVar() bool {
+	// Reads the project Id from the environment variable PROJECT_ID
+	projectIdFromEnv := os.Getenv("STACKIT_PROJECT_ID")
```
joaopalet, line 85: There is `value, valueSet := os.LookupEnv("SOME_ENV_VAR")` which already returns a bool that is true if the value is set

joaopalet (approved):

PR #398, opened 2024-07-02
Check for service enablement before bucket creation

joaopalet:
- Also fixes the check in SKE and uniformizes the approaches

### DIFF
--- a/internal/cmd/ske/cluster/create/create.go
+++ b/internal/cmd/ske/cluster/create/create.go
@@ -87,7 +87,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				}
 			}
 
-			// Check if SKE is enabled for this project
+			// Check if the project is enabled before trying to create
 			enabled, err := skeUtils.ProjectEnabled(ctx, apiClient, model.ProjectId)
 			if err != nil {
 				return err
--- a/internal/cmd/ske/cluster/create/create.go
+++ b/internal/cmd/ske/cluster/create/create.go
@@ -93,7 +93,9 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				return err
 			}
 			if !enabled {
-				return fmt.Errorf("SKE isn't enabled for this project, please run 'stackit ske enable'")
+				return &errors.ServiceDisabledError{
+					Service: "ske",
+				}
 			}
 
 			// Check if cluster exists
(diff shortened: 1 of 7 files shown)


internal/cmd/ske/cluster/create/create.go
```diff
@@ -87,13 +87,15 @@ func NewCmd(p *print.Printer) *cobra.Command {
...
-			// Check if SKE is enabled for this project
+			// Check if the project is enabled before trying to create
 			enabled, err := skeUtils.ProjectEnabled(ctx, apiClient, model.ProjectId)
 			if err != nil {
 				return err
 			}
 			if !enabled {
-				return fmt.Errorf("SKE isn't enabled for this project, please run 'stackit ske enable'")
+				return &errors.ServiceDisabledError{
+					Service: "ske",
```
DiogoFerrao, line 97:
```suggestion
					Service: "SKE",
```

DiogoFerrao: I think we are using the capitalized `SKE` everywhere

internal/pkg/services/object-storage/utils/utils.go
```diff
@@ -3,15 +3,33 @@ package utils
...
+		if !ok {
+			return false, err
+		}
+		if oapiErr.StatusCode == http.StatusNotFound {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
```
DiogoFerrao, line 31: Great idea 💡

internal/pkg/errors/errors.go
```diff
@@ -133,6 +133,11 @@ The profile name can only contain lowercase letters, numbers, and "-" and cannot
 
 	USAGE_TIP = `For usage help, run:
   $ %s --help`
+
+	SERVICE_DISABLED_ERROR = `This service isn't enabled for the current project.
+
```
DiogoFerrao, line 138:
```suggestion
	SERVICE_DISABLED = `This service isn't enabled for the current project.

```

internal/cmd/ske/cluster/create/create.go
joaopalet: This is just used for the command suggestion to enable (as `stackit %s enable`), similarly to what we do in other errors

DiogoFerrao (approved):

PR #400, opened 2024-07-04
Add support for custom IDP endpoint and client ID

joaopalet:


internal/cmd/config/set/set.go
```diff
@@ -124,6 +126,8 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
 
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().String(sessionTimeLimitFlag, "", "Maximum time before authentication is required again. After this time, you will be prompted to login again to execute commands that require authentication. Can't be larger than 24h. Requires authentication after being set to take effect. Examples: 3h, 5h30m40s (BETA: currently values greater than 2h have no effect)")
+	cmd.Flags().String(identityProviderCustomEndpointFlag, "", "Identity Provider base URL, used for user authentication")
```
DiogoFerrao, line 129: I think that here we could say that the IDP needs to be OICD compliant (with a relevant link to the protocol documentation)

internal/cmd/config/set/set.go
```diff
@@ -124,6 +126,8 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
...
+	cmd.Flags().String(identityProviderCustomClientIdFlag, "", "Identity Provider client ID, used for user authentication")
```
DiogoFerrao, line 130: Also here, its not very clear what the `client ID` is, maybe we could link to some documentation as well

DiogoFerrao:
In general, I think we should document the process of configuring a custom IDP in a README file since its quite different from the others and could be complex

joaopalet:
> In general, I think we should document the process of configuring a custom IDP in a README file since its quite different from the others and could be complex.
> 
> Maybe this is the correct place to document the process, and then just refer the user to this guide in the flags, instead of cluttering the flag documentation

I've added a bit more information to the flags description and have asked a colleague from the IDP for a suggestion on documentation to link.

I'm not sure the README is the right place to document these, but definitely having a page with more information about it makes sense. Maybe creating a page in the STACKIT KB?

internal/cmd/config/set/set.go
joaopalet: As discussed, will only allow STACKIT domains for now

internal/cmd/config/set/set.go
joaopalet: Removed flag as it's not needed for other stages of the STACKIT IDP

internal/pkg/auth/utils_test.go
```diff
@@ -0,0 +1,57 @@
...
+			expected:          defaultIDPEndpoint,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			viper.Reset()
+
+			if tt.idpCustomEndpoint != "" {
+				viper.Set(config.IdentityProviderCustomEndpointKey, tt.idpCustomEndpoint)
+			}
```
DiogoFerrao, line 41: I think you can just set it, it will be empty

DiogoFerrao (approved):

PR #404, opened 2024-07-05
Migrate to new STACKIT IDP

joaopalet:
> [!WARNING]
> Should be merged in the **8th of July**, to be released in the **9th of July**, as announced [here](https://github.com/stackitcloud/stackit-cli#warning-new-stackit-idp)

### DIFF
--- a/internal/pkg/auth/user_login.go
+++ b/internal/pkg/auth/user_login.go
@@ -90,7 +90,7 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 		Endpoint: oauth2.Endpoint{
 			AuthURL: fmt.Sprintf("%s/authorize", idpEndpoint),
 		},
-		Scopes:      []string{"openid offline_access"},
+		Scopes:      []string{"openid offline_access email"},
 		RedirectURL: redirectURL,
 	}
 
--- a/internal/pkg/auth/user_login.go
+++ b/internal/pkg/auth/user_login.go
@@ -72,6 +72,7 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 	for i := range configuredPortRange {
 		port = defaultPort + i
 		portString := fmt.Sprintf(":%s", strconv.Itoa(port))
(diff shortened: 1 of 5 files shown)


internal/pkg/auth/user_login.go
```diff
@@ -60,22 +65,31 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
...
 
-	listener, err := net.Listen("tcp", ":0")
-	if err != nil {
-		return fmt.Errorf("bind port for login redirect: %w", err)
+	var redirectURL string
+	var listener net.Listener
+	var listenerErr error
+	var port int
+	for i := range configuredPortRange {
+		port = defaultPort + i
```
DiogoFerrao, line 73: I think we could add a Debug log at the start saying: `trying to bind port X for login redirect`

This way it also clarifies the debug log below `"unable to bind port %d for login redirect: %s"` by making it clear that we are trying different ports and it is not a failure (at least yet)

internal/pkg/auth/user_login.go
```diff
@@ -60,22 +65,31 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
...
 	}
-	redirectURL := fmt.Sprintf("http://localhost:%d", address.Port)
 
 	conf := &oauth2.Config{
 		ClientID: cliClientID,
 		Endpoint: oauth2.Endpoint{
 			AuthURL: fmt.Sprintf("%s/authorize", idpEndpoint),
 		},
-		Scopes:      []string{"openid"},
+		Scopes:      []string{"openid offline_access"},
```
DiogoFerrao, line 92: What is this?

joaopalet: The `offline_access` scope signals that we are also requesting a refresh token, that can be used to refresh the access token (https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess)

DiogoFerrao (approved):

DiogoFerrao (approved):

PR #422, opened 2024-07-22
Onboard IaaS network-area `create` and `update` commands

DiogoFerrao:


internal/pkg/orgname/orgname.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// GetOrganizationName returns the name of an organization by its ID.
+func GetOrganizationName(ctx context.Context, p *print.Printer, orgId string) (string, error) {
+	apiClient, err := client.ConfigureClient(p)
+	if err != nil {
+		return "", fmt.Errorf("configure resource manager client: %w", err)
+	}
+	req := apiClient.GetOrganization(ctx, orgId)
+	resp, err := req.Execute()
```
joaopalet, line 18: This is a new pattern in the CLI I think, right? We could also make this a util of the `resource-manager` service, and initialise the client in the `RunE` of the command which would be more in line with what we were doing until now. What do you think?

internal/cmd/beta/network-area/create/create.go
```diff
@@ -0,0 +1,196 @@
...
+	case print.YAMLOutputFormat:
+		details, err := yaml.MarshalWithOptions(networkArea, yaml.IndentSequence(true))
+		if err != nil {
+			return fmt.Errorf("marshal network area: %w", err)
+		}
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		p.Outputf("Created network area for project %q.\n", projectLabel)
```
joaopalet, line 193: We should also show the area ID here

DiogoFerrao: Forgot to update here

internal/pkg/orgname/orgname.go
DiogoFerrao: Making it part of the utils of the `resource-manager` makes sense. I followed the logic of the `projectname` package, that does something similar but for the project name. However, this `GetOrganizationName` would not be used as often, so it may be better to implement in a utils package

internal/cmd/beta/network-area/create/create.go
```diff
@@ -0,0 +1,203 @@
...
+			rmApiClient, err := rmClient.ConfigureClient(p)
+			if err == nil {
+				orgLabel, err = rmUtils.GetOrganizationName(ctx, rmApiClient, *model.OrganizationId)
+				if err != nil {
+					p.Debug(print.ErrorLevel, "get organization name: %v", err)
+					orgLabel = *model.OrganizationId
+				}
+			} else {
+				p.Debug(print.ErrorLevel, "configure resource manager client: %v", err)
+			}
```
joaopalet, line 89: Wouldn't it be simpler to have something like this?

```suggestion
rmApiClient, err := rmClient.ConfigureClient(p)
if err != nil {
	p.Debug(print.ErrorLevel, "configure resource manager client: %v", err)
} 
orgLabel, err := rmUtils.GetOrganizationName(ctx, rmApiClient, *modelOrganizationId)
if err != nil {
	p.Debug(print.ErrorLevel, "get organization name: %v", err)
	orgLabel = *model.OrganizationId
}
```

joaopalet (approved):

internal/cmd/beta/network-area/create/create.go
DiogoFerrao: Hmm yes, its simpler. But if the configuration of the client errors, the `GetOrganizationName` call won't work either and the user will get 2 errors, when it's really just the one about the client

joaopalet: Ah yes we don't want to return earlier

joaopalet (approved):

PR #431, opened 2024-07-26
Onboard `routes create` and `routes delete` command

DiogoFerrao:


internal/cmd/beta/network-area/routes/create/create.go
```diff
@@ -0,0 +1,181 @@
...
+	Nexthop        *string
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates a static route in a STACKIT Network Area (SNA)",
+		Long: fmt.Sprintf("%s\n%s\n",
+			"Creates a static route in a STACKIT Network Area (SNA).",
+			"This command is currently asynchonous only due to limitations in the waiting functionality of the SDK. This will be updated in a future release.",
```
joaopalet, line 42: Let's just try to confirm first with the team if the request is even async or not. Because the created routes already come in the API response, right?

internal/cmd/beta/network-area/routes/create/create.go
```diff
@@ -0,0 +1,181 @@
...
+// getRouteFromAPIResponse returns the static route from the API response that matches the prefix and nexthop
+// This works because static routes are unique by prefix and nexthop
+func getRouteFromAPIResponse(prefix, nexthop string, routes *[]iaas.Route) (iaas.Route, error) {
+	for _, route := range *routes {
+		if *route.Prefix == prefix && *route.Nexthop == nexthop {
+			return route, nil
+		}
+	}
+	return iaas.Route{}, fmt.Errorf("new static route not found in API response")
+}
```
joaopalet, line 181: This could be moved to the `utils` package

internal/cmd/beta/network-area/routes/create/create.go
joaopalet: Also the same applies for the delete, right?

joaopalet (approved):

PR #438, opened 2024-08-02
Onboard IaaS network

GokceGK:


internal/cmd/beta/network/create/create.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/spinner"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas/wait"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	nameFlag         = "name"
+	dnsServersFlag   = "dns-servers"
```
joaopalet, line 26: The Portal calls these `DNS Name Servers`, should we name it `dns-name-servers` instead so it matches?

internal/cmd/beta/network/create/create.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().StringP(nameFlag, "n", "", "Network name")
+	cmd.Flags().StringSlice(dnsServersFlag, []string{}, "List of DNS servers/nameservers IPs")
+	cmd.Flags().Int64(prefixLengthFlag, 0, "The default prefix length for networks")
```
joaopalet, line 109: Is it the _default_ prefix length or just the prefix length? Is there any way to change it for specific nameservers?

joaopalet: ALso, is it the prefix lenght of the _network_ or of the name servers?

internal/cmd/beta/network/create/create.go
```diff
@@ -0,0 +1,178 @@
...
+
+	payload := iaas.CreateNetworkPayload{
+		Name: model.Name,
+		AddressFamily: &iaas.CreateNetworkAddressFamily{
+			Ipv4: &iaas.CreateNetworkIPv4{
+				Nameservers:  model.DnsServers,
+				PrefixLength: model.PrefixLength,
+			},
+		},
+	}
```
joaopalet, line 151: The API seems to allow some more fields, e.g. `Ipv6` in the `AddressFamily`, `routed` at the root level, `gateway` inside `Ipv4` or `Ipv6`. Shouldn't we also have those in the CLI?

internal/cmd/beta/network/create/create.go
```diff
@@ -0,0 +1,178 @@
...
+		Short: "Creates a network",
+		Long:  "Creates a network.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a network with name "network-1"`,
+				`$ stackit beta network create --name network-1`,
+			),
+			examples.NewExample(
+				`Create a network with name "network-1" with dns servers and a prefix length`,
```
joaopalet, line 49:
```suggestion
				`Create a network with name "network-1" with DNS servers and a prefix length`,
```

internal/cmd/beta/network/describe/describe.go
```diff
@@ -0,0 +1,156 @@
...
+		table.AddRow("ID", *network.NetworkId)
+		table.AddSeparator()
+		table.AddRow("NAME", *network.Name)
+		table.AddSeparator()
+		table.AddRow("STATE", *network.State)
+		table.AddSeparator()
+		table.AddRow("PUBLIC IP", *network.PublicIp)
+		table.AddSeparator()
+		if len(nameservers) > 0 {
+			table.AddRow("NAME SERVERS", strings.Join(nameservers, ","))
```
joaopalet, line 142:
```suggestion
			table.AddRow("NAME SERVERS", strings.Join(nameservers, ", "))
```

internal/cmd/beta/network/describe/describe.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		table.AddSeparator()
+		if len(prefixes) > 0 {
+			table.AddRow("PREFIXES", strings.Join(prefixes, ","))
```
joaopalet, line 146:
```suggestion
			table.AddRow("PREFIXES", strings.Join(prefixes, ", "))
```

internal/cmd/beta/network/update/update.go
```diff
@@ -0,0 +1,156 @@
...
+		Short: "Updates a network",
+		Long:  "Updates a network.",
+		Args:  args.SingleArg(networkIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(
+				`Update network with ID "xxx" with new name "network-1-new"`,
+				`$ stackit beta network update xxx --name network-1-new`,
+			),
+			examples.NewExample(
+				`Update network with ID "xxx" with new name "network-1-new" and new dns servers`,
```
joaopalet, line 49:
```suggestion
				`Update network with ID "xxx" with new name "network-1-new" and new DNS servers`,
```

internal/cmd/beta/network/create/create.go
GokceGK: I think we should add those new fields. They were not in refinement, so I did not add but refinement might be outdated after api changes.

internal/cmd/beta/network/create/create.go
GokceGK: it is prefix length of the network. 
And it is just prefix length, in api docs and in TFP no mention of default. I will adapt the description

joaopalet (approved):

PR #444, opened 2024-08-07
Change network payload handling

GokceGK:


internal/cmd/beta/network/update/update.go
```diff
@@ -151,17 +151,27 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
+		}
+	}
+
+	if model.IPv4DnsNameServers != nil {
+		addressFamily = &iaas.UpdateNetworkAddressFamily{
+			Ipv4: &iaas.UpdateNetworkIPv4{
+				Nameservers: model.IPv4DnsNameServers,
+			},
+		}
+	}
```
joaopalet, line 170: This doesn't allow setting both, right? Or are they mutually exclusive?

joaopalet: Otherwise I think we can do something like
```go
addressFamily := &iaas.UpdateNetworkAddressFamily{}

if model.IPv4DnsNameServers != nil {
		addressFamily.Ipv4 = &iaas.V1UpdateNetworkIPv4{
				Nameservers: model.IPv4DnsNameServers,
        },
}

if model.IPv6DnsNameServers != nil {
		addressFamily.Ipv6 = &iaas.V1UpdateNetworkIPv6{
				Nameservers: model.IPv6DnsNameServers,
        },
}
```

internal/cmd/beta/network/update/update_test.go
```diff
@@ -80,9 +80,6 @@ func fixturePayload(mods ...func(payload *iaas.PartialUpdateNetworkPayload)) iaa
 			Ipv4: &iaas.UpdateNetworkIPv4{
 				Nameservers: utils.Ptr([]string{"1.1.1.0", "1.1.2.0"}),
 			},
-			Ipv6: &iaas.V1UpdateNetworkIPv6{
```
joaopalet, line 83: And then this could be re-added, as currently it's excluding it from the payload even though it's specified in the input model

internal/cmd/beta/network/create/create.go
joaopalet, line 1: Same for the `create`

internal/cmd/beta/network/update/update.go
GokceGK: as I understood from the API, in the future, it can be possible to have ipv4 and ipv6 at the same time in the network creation payload. But I will verify with the IaaS team. 

I think in the meantime, we can use your suggestion. That looks good

joaopalet (approved):

PR #451, opened 2024-08-19
Extend config options

GokceGK:


internal/cmd/auth/activate-service-account/activate_service_account.go
```diff
@@ -100,17 +100,18 @@ func configureFlags(cmd *cobra.Command) {
...
+	jwksCustomEndpoint := viper.GetString(config.JwksCustomEndpointKey)
+
 	model := inputModel{
 		ServiceAccountToken:   flags.FlagToStringValue(p, cmd, serviceAccountTokenFlag),
 		ServiceAccountKeyPath: flags.FlagToStringValue(p, cmd, serviceAccountKeyPathFlag),
 		PrivateKeyPath:        flags.FlagToStringValue(p, cmd, privateKeyPathFlag),
-		TokenCustomEndpoint:   flags.FlagToStringValue(p, cmd, tokenCustomEndpointFlag),
-		JwksCustomEndpoint:    flags.FlagToStringValue(p, cmd, jwksCustomEndpointFlag),
+		TokenCustomEndpoint:   tokenCustomEndpoint,
+		JwksCustomEndpoint:    jwksCustomEndpoint,
```
joaopalet, line 114: For the other custom endpoints (e.g. the services) we don't add them to the input model of the command, so for uniformity I would follow the same approach here (and just use the values directly when needed)

joaopalet: Alternatively we could consider adding the relevant custom endpoint (if any) to each command's input model

internal/cmd/config/set/set.go
```diff
@@ -131,6 +134,7 @@ func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().String(sessionTimeLimitFlag, "", "Maximum time before authentication is required again. After this time, you will be prompted to login again to execute commands that require authentication. Can't be larger than 24h. Requires authentication after being set to take effect. Examples: 3h, 5h30m40s (BETA: currently values greater than 2h have no effect)")
 	cmd.Flags().String(identityProviderCustomEndpointFlag, "", "Identity Provider base URL, used for user authentication")
 	cmd.Flags().String(identityProviderCustomClientIdFlag, "", "Identity Provider client ID, used for user authentication")
+	cmd.Flags().String(allowedUrlDomainFlag, "", "Domain name, used for the verification of the URLs that are given in the IDP endpoint and curl commands")
```
joaopalet, line 137:
```suggestion
	cmd.Flags().String(allowedUrlDomainFlag, "", `Domain name, used for the verification of the URLs that are given in the custom identidy provider endpoint and "STACKIT curl" command`)
```

internal/cmd/config/unset/unset.go
```diff
@@ -122,6 +128,9 @@ func NewCmd(p *print.Printer) *cobra.Command {
 			if model.IdentityProviderCustomClientID {
 				viper.Set(config.IdentityProviderCustomClientIdKey, "")
 			}
+			if model.AllowedUrlDomain {
```
joaopalet, line 131: Isn't the logic for the JWKS and token custom endpoints missing from this part?

internal/pkg/config/config.go
```diff
@@ -151,6 +157,7 @@ func setConfigDefaults() {
 	viper.SetDefault(SessionTimeLimitKey, SessionTimeLimitDefault)
 	viper.SetDefault(IdentityProviderCustomEndpointKey, "")
 	viper.SetDefault(IdentityProviderCustomClientIdKey, "")
+	viper.SetDefault(AllowedUrlDomainKey, "")
```
joaopalet, line 160: Couldn't we set the default value explicitly in the flag? Instead of living just in a global variable in the package?

joaopalet: This would also make it more transparent to the user

internal/pkg/utils/utils.go
```diff
@@ -51,7 +57,7 @@ func ConvertInt64PToFloat64P(i *int64) *float64 {
 	return &f
 }
 
-func ValidateSTACKITURL(value string) error {
+func ValidateURL(value string) error {
```
joaopalet, line 60:
```suggestion
func ValidateURLDomain(value string) error {
```

GokceGK: Renamed method

internal/cmd/auth/activate-service-account/activate_service_account.go
GokceGK: I have removed the endpoints from the input model and carried them to storeFlags function

internal/pkg/config/config.go
GokceGK: I changed the default value handling :)

internal/pkg/utils/utils.go
```diff
@@ -60,8 +62,15 @@ func ValidateSTACKITURL(value string) error {
...
 		return fmt.Errorf("bad url")
 	}
-	if !strings.HasSuffix(urlHost, "stackit.cloud") {
-		return fmt.Errorf(`only urls belonging to STACKIT are allowed, hostname must end in "stackit.cloud"`)
+
+	allowedUrlDomain := viper.GetString(config.AllowedUrlDomainKey)
+
+	if allowedUrlDomain == "" {
+		allowedUrlDomain = config.AllowedUrlDomainDefault
+	}
```
joaopalet, line 70: Now that a default is available this is probably not needed, unless we really want to disallow users to "allow any domain" by setting the flag to `""`

GokceGK: I think it should not be allowed to set it to empty.

vicentepinto98: In this case the user is setting explicitly the allowed domain to be empty, so yeah its like an "allow any domain" configuration. 
I think it could be misleading to just set the default in this case but I agree that it is a security risk. I would suggest a middle term:
- add a warning when the user sets this config option to empty, but not use the default here

GokceGK: removed condition and added warning

internal/cmd/config/set/set.go
```diff
@@ -131,6 +134,7 @@ func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().String(sessionTimeLimitFlag, "", "Maximum time before authentication is required again. After this time, you will be prompted to login again to execute commands that require authentication. Can't be larger than 24h. Requires authentication after being set to take effect. Examples: 3h, 5h30m40s (BETA: currently values greater than 2h have no effect)")
 	cmd.Flags().String(identityProviderCustomEndpointFlag, "", "Identity Provider base URL, used for user authentication")
 	cmd.Flags().String(identityProviderCustomClientIdFlag, "", "Identity Provider client ID, used for user authentication")
+	cmd.Flags().String(allowedUrlDomainFlag, "", `Domain name, used for the verification of the URLs that are given in the custom identidy provider endpoint and "STACKIT curl" command`)
```
vicentepinto98, line 137:
```suggestion
	cmd.Flags().String(allowedUrlDomainFlag, "", `Domain name, used for the verification of the URLs that are given in the custom identity provider endpoint and "STACKIT curl" command`)
```

vicentepinto98 (approved):

PR #452, opened 2024-08-20
Add error description to output

GokceGK:


internal/pkg/auth/user_login.go
```diff
@@ -135,8 +135,9 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 
 		// Get the authorization code
 		code := r.URL.Query().Get("code")
+		errDescription := r.URL.Query().Get("error_description")
 		if code == "" {
-			errServer = fmt.Errorf("could not find 'code' URL parameter")
+			errServer = fmt.Errorf("could not find 'code' URL parameter.\nError: %s", errDescription)
```
vicentepinto98, line 140:
```suggestion
			errServer = fmt.Errorf("could not find 'code' URL parameter: %s", errDescription)
```

And probably also check and only concatenate `errDescription` if it is not empty

GokceGK: Added the check

internal/pkg/auth/user_login.go
```diff
@@ -135,8 +135,12 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 
 		// Get the authorization code
 		code := r.URL.Query().Get("code")
+		errDescription := r.URL.Query().Get("error_description")
 		if code == "" {
 			errServer = fmt.Errorf("could not find 'code' URL parameter")
+			if errDescription != "" {
+				errServer = fmt.Errorf("%w.\nError: %s", errServer, errDescription)
```
vicentepinto98, line 142:
```suggestion
				errServer = fmt.Errorf("%w: %s", errServer, errDescription)
```

vicentepinto98: better to not introduce a paragraph and also "Error" is redundant

vicentepinto98 (approved):

PR #502, opened 2024-11-11
Add update command to network-area route

Fyusel:
Support to update labels for a network-area route.

### DIFF
--- a/internal/cmd/beta/network-area/route/update/update.go
+++ b/internal/cmd/beta/network-area/route/update/update.go
@@ -1,23 +1,22 @@
-package describe
+package update
 
 import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	"github.com/goccy/go-yaml"
+	"github.com/spf13/cobra"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/examples"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/flags"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/services/iaas/client"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/tables"
+	iaasUtils "github.com/stackitcloud/stackit-cli/internal/pkg/services/iaas/utils"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
 	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
-
-	"github.com/spf13/cobra"
 )
 
 const (
@@ -25,6 +24,8 @@ const (
 
 	organizationIdFlag = "organization-id"
 	networkAreaIdFlag  = "network-area-id"
+
+	labelFlag = "labels"
 )
 
 type inputModel struct {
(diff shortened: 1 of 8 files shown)


internal/cmd/beta/network-area/route/describe/describe.go
```diff
@@ -137,6 +138,19 @@ func outputResult(p *print.Printer, outputFormat string, route *iaas.Route) erro
...
+		for _, v := range *route.GetLabels() {
+			count++
+			if value, ok := v.(string); ok {
+				sb.WriteString(value)
+				if len(*route.GetLabels()) != count {
+					sb.WriteString(" ")
+				}
+			}
+		}
+		table.AddRow("Labels", sb.String())
```
joaopalet, line 153: Here I think we could avoid the strings.Builder and make it something like this. I think it's simpler to read and also more inline with other commands


```go
if route.Labels != nil && len(*route.Labels) > 0 {
    labels := []string{}
    for key, value := range *route.Labels {
        labels = append(labels, fmt.Sprintf("%s: %s", key, value))
    }

    table.AddSeparator()
    table.AddRow("Labels", strings.Join(labels, "\n"))
}
```

joaopalet: Also, this has two other benefits:
- Only shows the Labels row if there is at least a label
- Shows both key and value of each label, one per row

internal/cmd/beta/network-area/route/describe/describe.go
```diff
@@ -137,6 +138,19 @@ func outputResult(p *print.Printer, outputFormat string, route *iaas.Route) erro
 		table.AddRow("Prefix", *route.Prefix)
 		table.AddSeparator()
 		table.AddRow("Nexthop", *route.Nexthop)
```
joaopalet, line 140: Btw, could you make all row names capitalized? We have it like that in all other commands, not sure why this one wasn't 😬

joaopalet: Like `table.AddRow("PREFIX", *route.Prefix)`

internal/cmd/beta/network-area/route/update/update.go
```diff
@@ -0,0 +1,166 @@
...
+		Short: "Updates a static route in a STACKIT Network Area (SNA)",
+		Long: fmt.Sprintf("%s\n%s\n",
+			"Updates a static route in a STACKIT Network Area (SNA).",
+			"This command is currently asynchonous only due to limitations in the waiting functionality of the SDK. This will be updated in a future release.",
+		),
+		Args: args.SingleArg(routeIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(
+				`Updates the label(s) of a static route with ID "xxx" in a STACKIT Network Area with ID "yyy" in organization with ID "zzz"`,
+				"$ stackit beta network-area route update xxx --label key=value,foo=bar --organization-id yyy --network-area-id zzz",
```
joaopalet, line 51: Since we have the flag in the singular form:
```suggestion
				"$ stackit beta network-area route update xxx --label key=value --label foo=bar --organization-id yyy --network-area-id zzz",
```

joaopalet: Alternatively, and I think this could even make it more clear, the flag could have the name in the plural `labels`

joaopalet (commented):
Overall very nice first PR, thanks for having a look at how things are done in other commands and following the same here! This is really the idea for the CLI codebase

internal/cmd/beta/network-area/route/update/update.go
Fyusel: Changed to to `labels`

internal/cmd/beta/network-area/route/describe/describe.go
```diff
@@ -134,9 +135,18 @@ func outputResult(p *print.Printer, outputFormat string, route *iaas.Route) erro
...
+		table.AddRow("NEXTHOP", *route.Nexthop)
+		table.AddSeparator()
+		if route.Labels != nil && len(*route.Labels) > 0 {
+			labels := []string{}
+			for key, value := range *route.Labels {
+				labels = append(labels, fmt.Sprintf("%s: %s", key, value))
+			}
+			table.AddSeparator()
+			table.AddRow("LABELS", strings.Join(labels, "\n"))
+		}
```
joaopalet, line 149: The separation from line 141 is not needed, right? As if the network has labels, it will be added in line 147 already

Fyusel: Yes, you are right :+1: 
Line removed

joaopalet (approved):

PR #503, opened 2024-11-13
Onboard IaaS volume commands

GokceGK:
Volume

- list
- describe
- create
- delete
- resize
- update
- performance-class
       - list
       - describe

### DIFF
--- a/internal/cmd/beta/volume/create/create.go
+++ b/internal/cmd/beta/volume/create/create.go
@@ -53,16 +53,16 @@ func NewCmd(p *print.Printer) *cobra.Command {
 		Args:  args.NoArgs,
 		Example: examples.Build(
 			examples.NewExample(
-				`Create a volume with availability zone "eu01-1"`,
-				`$ stackit beta volume create --availability-zone eu01-1`,
+				`Create a volume with availability zone "eu01-1" and size 64 GB`,
+				`$ stackit beta volume create --availability-zone eu01-1 --size 64`,
 			),
 			examples.NewExample(
-				`Create a volume with name "volume-1"`,
-				`$ stackit beta volume create --availability-zone eu01-1 --name volume-1`,
+				`Create a volume with name "volume-1", source id "xxx" and type "image"`,
+				`$ stackit beta volume create --availability-zone eu01-1 --name volume-1 --source-id xxx --source-type image`,
 			),
 			examples.NewExample(
-				`Create a volume with availability zone "eu01-1", performance class "storage_premium_perf1"`,
-				`$ stackit beta volume create --availability-zone eu01-1 --performance-class storage_premium_perf1`,
+				`Create a volume with availability zone "eu01-1", performance class "storage_premium_perf1" and size 64 GB`,
+				`$ stackit beta volume create --availability-zone eu01-1 --performance-class storage_premium_perf1 --size 64`,
 			),
 		),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/internal/cmd/beta/volume/create/create.go
+++ b/internal/cmd/beta/volume/create/create.go
@@ -57,7 +57,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				`$ stackit beta volume create --availability-zone eu01-1 --size 64`,
 			),
 			examples.NewExample(
-				`Create a volume with name "volume-1", source id "xxx" and type "image"`,
+				`Create a volume with name "volume-1", from a source image with ID "xxx"`,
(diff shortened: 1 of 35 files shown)


internal/cmd/beta/volume/create/create.go
```diff
@@ -0,0 +1,219 @@
...
+				`$ stackit beta volume create --availability-zone eu01-1`,
+			),
+			examples.NewExample(
+				`Create a volume with name "volume-1"`,
+				`$ stackit beta volume create --availability-zone eu01-1 --name volume-1`,
+			),
+			examples.NewExample(
+				`Create a volume with availability zone "eu01-1", performance class "storage_premium_perf1"`,
+				`$ stackit beta volume create --availability-zone eu01-1 --performance-class storage_premium_perf1`,
+			),
```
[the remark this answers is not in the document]
joaopalet: Either way, we should have an example to create a volume given a size, at least one from an image for example

internal/cmd/beta/volume/describe/describe.go
```diff
@@ -0,0 +1,156 @@
...
+		table.AddRow("VOLUME SIZE (GB)", *volume.Size)
+		table.AddSeparator()
+		table.AddRow("PERFORMANCE CLASS", *volume.PerformanceClass)
+		table.AddSeparator()
+		table.AddRow("AVAILABILITY ZONE", *volume.AvailabilityZone)
+		table.AddSeparator()
+		table.AddRow("SOURCE", sourceId)
+		table.AddSeparator()
+		table.AddRow("SERVER", serverId)
+		table.AddSeparator()
```
joaopalet, line 148: This will show these field even if they're empty, right? We can show conditionally

joaopalet: Also we could show the size, if the field is present

internal/cmd/beta/volume/list/list.go
```diff
@@ -0,0 +1,181 @@
...
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Lists all volumes`,
+				"$ stackit beta volume list",
+			),
+			examples.NewExample(
+				`Lists all volumes which contains the label xxx`,
+				"$ stackit beta volume list --label-selector xxx",
+			),
```
joaopalet, line 48: 👍

internal/cmd/beta/volume/list/list.go
```diff
@@ -0,0 +1,181 @@
...
+		details, err := yaml.MarshalWithOptions(volumes, yaml.IndentSequence(true))
+		if err != nil {
+			return fmt.Errorf("marshal volume: %w", err)
+		}
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		table := tables.NewTable()
+		table.SetHeader("ID", "Name", "Status", "Server", "Public IP")
```
joaopalet, line 167: The header columns are not matching the row values

internal/cmd/beta/volume/update/update.go
```diff
@@ -0,0 +1,173 @@
...
+	payload := iaas.UpdateVolumePayload{
+		Name:        model.Name,
+		Description: model.Description,
+		Labels:      labelsMap,
+	}
+
+	return req.UpdateVolumePayload(payload)
+}
+
+func outputResult(p *print.Printer, model *inputModel, projectLabel string, volume *iaas.Volume) error {
```
joaopalet, line 151: From the `RunE`, we are passing the volume name as third argument (and not the project label)

joaopalet: The output text in `default:` needs to be adjusted also

internal/cmd/beta/volume/resize/resize.go
```diff
@@ -0,0 +1,128 @@
...
+			p.Outputf("Resized volume %q.\n", volumeLabel)
+			return nil
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Int64(sizeFlag, 0, "Volume size")
```
joaopalet, line 88:
```suggestion
	cmd.Flags().Int64(sizeFlag, 0, "Volume size (GB)")
```

internal/cmd/beta/volume/describe/describe.go
GokceGK: I think volume will always have a size even if we dont parse it in creation

internal/cmd/beta/volume/create/create.go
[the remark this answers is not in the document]
GokceGK: nice catch, thx. Edited the examples to cover both cases

joaopalet (approved):

PR #504, opened 2024-11-14
IaaS volume command changes

GokceGK:


internal/cmd/beta/volume/create/create.go
```diff
@@ -122,7 +126,7 @@ func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().String(availabilityZoneFlag, "", "Availability zone")
 	cmd.Flags().StringP(nameFlag, "n", "", "Volume name")
 	cmd.Flags().String(descriptionFlag, "", "Volume description")
-	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value string pairs which can be attached to a volume. A label can be provided with the format key=value and the flag can be used multiple times to provide a list of labels")
+	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value string pairs which can be attached to a volume. A label can be provided with the format key=value. To provide a list of labels, key=value pairs must be separated by commas(,) e.g. --labels key=value,foo=bar.")
```
joaopalet, line 129: I think we could have this shorter, as we already have an example in the command examples

```suggestion
	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value string pairs which can be attached to a volume. E.g., '--labels key1=value1,key2=value2,...'")
```

internal/cmd/beta/volume/update/update.go
```diff
@@ -96,7 +100,7 @@ func NewCmd(p *print.Printer) *cobra.Command {
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP(nameFlag, "n", "", "Volume name")
 	cmd.Flags().String(descriptionFlag, "", "Volume description")
-	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value string pairs which can be attached to a volume. A label can be provided with the format key=value and the flag can be used multiple times to provide a list of labels")
+	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value string pairs which can be attached to a volume. A label can be provided with the format key=value. To provide a list of labels, key=value pairs must be separated by commas(,) e.g. --labels key=value,foo=bar.")
```
joaopalet, line 103: Same here

GokceGK: applied and generated docs :)

joaopalet (approved):

PR #505, opened 2024-11-15
Onboard IaaS network interface command

Fyusel:
Supported subcommands: create, delete, update, list, describe


internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,242 @@
...
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Var(flags.UUIDFlag(), networkIdFlag, "Network ID")
+	cmd.Flags().StringSlice(allowedAddressesFlag, nil, "List of allowed IPs")
+	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value string pairs which can be attached to a network-interface. A label can be provided with the format key=value and the flag can be used multiple times to provide a list of labels")
+	cmd.Flags().StringP(nameFlag, "n", "", "Network interface name")
+	cmd.Flags().BoolP(nicSecurityFlag, "b", defaultEnabled, "If this is set to false, then no security groups will apply to this network interface.")
```
DiogoFerrao, line 108: I think that `defaultNicSecurityFlag` is a better name than `defaultEnabled` here

internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+	}
+
+	// check security groups size and regex
+	securityGroups := flags.FlagToStringSlicePointer(p, cmd, securityGroupsFlag)
+	if securityGroups != nil && len(*securityGroups) > 0 {
+		securityGroupsRegex := regexp.MustCompile(securityGroupsRegex)
+		// iterate over them
+		for _, value := range *securityGroups {
+			if len(value) != 36 {
```
DiogoFerrao, line 157: It would be better if this `36` would be extracted to a `const` declared above, maybe even with a comment explaining that it is the current API limit

internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,242 @@
...
+	if globalFlags.ProjectId == "" {
+		return nil, &errors.ProjectIdError{}
+	}
+
+	allowedAddresses := flags.FlagToStringSlicePointer(p, cmd, allowedAddressesFlag)
+	var allowedAddressesInner []iaas.AllowedAddressesInner
+	if allowedAddresses != nil && len(*allowedAddresses) > 0 {
+		allowedAddressesInner = make([]iaas.AllowedAddressesInner, len(*allowedAddresses))
+		for k, v := range *allowedAddresses {
+			// linter issue
```
DiogoFerrao, line 127: Is this intended?

internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,242 @@
...
+			value := v
+			allowedAddressesInner[k].String = &value
```
DiogoFerrao, line 129: Can't this just be `allowedAddressesInner[k].String = &v`?

DiogoFerrao: Although maybe `address` is a better name than `v` for this local variable

internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+	}
+
+	// check name length <= 63 and regex must apply
+	name := flags.FlagToStringPointer(p, cmd, nameFlag)
+	if name != nil {
+		if len(*name) > 63 {
```
DiogoFerrao, line 136: Also here, this could be extracted to a const, including the error message

internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,242 @@
...
+	req := apiClient.UpdateNic(ctx, model.ProjectId, *model.NetworkId, model.NicId)
+
+	var labelsMap *map[string]interface{}
+	if model.Labels != nil && len(*model.Labels) > 0 {
+		// convert map[string]string to map[string]interface{}
+		labelsMap = utils.Ptr(map[string]interface{}{})
+		for k, v := range *model.Labels {
+			(*labelsMap)[k] = v
+		}
+	}
```
DiogoFerrao, line 208:
```suggestion
	var labelsMap *map[string]interface{}
	if model.Labels != nil && len(*model.Labels) > 0 {
		// convert map[string]string to map[string]interface{}
        convertedMap := make(map[string]interface{}, len(*model.Labels))
        for k, v := range *model.Labels {
            convertedMap[k] = v
        }
        labelsMap = &convertedMap
	}
```

DiogoFerrao: Sorry, very minor gripe here, but you can directly make with the needed size and also avoid the ugly `(*x)` assignment

internal/cmd/beta/network-interface/update/update_test.go
```diff
@@ -0,0 +1,286 @@
...
+			isValid: false,
+		},
+		{
+			description: "name invalid",
+			argValues:   fixtureArgValues(),
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[nameFlag] = "test?"
+			}),
+			isValid: false,
+		},
```
DiogoFerrao, line 182: Is an empty `name` valid? If not, it could also be included in the testing

internal/cmd/beta/network-interface/list/list.go
```diff
@@ -0,0 +1,201 @@
...
+		if err != nil {
+			return fmt.Errorf("marshal nics: %w", err)
+		}
+		p.Outputln(string(details))
+
+		return nil
+	case print.YAMLOutputFormat:
+		details, err := yaml.MarshalWithOptions(nics, yaml.IndentSequence(true))
+		if err != nil {
+			return fmt.Errorf("marshal nic: %w", err)
```
DiogoFerrao, line 162:
```suggestion
			return fmt.Errorf("marshal nics: %w", err)
```

internal/cmd/beta/network-interface/list/list.go
```diff
@@ -0,0 +1,201 @@
...
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		table := tables.NewTable()
+		table.SetHeader("ID", "Name", "Allowed Addresses", "IPv4", "IPv6", "Nic Security", "Security Groups", "Status", "Type")
+
+		for _, nic := range nics {
+			name := ""
+			allowedAddreses := []string{}
```
DiogoFerrao, line 173:
```suggestion
			allowedAddresses := []string{}
```

internal/cmd/beta/network-interface/describe/describe.go
```diff
@@ -0,0 +1,181 @@
...
+	cmd := &cobra.Command{
+		Use:   "describe",
+		Short: "Describes a network interface",
+		Long:  "Describes a network interface.",
+		Args:  args.SingleArg(nicIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(
+				`Describes network interface with nic id "xxx" and network ID "yyy"`,
+				`$ stackit beta network-interface describe xxx --network-id yyy`,
+			),
```
DiogoFerrao, line 45: You could also include an example using the `--output` flag, to maintain consistency with other commands

internal/cmd/beta/network-interface/describe/describe.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		if nic.Ipv6 != nil {
+			table.AddRow("IPV6", *nic.Ipv6)
+			table.AddSeparator()
+		}
+		table.AddRow("MAC", *nic.Mac)
+		table.AddSeparator()
+		table.AddRow("NIC SECURITY", *nic.NicSecurity)
+		if nic.AllowedAddresses != nil && len(*nic.AllowedAddresses) > 0 {
+			allowedAddreses := []string{}
```
DiogoFerrao, line 151:
```suggestion
			allowedAddresses := []string{}
```

internal/cmd/beta/network-interface/describe/describe.go
```diff
@@ -0,0 +1,181 @@
...
+			for _, value := range *nic.AllowedAddresses {
+				allowedAddreses = append(allowedAddreses, *value.String)
+			}
```
DiogoFerrao, line 154: This should be protected by a nil check, I think the `AllowedAddresses` could be `nil`, right?

internal/cmd/beta/network-interface/describe/describe.go
```diff
@@ -0,0 +1,181 @@
...
+			table.AddSeparator()
+			table.AddRow("ALLOWED ADDRESSES", strings.Join(allowedAddreses, "\n"))
+		}
+		if nic.Labels != nil && len(*nic.Labels) > 0 {
+			labels := []string{}
+			for key, value := range *nic.Labels {
+				labels = append(labels, fmt.Sprintf("%s: %s", key, value))
+			}
```
DiogoFerrao, line 162: Same here

internal/cmd/beta/network-interface/delete/delete.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+
+			// Configure API client
+			apiClient, err := client.ConfigureClient(p)
+			if err != nil {
+				return err
+			}
+
+			if !model.AssumeYes {
+				prompt := fmt.Sprintf("Are you sure you want to delete the network interface %q?", model.NicId)
```
DiogoFerrao, line 56:
```suggestion
				prompt := fmt.Sprintf("Are you sure you want to delete the network interface %q? (This cannot be undone)", model.NicId)
```

DiogoFerrao: Mostly for consistency, we also do it in the other commands

internal/cmd/beta/network-interface/create/create.go
```diff
@@ -0,0 +1,247 @@
...
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates a network interface",
+		Long:  "Creates a network interface.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a network interface with allowed addresses "1.1.1.1,8.8.8.8,9.9.9.9" with ipv4 address "1.2.3.4" with label "key=value" with name "NAME" with security groups "UUID1,UUID2" with nic security enabled for network with ID "xxx"`,
+				`$ stackit beta network-interface create --network-id xxx --allowed-addresses "1.1.1.1,8.8.8.8,9.9.9.9" --ipv4 1.2.3.4 --labels key=value --name NAME --security-groups "UUID1,UUID2" --nic-security"`,
```
DiogoFerrao, line 58:
```suggestion
				`$ stackit beta network-interface create --network-id xxx --allowed-addresses "1.1.1.1,8.8.8.8,9.9.9.9" --ipv4 1.2.3.4 --labels key=value,key2=value2 --name NAME --security-groups "UUID1,UUID2" --nic-security"`,
```

DiogoFerrao: I think it's useful to show what the expected format is when providing multiple labels

internal/cmd/beta/network-interface/create/create.go
```diff
@@ -0,0 +1,247 @@
...
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates a network interface",
+		Long:  "Creates a network interface.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a network interface with allowed addresses "1.1.1.1,8.8.8.8,9.9.9.9" with ipv4 address "1.2.3.4" with label "key=value" with name "NAME" with security groups "UUID1,UUID2" with nic security enabled for network with ID "xxx"`,
```
DiogoFerrao, line 57:
```suggestion
				`Create a network interface with allowed addresses, an IPv4 address, labels, a name, security groups and nic security enabled for network with ID "xxx"`,
```

DiogoFerrao: For more complicated commands such as this, I think it's better to ommit the specific values for each flag in the description of the example (see for instance the `network-area create`)

internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,242 @@
...
+			examples.NewExample(
+				`Updates a network interface with nic id "xxx" and network-id "yyy" with new allowed addresses "1.1.1.1,8.8.8.8,9.9.9.9" and with new label "key=value"`,
+				`$ stackit beta network-interface update xxx --network-id yyy --allowed-addresses "1.1.1.1,8.8.8.8,9.9.9.9" --labels key=value"`,
+			),
+			examples.NewExample(
+				`Updates a network interface with nic id "xxx" and network-id "yyy" with new name "nic-name-new"`,
+				`$ stackit beta network-interface update xxx --network-id yyy --name nic-name-new"`,
+			),
+			examples.NewExample(
+				`Updates a network interface with nic id "xxx" and network-id "yyy" with security groups removed and with nic-security disabled`,
```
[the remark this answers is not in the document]
DiogoFerrao: In general, in english you can ommit the repeated `with` if it is in a sequence, which makes the sentence less repetitive and verbose. This applies to a few of the example descriptions, I only suggested changes for some

internal/cmd/beta/network-interface/create/create.go
```diff
@@ -0,0 +1,247 @@
...
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Var(flags.UUIDFlag(), networkIdFlag, "Network ID")
+	cmd.Flags().StringSlice(allowedAddressesFlag, nil, "List of allowed IPs")
+	cmd.Flags().StringP(ipv4Flag, "i", "", "IPv4 address")
+	cmd.Flags().StringP(ipv6Flag, "s", "", "IPv6 address")
+	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value string pairs which can be attached to a network-interface. A label can be provided with the format key=value and the flag can be used multiple times to provide a list of labels")
```
DiogoFerrao, line 109:
```suggestion
	cmd.Flags().StringToString(labelFlag, nil, "Labels are key-value pairs which can be attached to a network-interface. The expected value is a comma separated string (e.g 'key1=value1,key2=value2')" 
```

DiogoFerrao: Just a suggestion, but I think we should try to keep these flag descriptions as short as possible and providing an example is always a very good way of showing the expected format

internal/cmd/beta/network-interface/create/create.go
```diff
@@ -0,0 +1,247 @@
...
+			// linter issue
+			value := v
+			allowedAddressesInner[k].String = &value
+		}
+	}
+
+	// check name length <= 63 and regex must apply
+	name := flags.FlagToStringPointer(p, cmd, nameFlag)
+	if name != nil {
+		if len(*name) > 63 {
```
DiogoFerrao, line 138: Same comment for the const

internal/cmd/beta/network-interface/create/create.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+	}
+
+	// check security groups size and regex
+	securityGroups := flags.FlagToStringSlicePointer(p, cmd, securityGroupsFlag)
+	if securityGroups != nil && len(*securityGroups) > 0 {
+		securityGroupsRegex := regexp.MustCompile(securityGroupsRegex)
+		// iterate over them
+		for _, value := range *securityGroups {
+			if len(value) != 36 {
```
DiogoFerrao, line 159: And also here

internal/cmd/beta/network-interface/create/create.go
```diff
@@ -0,0 +1,247 @@
...
+func buildRequest(ctx context.Context, model *inputModel, apiClient *iaas.APIClient) iaas.ApiCreateNicRequest {
+	req := apiClient.CreateNic(ctx, model.ProjectId, *model.NetworkId)
+
+	var labelsMap *map[string]interface{}
+	if model.Labels != nil && len(*model.Labels) > 0 {
+		// convert map[string]string to map[string]interface{}
+		labelsMap = utils.Ptr(map[string]interface{}{})
+		for k, v := range *model.Labels {
+			(*labelsMap)[k] = v
+		}
```
DiogoFerrao, line 210: Same suggestion here

internal/cmd/beta/network-interface/update/update.go
Fyusel: I got a linter issue from CI: 

```
G601: Implicit memory aliasing in for loop. (gosec)
			allowedAddressesInner[k].String = &v
```
This linter issue should disappear with Go 1.22 but I just saw that we are using Go 1.22. I will try to find out why the linter is still complaining about this (the reassigning is a suggested workaround to prevent the linter issue).

internal/cmd/beta/network-interface/update/update_test.go
Fyusel: `name` is optional but an empty string "" is not valid. I will add this as well to the test.

internal/cmd/beta/network-interface/create/create.go
Fyusel: :+1:  I just copied this from the project command. 
There is a different description for the volume command and the same for the route command (but this is because I implemented the route command as well and copied the description from project command too).

I think we should use the same description everywhere. What do you think?

internal/cmd/beta/network-interface/update/update.go
Fyusel: It seems to be that we need to update the linter at least to version 1.60.3 to prevent this to pop up. When updating locally there are more linter issues raised which needs to be fixed as well (also some of the configured linters are deprecated and need to be replaced/removed).

How is this done on your side? I would do a different PR for this topic in general. We can keep this PR open until the linter is updated in order to not push the workaround within this PR (or not to forget to remove the workaround afterwards).

internal/cmd/beta/network-interface/describe/describe.go
DiogoFerrao: Hmm, yeah sorry my bad, don't know what happened there

internal/cmd/beta/network-interface/describe/describe.go
DiogoFerrao: Also same here, my mistake

internal/cmd/beta/network-interface/update/update.go
```diff
@@ -0,0 +1,245 @@
...
+				`Updates a network interface with nic id "xxx" and network-id "yyy" to new allowed addresses "1.1.1.1,8.8.8.8,9.9.9.9" and new labels "key=value,key2=value2"`,
+				`$ stackit beta network-interface update xxx --network-id yyy --allowed-addresses "1.1.1.1,8.8.8.8,9.9.9.9" --labels key=value,key2=value2"`,
+			),
+			examples.NewExample(
+				`Updates a network interface with nic id "xxx" and network-id "yyy" with new name "nic-name-new"`,
+				`$ stackit beta network-interface update xxx --network-id yyy --name nic-name-new"`,
+			),
+			examples.NewExample(
+				`Updates a network interface with nic id "xxx" and network-id "yyy" to include the security group "3bc3cdf8-5270-4aa1-a5c2-e301619d5333"`,
+				`$ stackit beta network-interface update xxx --network-id yyy --security-groups "3bc3cdf8-5270-4aa1-a5c2-e301619d5333""`,
```
DiogoFerrao, line 67:
```suggestion
				`$ stackit beta network-interface update xxx --network-id yyy --security-groups "3bc3cdf8-5270-4aa1-a5c2-e301619d5333"`,
```

internal/cmd/beta/network-interface/update/update.go
DiogoFerrao: I understand the comment now, it was a work-around for the linter issue

internal/cmd/beta/network-interface/update/update.go
DiogoFerrao: As discussed, I think your suggestion is the best approach. Lets updated the linter, merge that PR and then come back to this issue

internal/cmd/beta/network-interface/create/create.go
Fyusel: I changed it to the `volume` version. Now we only need to change the `project` as well

internal/cmd/beta/network-interface/update/update.go
Fyusel: Linter was updated, I change the corresponding lines accordingly.

DiogoFerrao (approved):

GokceGK (approved):

PR #507, opened 2024-11-19
feat: Update golangci-lint version to 1.62.0 and go version to 1.23

Fyusel:
Updated Linter and go version.
Removed deprecated linter:
- exportloopref 
- structcheck
- deadcode
- varcheck

### DIFF
--- a/internal/pkg/cache/cache.go
+++ b/internal/pkg/cache/cache.go
@@ -43,7 +43,7 @@ func PutObject(identifier string, data []byte) error {
 		return ErrorInvalidCacheIdentifier
 	}
 
-	err := os.MkdirAll(cacheFolderPath, os.ModePerm)
+	err := os.MkdirAll(cacheFolderPath, 0o750)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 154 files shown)


internal/pkg/cache/cache.go
```diff
@@ -43,7 +43,7 @@ func PutObject(identifier string, data []byte) error {
 		return ErrorInvalidCacheIdentifier
 	}
 
-	err := os.MkdirAll(cacheFolderPath, os.ModePerm)
+	err := os.MkdirAll(cacheFolderPath, 0o750)
```
DiogoFerrao, line 46: What is the linting error here? Merely curiosity

Fyusel: internal/pkg/cache/cache.go:46:9: G301: Expect directory permissions to be 0750 or less (gosec)
	err := os.MkdirAll(cacheFolderPath, os.ModePerm)
	
ModePerm seems to set it to 777

DiogoFerrao (approved):

PR #517, opened 2024-11-28
Extend network commands with new fields

GokceGK:


internal/cmd/beta/network/create/create.go
```diff
@@ -113,11 +135,20 @@ func NewCmd(p *print.Printer) *cobra.Command {
 
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP(nameFlag, "n", "", "Network name")
-	cmd.Flags().StringSlice(ipv4DnsNameServersFlag, []string{}, "List of DNS name servers for IPv4")
+	cmd.Flags().StringSlice(ipv4DnsNameServersFlag, []string{}, "List of DNS name servers for IPv4. Nameservers cannot be defined for routed networks.")
 	cmd.Flags().Int64(ipv4PrefixLengthFlag, 0, "The prefix length of the IPv4 network")
-	cmd.Flags().StringSlice(ipv6DnsNameServersFlag, []string{}, "List of DNS name servers for IPv6")
+	cmd.Flags().String(ipv4PrefixFlag, "", "The IPv4 prefix of the network (CIDR)")
+	cmd.Flags().String(ipv4GatewayFlag, "", "The IPv4 gateway of a network. If not specified, the first IP of the network will be assigned as the gateway. If 'null' is sent, then the network doesn't have a gateway.")
```
joaopalet, line 141: This should be adjusted here, for no network we have the `--no-ipv4-network` flag

internal/cmd/beta/network/create/create.go
```diff
@@ -113,11 +135,20 @@ func NewCmd(p *print.Printer) *cobra.Command {
...
+	cmd.Flags().StringSlice(ipv6DnsNameServersFlag, []string{}, "List of DNS name servers for IPv6. Nameservers cannot be defined for routed networks.")
 	cmd.Flags().Int64(ipv6PrefixLengthFlag, 0, "The prefix length of the IPv6 network")
+	cmd.Flags().String(ipv6PrefixFlag, "", "The IPv6 prefix of the network (CIDR)")
+	cmd.Flags().String(ipv6GatewayFlag, "", "The IPv6 gateway of a network. If not specified, the first IP of the network will be assigned as the gateway. If 'null' is sent, then the network doesn't have a gateway.")
```
joaopalet, line 145: Same here

internal/cmd/beta/network/create/create.go
```diff
@@ -113,11 +135,20 @@ func NewCmd(p *print.Printer) *cobra.Command {
...
+	cmd.Flags().Bool(routedFlag, false, "If set to true, the network is routed and therefore accessible from other networks")
+	cmd.Flags().Bool(noIpv4Gateway, false, "If set to true, the network doesn't have an IPv4 gateway.")
```
joaopalet, line 147: Soma flag descriptions are ending in `.` and other not. I think usually we end them without punctuation across other commands

internal/cmd/beta/network/list/list.go
```diff
@@ -151,15 +151,20 @@ func outputResult(p *print.Printer, outputFormat string, networks []iaas.Network
...
 			publicIp := ""
 			if network.PublicIp != nil {
 				publicIp = *network.PublicIp
 			}
 
-			table.AddRow(*network.NetworkId, *network.Name, *network.State, publicIp)
+			routed := false
+			if network.Routed != nil {
+				routed = *network.Routed
+			}
```
joaopalet, line 165: :+1:

internal/cmd/beta/network/create/create.go
GokceGK: Removed the dots from the sentence endings

internal/cmd/beta/network/create/create.go
GokceGK: edited description

internal/cmd/beta/network/create/create.go
```diff
@@ -50,12 +64,20 @@ func NewCmd(p *print.Printer) *cobra.Command {
...
+				`$ stackit beta network create --name network-1 --no-ipv4-gateway`,
+			),
+			examples.NewExample(
+				`Create an IPv4 network with name "network-1" with DNS name servers, a prefix, a gateway and a prefix length`,
+				`$ stackit beta network create --name network-1  --ipv4-dns-name-servers "1.1.1.1,8.8.8.8,9.9.9.9" --ipv4-prefix-length 25 --ipv4-prefix "10.1.2.0/24" --ipv4-gateway "10.1.2.3"`,
+			),
+			examples.NewExample(
+				`Create an IPv6 network with name "network-1" with DNS name servers, a prefix, a gateway and a prefix length`,
+				`$ stackit beta network create --name network-1  --ipv6-dns-name-servers "2001:4860:4860::8888,2001:4860:4860::8844" --ipv6-prefix-length 56 --ipv6-prefix "2001:4860:4860::8888" --ipv6-gateway "2001:4860:4860::8888"`,
 			),
```
joaopalet, line 81: For some Terraform provider example, the IaaS colleagues said that providing a prefix and a prefix length is redundant, and the prefi length will be ignored if prefix is set. So we should adjust the examples also here

internal/cmd/beta/network/list/list.go
joaopalet: This same handling could also be applied to the output of the `describe`, right?

internal/cmd/beta/network/create/create.go
GokceGK: updated the examples

internal/cmd/beta/network/list/list.go
GokceGK: I adapted the routed row handling

internal/cmd/beta/network/create/create_test.go
```diff
@@ -155,27 +171,69 @@ func TestParseInput(t *testing.T) {
...
+		},
+		{
+			description: "use routed true",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[nonRoutedFlag] = "false"
+			}),
+			isValid: true,
+			expectedModel: fixtureInputModel(func(model *inputModel) {
+				model.NonRouted = false
 			}),
```
joaopalet, line 237: This is aleady the default, right?

joaopalet:
```suggestion
			description: "non-routed network",
			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
				flagValues[nonRoutedFlag] = "true"
			}),
			isValid: true,
			expectedModel: fixtureInputModel(func(model *inputModel) {
				model.NonRouted = true
			}),
```

internal/cmd/beta/network/create/create_test.go
joaopalet, line 1: Missing some more unit tests for the `buildRequest` function

internal/cmd/beta/network/describe/describe.go
```diff
@@ -152,6 +152,19 @@ func outputResult(p *print.Printer, outputFormat string, network *iaas.Network)
 			table.AddSeparator()
 		}
 
+		routed := true
+		if network.Routed != nil {
+			routed = *network.Routed
+		}
```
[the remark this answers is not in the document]
GokceGK: normally it should not be possible to have routed field empty. But somehow I managed to create a broken network in my project and in there routed field did not have any value and cli had NPE in list and describe commands. That is why I added this control.

internal/cmd/beta/network/create/create_test.go
GokceGK: added some more unit tests

internal/cmd/beta/network/describe/describe.go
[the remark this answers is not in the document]
joaopalet: Alright. But then the question is just if for some reason the `routed` field is not present in the response, I think we should either show `false` or not show the value in the output. Showing `true` might be misleading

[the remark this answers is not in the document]
GokceGK: you are right, I set them back to false. That was also the initial idea of implementation

vicentepinto98 (approved):

PR #513, opened 2024-11-27
Onboard IaaS Public IP commands

GokceGK:
stackit beta public-ip

create
delete
update
describe
list
associate
disassociate

PR #518, opened 2024-11-29
Add commands to attach/detach public IP from servers

joaopalet:
- Add commands to attach/detach public IP from servers
  - `$ stackit beta server public-ip attach/detach`

Also:
- Adjusts table output of `server describe` command to show separate tables when `--details` is passed
- Adjusts some command group descriptions

### DIFF
--- a/internal/cmd/beta/server/describe/describe.go
+++ b/internal/cmd/beta/server/describe/describe.go
@@ -224,7 +224,7 @@ func outputResult(p *print.Printer, model *inputModel, server *iaas.Server) erro
 					nicsTable.AddRow("PUBLIC IP", *nic.PublicIp)
 				}
 
-				nicsTable.Display(p)
+				err := nicsTable.Display(p)
 				if err != nil {
 					return fmt.Errorf("render table: %w", err)
 				}
--- a/internal/cmd/beta/server/describe/describe.go
+++ b/internal/cmd/beta/server/describe/describe.go
@@ -11,7 +11,6 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/examples"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/flags"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/services/iaas/client"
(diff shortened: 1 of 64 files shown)


internal/pkg/services/iaas/utils/utils.go
```diff
@@ -16,6 +17,18 @@ type IaaSClient interface {
 	GetNetworkAreaRangeExecute(ctx context.Context, organizationId, areaId, networkRangeId string) (*iaas.NetworkRange, error)
 }
 
+func GetPublicIP(ctx context.Context, apiClient IaaSClient, projectId, publicIpId string) (ip, associatedResource string, err error) {
```
joaopalet, line 20: The second return argument is not used in my PR, but I took this util function from [this open PR](https://github.com/stackitcloud/stackit-cli/pull/513/files#diff-4667d7303e488ebb891426ede3ae2da4ae33eab7629ca8e7145f37f4596598b8) from 
@GokceGK where it is needed

internal/cmd/beta/server/describe/describe.go
```diff
@@ -184,35 +187,50 @@ func outputResult(p *print.Printer, model *inputModel, server *iaas.Server) erro
...
+		if err != nil {
+			return fmt.Errorf("render table: %w", err)
+		}
+
+		// If no --details is set, we only display the overview table
+		if !model.Details {
+			return nil
+		}
+
+		if server.Nics != nil && len(*server.Nics) > 0 {
```
vicentepinto98, line 212: why not have just one table with the list of Nics? With a column for each field and a row for each nic

internal/cmd/beta/server/describe/describe.go
```diff
@@ -184,35 +187,50 @@ func outputResult(p *print.Printer, model *inputModel, server *iaas.Server) erro
...
-				table.AddRow("SERVICE ACCOUNTS", strings.Join(emails, "\n"))
-				table.AddSeparator()
-			}
+		err := table.Display(p)
+		if err != nil {
+			return fmt.Errorf("render table: %w", err)
+		}
+
+		// If no --details is set, we only display the overview table
+		if !model.Details {
```
vicentepinto98, line 208: aren't there more tables to add besides the attached nics, like the public IP for example? 
If not, I'm not sure why we need to add this "--details" flag instead of just adding the nic row to the existing table

internal/cmd/beta/server/describe/describe.go
joaopalet: Good point, looks much better yes. Since the overview is also a vertical table I had added it the same way

internal/cmd/beta/server/describe/describe.go
joaopalet: Ah, ok I agree. Because if they don't know the existence of the flag would never see the NICs (when using `pretty` output)

joaopalet: WIll remove it then

vicentepinto98 (approved):

PR #532, opened 2024-12-09
Onboard keypair commands

marceljk:
Onboard key-pair commands:

- create
- delete
- describe
- list
- update


internal/cmd/beta/key-pair/list/list.go
```diff
@@ -0,0 +1,171 @@
...
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	Limit         *int64
+	LabelSelector *string
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "Lists all SSH Keypairs",
```
vicentepinto98, line 37:
```suggestion
		Short: "Lists all key pairs",
```

Same thing here, we should be uniform on the command descriptions.
We can go for `SSH Key Pairs` instead, but then it should be the case for every command

internal/cmd/beta/key-pair/describe/describe.go
```diff
@@ -0,0 +1,163 @@
...
+
+			return outputResult(p, model.OutputFormat, model.PublicKey, resp)
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Bool(publicKeyFlag, false, "Show only the public key")
```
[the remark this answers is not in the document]
vicentepinto98: In that case we wouldn't truncate it, does it look good in the table or it doesn't fit nicely?

internal/cmd/beta/key-pair/update/update.go
```diff
@@ -0,0 +1,143 @@
...
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update key pair",
+		Long:  "Update key pair.",
+		Args:  args.SingleArg(keypairNameArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Update the labels of a keypair KEYPAIR_NAME with "key=value,key1=value1"`,
+				"$ stackit beta key-pair update --labels key=value,key1=value1",
```
vicentepinto98, line 41:
```suggestion
				"$ stackit beta key-pair update KEYPAIR_NAME --labels key=value,key1=value1",
```

internal/cmd/beta/key-pair/describe/describe.go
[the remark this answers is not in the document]
vicentepinto98: Alright I see your point. 
Then I suggest adjusting also the JSON and YAML output to only show that the public key when that flag is passed

internal/cmd/beta/key-pair/update/update.go
```diff
@@ -0,0 +1,143 @@
...
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update Key Pair",
+		Long:  "Update Key Pair.",
+		Args:  args.SingleArg(keyPairNameArg, nil),
+		Example: examples.Build(
+			examples.NewExample(
+				`Update the labels of a Key Pair KEY_PAIR_NAME with "key=value,key1=value1"`,
```
vicentepinto98, line 40:
```suggestion
				`Update the labels of a Key Pair with name "KEY_PAIR_NAME" with "key=value,key1=value1"`,
```

internal/cmd/beta/key-pair/create/create.go
```diff
@@ -0,0 +1,162 @@
...
+	*globalflags.GlobalFlagModel
+	Name      *string
+	PublicKey *string
+	Labels    *map[string]string
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a Key Pair",
```
vicentepinto98, line 36:
```suggestion
		Short: "Create a key pair",
```

I would use lower case everywhere (error messages, examples, command descriptions). 
In the CLI we only use capital letter for names of services such as `PostgreSQL` or `DNS`, not for the name of the resources

vicentepinto98:
We should add the new command group in the `Available Services` section in the `README`, in the `IaaS` row

internal/cmd/beta/key-pair/create/create.go
vicentepinto98, line 36:
```suggestion
		Short: "Creates a Key Pair",
```

All command descriptions, both sort and long, should be in the present tense, for uniformity in the whole CLI

internal/cmd/beta/key-pair/create/create.go
```diff
@@ -0,0 +1,162 @@
...
+
+			return outputResult(p, model, resp)
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().String(nameFlag, "", "key pair name")
```
vicentepinto98, line 93:
```suggestion
	cmd.Flags().String(nameFlag, "", "Key pair name")
```

vicentepinto98 (approved):

PR #536, opened 2024-12-12
Onboard IaaS server action commands

Fyusel:
Supported commands: start, stop, reboot, deallocate, resize, console, log, rescue, unrescue

### DIFF
--- a/internal/cmd/beta/server/console/console.go
+++ b/internal/cmd/beta/server/console/console.go
@@ -1,0 +1,140 @@
+package console
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/url"
+
+	"github.com/goccy/go-yaml"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/examples"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/services/iaas/client"
+	iaasUtils "github.com/stackitcloud/stackit-cli/internal/pkg/services/iaas/utils"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	serverIdArg = "SERVER_ID"
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	ServerId string
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "console",
+		Short: "Gets a URL for server remote console",
+		Long:  "Gets a URL for server remote console.",
+		Args:  args.SingleArg(serverIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(
+				`Get a URL for the server remote console with server ID "xxx"`,
+				"$ stackit beta server console xxx",
+			),
+			examples.NewExample(
+				`Get a URL for the server remote console with server ID "xxx" in JSON format`,
+				"$ stackit beta server console xxx --output-format json",
+			),
+		),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			model, err := parseInput(p, cmd, args)
+			if err != nil {
+				return err
+			}
+
+			// Configure API client
+			apiClient, err := client.ConfigureClient(p)
+			if err != nil {
+				return err
+			}
+
+			serverLabel, err := iaasUtils.GetServerName(ctx, apiClient, model.ProjectId, model.ServerId)
+			if err != nil {
+				p.Debug(print.ErrorLevel, "get server name: %v", err)
+				serverLabel = model.ServerId
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("server console: %w", err)
+			}
+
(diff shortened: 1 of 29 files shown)


internal/pkg/flags/flag_to_value.go
```diff
@@ -159,6 +159,17 @@ func FlagWithDefaultToStringValue(p *print.Printer, cmd *cobra.Command, flag str
...
+// Returns the bool value set on the flag. If no value is set, returns the flag's default value.
+// Returns false if the flag value can not be converted to bool or if the flag does not exist.
+func FlagWithDefaultToBoolValue(p *print.Printer, cmd *cobra.Command, flag string) bool {
+	value, err := cmd.Flags().GetBool(flag)
+	if err != nil {
+		p.Debug(print.ErrorLevel, "convert flag with default to bool value: %v", err)
+		return false
+	}
+	return value
+}
```
joaopalet, line 171: This function is exactly the same as `FlagToBoolValue`, right?

joaopalet: I don't think we need it, also because bool flags always have a default (they are always either `true` or `false`)

internal/cmd/beta/server/start/start.go
```diff
@@ -0,0 +1,115 @@
...
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	ServerId string
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "start",
+		Short: "Start an existing server or allocates the server if deallocated",
+		Long:  "Start an existing server or allocates the server if deallocated.",
```
joaopalet, line 35:
```suggestion
		Short: "Starts an existing server or allocates the server if deallocated",
		Long:  "Starts an existing server or allocates the server if deallocated.",
```

joaopalet: Same for all other commands in the PR (to match the style of all descriptions in the CLI)

internal/cmd/beta/server/start/start.go
```diff
@@ -0,0 +1,115 @@
...
+				s := spinner.New(p)
+				s.Start("Starting server")
+				_, err = wait.StartServerWaitHandler(ctx, apiClient, model.ProjectId, model.ServerId).WaitWithContext(ctx)
+				if err != nil {
+					return fmt.Errorf("wait for server starting: %w", err)
+				}
+				s.Stop()
+			}
+
+			p.Info("Server %q started!\n", serverLabel)
```
joaopalet, line 80: Sincr the operation is not synchronous, we should only output `Started ...` if we have called the waiter. Otherwise it should say `Triggered start of ...` (or something similar). You can check e.g. https://github.com/stackitcloud/stackit-cli/blob/9324319c69c607028f0e2e3381348e0b861898f4/internal/cmd/postgresflex/instance/create/create.go#L295

joaopalet: Also not sure about the `!` (just becasue I don't think it's present in other commands)

joaopalet: Samr for other commands in the PR

internal/cmd/beta/server/reboot/reboot.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("server reboot: %w", err)
+			}
+
+			p.Info("Server %q rebooted!\n", serverLabel)
```
joaopalet, line 85: Is this operation synchronous?

internal/cmd/beta/server/log/log.go
```diff
@@ -0,0 +1,138 @@
...
+			log := *resp.GetOutput()
+			lines := strings.Split(log, "\n")
+			p.Info("lines = %d", len(lines))
+
+			if len(lines) > int(*model.Length) {
+				// Truncate output and show most recent logs
+				start := len(lines) - int(*model.Length)
+				p.Info("Log for server %q\n%s", serverLabel, strings.Join(lines[start:], "\n"))
+				return nil
+			}
```
joaopalet, line 86: This output should be handled in an `outputResult` function (similar to other commands), to account for the different output formats: JSON, YAML, or "pretty"

joaopalet: In "pretty" we are free to display however we want, does not necessarily need to be a table

internal/cmd/beta/server/console/console.go
```diff
@@ -0,0 +1,109 @@
...
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("server console: %w", err)
+			}
+
+			// unescape url in order to get rid of e.g. %40
+			unescapedURL, err := url.PathUnescape(*resp.GetUrl())
+			if err != nil {
+				return fmt.Errorf("unescape url: %w", err)
+			}
```
joaopalet, line 72: Same regarding the supporting different output formats

internal/cmd/beta/server/reboot/reboot.go
Fyusel: At least there is no wait function for reboot. Needs to be extended in the sdk first.

internal/cmd/beta/server/start/start.go
```diff
@@ -0,0 +1,119 @@
...
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "start",
+		Short: "Starts an existing server or allocates the server if deallocated",
+		Long:  "Starts an existing server or allocates the server if deallocated.",
+		Args:  args.SingleArg(serverIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(
+				`Starts an existing server with ID "xxx"`,
```
joaopalet, line 39: Nitpick: for the examples, they are always in the form of "Start" (instead of "Starts")

joaopalet: Meaning the description of the command should be "Starts ...", while each exemple would be in the form of "Start ..."

vicentepinto98 (approved):

PR #539, opened 2024-12-16
Add command to import config profiles

marceljk:
- Add functionality to import config profiles
- Add tests


internal/cmd/config/profile/import/import.go
```diff
@@ -0,0 +1,106 @@
...
+		Use:   "import",
+		Short: "Imports a CLI configuration profile",
+		Long:  "Imports a CLI configuration profile.",
+		Example: examples.Build(
+			examples.NewExample(
+				`Import a config with name "PROFILE_NAME" from file "./config.json"`,
+				"$ stackit config profile --name PROFILE_NAME --config `@./config.json`",
+			),
+			examples.NewExample(
+				`Import a config with name "PROFILE_NAME" from file "./config.json" and set not as active`,
```
GokceGK, line 38: 'set not as active' sounds not so understandable

maybe we can say `set as inactive` or 'do not set as active'

internal/cmd/config/profile/import/import.go
```diff
@@ -0,0 +1,106 @@
...
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "import",
+		Short: "Imports a CLI configuration profile",
+		Long:  "Imports a CLI configuration profile.",
+		Example: examples.Build(
+			examples.NewExample(
+				`Import a config with name "PROFILE_NAME" from file "./config.json"`,
+				"$ stackit config profile --name PROFILE_NAME --config `@./config.json`",
```
GokceGK, line 35: command is missing `stackit config profile import`

internal/cmd/config/profile/import/import.go
```diff
@@ -0,0 +1,106 @@
...
+			),
+			examples.NewExample(
+				`Import a config with name "PROFILE_NAME" from file "./config.json" and set not as active`,
+				"$ stackit config profile --name PROFILE_NAME --config `@./config.json` --no-set",
```
GokceGK, line 39: here also command missing

internal/cmd/config/profile/import/import.go
```diff
@@ -0,0 +1,106 @@
+package _import
```
GokceGK, line 1: generally we do not have underscore _ before package names

marceljk: i used there an underscore before the package name because import is an keyword of golang and can not be used as package name. Renamed it to importProfile

GokceGK (approved):

PR #544, opened 2024-12-18
Onboard command to export config profiles

marceljk:


internal/pkg/config/profiles_test.go
```diff
@@ -182,3 +183,87 @@ func TestImportProfile(t *testing.T) {
...
+	})
+
+	// Create prerequisite profile
+	p := print.NewPrinter()
+	profileName := "export-profile-test"
+	err = CreateProfile(p, profileName, true, true)
+	if err != nil {
+		t.Fatalf("could not create prerequisite profile, %v", err)
+	}
+	InitConfig()
```
joaopalet, line 209: This would make the tests run in the real config directory, right? Possible messing with existing profiles

joaopalet: Maybe we could define a `InitTestConfig` that would use some directory inside this package to simulate the real one

internal/pkg/config/profiles_test.go
```diff
@@ -182,3 +183,87 @@ func TestImportProfile(t *testing.T) {
...
+		}(testDir)
+	})
+
+	// Create prerequisite profile
+	p := print.NewPrinter()
+	profileName := "export-profile-test"
+	err = CreateProfile(p, profileName, true, true)
+	if err != nil {
+		t.Fatalf("could not create prerequisite profile, %v", err)
+	}
```
joaopalet, line 208: This would allow us even to have more control and test the case where the profile doesn't exist

internal/pkg/config/profiles.go
```diff
@@ -397,3 +398,45 @@ func ImportProfile(p *print.Printer, profileName, config string, setAsActive boo
...
+
+	exportFileName := fmt.Sprintf("%s.%s", profile, configFileExtension)
+	exportFilePath := filePath
+	if !strings.HasSuffix(exportFilePath, fmt.Sprintf(".%s", configFileExtension)) {
+		exportFilePath = filepath.Join(filePath, exportFileName)
+	}
+
+	_, err = os.Stat(exportFilePath)
+	if err == nil {
+		return fmt.Errorf("file %q already exists in the export path. Delete the existing file or define a different export path", exportFileName)
```
joaopalet, line 429: I think you meant to use `errors.FILE_ALREADY_EXISTS` here, right?

internal/cmd/config/profile/export/export.go
```diff
@@ -0,0 +1,98 @@
...
+
+			return nil
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().StringP(filePathFlag, "f", "", "If set, writes the payload to the given. If unset, writes the payload to you current directory with the name of the profile. E.g. '--file-path ~/my-config.json', '--file-path ~/'")
```
joaopalet, line 69:
```suggestion
	cmd.Flags().StringP(filePathFlag, "f", "", "If set, writes the config to the given. If unset, writes the config to you current directory with the name of the profile. E.g. '--file-path ~/my-config.json', '--file-path ~/'")
```

internal/pkg/config/profiles_test.go
```diff
@@ -182,3 +183,87 @@ func TestImportProfile(t *testing.T) {
...
+
+	// Create prerequisite profile
+	p := print.NewPrinter()
+	profileName := "export-profile-test"
+	err = CreateProfile(p, profileName, true, true)
+	if err != nil {
+		t.Fatalf("could not create prerequisite profile, %v", err)
+	}
+	initConfig(defaultConfigFolderPath)
```
joaopalet, line 215:
```suggestion
	testConfigFolderPath = filepath.Join(testDir, "config")
	err = os.Mkdir(defaultConfigFolderPath, 0o750)
	if err != nil {
		t.Fatal(err)
	}

	// Create prerequisite profile
	p := print.NewPrinter()
	profileName := "export-profile-test"
	err = CreateProfile(p, profileName, true, true)
	if err != nil {
		t.Fatalf("could not create prerequisite profile, %v", err)
	}
	initConfig(testConfigFolderPath)
```

internal/cmd/config/profile/export/export.go
```diff
@@ -0,0 +1,98 @@
...
+		GlobalFlagModel: globalFlags,
+		ProfileName:     profileName,
+		FilePath:        flags.FlagToStringValue(p, cmd, filePathFlag),
+	}
+
+	// If filePath contains does not contain a file name, then add a default name
+	if !strings.HasSuffix(model.FilePath, fmt.Sprintf(".%s", configFileExtension)) {
+		exportFileName := fmt.Sprintf("%s.%s", model.ProfileName, configFileExtension)
+		model.FilePath = filepath.Join(model.FilePath, exportFileName)
+	}
```
joaopalet, line 86: Hm, so if a user writes `--file-path ./config` the file will be written to `./config/PROFILE_NAME.json`?

joaopalet: What if e.g. they write ` --file-path ./config.txt`? This would try to create a file in `./config.txt/PROFILE_NAME.json`, right?

joaopalet: Maybe we don't need to force it to be a `.json` file, I think this is the most common approach

internal/pkg/config/profiles_test.go
marceljk: I set here the global variable `globalConfigFolderPath` from `config.go. This is used in the `CreateProfile()`function, to get the profile folder path by calling `GetProfileFolderPath()`

joaopalet: Don't you pass it to the `initConfig` function as argument? And only there the global variable is initialized, right?

I think currently you overwrite the global `defaultConfigFolderPath` in the test, then call `initConfig` with the new value (where it's overwritten again)

marceljk: Yes but I need to set the `defaultConfigFolderPath` before the call of the `CreateProfile()`. Because otherwise, it would create a profile folder in the normal path and not in the temp directory which I create for the test. And after the call of the `CreateProfile()`, i need to call the `initConfig()` and `Write()`. This will create the `cli-config.json` in the profile folder.

To avoid the duplicated set of `defaultConfigFolderPath`, I could remove the initialization of it in the `initConfig()` and add it to the `InitConfig()` before it calls `initConfig()`

marceljk: So this would avoid the duplicate initialization
```
func InitConfig() {
	defaultConfigFolderPath = getInitialConfigDir()
	initConfig()
}

func initConfig() {
	profileFilePath = getInitialProfileFilePath() // Profile file path is in the default config folder
        ...
}

```

joaopalet (dismissed):

PR #550, opened 2025-01-08
Onboard "region" global flag for region-specific requests

marceljk:
Add global flag "region" to set the region for region-specific requests

### DIFF
--- a/docs/stackit.md
+++ b/docs/stackit.md
@@ -20,6 +20,7 @@ stackit [flags]
   -h, --help                   Help for "stackit"
   -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
   -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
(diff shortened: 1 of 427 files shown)


bahkauv70 (commented):
The attribute defaultNodepoolAvailabilityZone in pkg/services/ske/utils/utils.go is currently set to "eu01-3". Must this be changed to the global region flag as well?

marceljk:
> > The attribute defaultNodepoolAvailabilityZone in pkg/services/ske/utils/utils.go is currently set to "eu01-3". Must this be changed to the global region flag as well?
> 
> Availability zone and region are different concepts as far as I know, I wouldn't change this one

I have checked its usage. It is used by the `ske cluster create` command when a user does not provide a payload to it. I'm not sure if this might cause errors in the future if a user sets the region to _eu02_ and doesn't provide a payload for the ske cluster.

vicentepinto98:
That's a good point, it could create issues. When SKE is available in the new region, we should check with them what should be the default value for the availability zone in that case and have some default mapping

vicentepinto98 (approved):

PR #559, opened 2025-01-14
Support for IaaS images

bahkauv70:


internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates images",
+		Long:  "Creates images.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(`Create a named imaged`, `$ stackit beta image create --name my-new-image --disk-format=raw --local-file-path=/my/raw/image`),
```
marceljk, line 88: Describe the examples a bit more like we do in other commands. [Like here](https://github.com/stackitcloud/stackit-cli/blob/5d5ae24841bcdd01c065a7d8519c61d3725e59a9/internal/cmd/beta/network-area/create/create.go#L54)

internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	configureFlags(cmd)
+	return cmd
+}
+
+func uploadFile(ctx context.Context, p *print.Printer, file *os.File, url string) (err error) {
+	var filesize int64
+	if stat, err := file.Stat(); err != nil {
+		p.Debug(print.DebugLevel, "create image: cannot open file %q: %w", file.Name(), err)
```
marceljk, line 154: When the image cannot be opened here, then it should return it here

internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,329 @@
...
+	uploadRequest.Header.Add("Content-Type", "application/octet-stream")
+	uploadRequest.ContentLength = filesize
+
+	uploadResponse, err := http.DefaultClient.Do(uploadRequest)
+	if err != nil {
+		return fmt.Errorf("create image: error contacting server for upload: %w", err)
+	}
+	defer func() {
+		if inner := uploadResponse.Body.Close(); inner != nil {
+			err = fmt.Errorf("error closing file: %wqq (%w)", inner, err)
```
marceljk, line 176: %wqq can be replaced with %w

internal/cmd/beta/image/create/create.go
marceljk, line 1: An upload can take some time depending on the internet connection of the user and the image size. I think it would be nice to add something like a spinner or progress bar, so that the user knows that it's still running.

internal/cmd/beta/image/list/list.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/projectname"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/services/iaas/client"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/tables"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	LabelSelector *string
```
marceljk, line 25: We have in other commands also a limit flag to show only a limited amount of entries. This could here also be added. For example https://github.com/stackitcloud/stackit-cli/blob/5d5ae24841bcdd01c065a7d8519c61d3725e59a9/internal/cmd/beta/network-area/list/list.go#L25

internal/cmd/beta/image/update/update.go
```diff
@@ -0,0 +1,283 @@
...
+)
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("update %s", imageIdArg),
+		Short: "Updates an image",
+		Long:  "Updates an image",
+		Args:  args.SingleArg(imageIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(`Update the name of image "xxx"`, `$ stackit beta image update xxx --name my-new-name`),
```
marceljk, line 114:
```suggestion
			examples.NewExample(`Update the name of an image with ID "xxx"`, `$ stackit beta image update xxx --name my-new-name`),
```

internal/cmd/beta/image/describe/describe.go
```diff
@@ -0,0 +1,173 @@
...
+const imageIdArg = "IMAGE_ID"
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("describe %s", imageIdArg),
+		Short: "Describes image",
+		Long:  "Describes an image by its internal ID.",
+		Args:  args.SingleArg(imageIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(`Describe image "xxx"`, `$ stackit beta image describe xxx`),
```
marceljk, line 36:
```suggestion
			examples.NewExample(`Describe image with ID "xxx"`, `$ stackit beta image describe xxx`),
```

internal/cmd/beta/image/create/create.go
bahkauv70: Extended the example

internal/cmd/beta/image/create/create.go
bahkauv70: Upps, typo

internal/cmd/beta/image/list/list.go
bahkauv70: Added support for limit flag

internal/cmd/beta/image/create/create.go
bahkauv70: Added a simple progress timer that ticks every 5s to indicate activity. A progress meter showing the percentage uploaded would be possible as well, but requires more effort (continously measure the contents read from the file), which imho is not warranted in ths use case

internal/cmd/beta/image/create/create.go
bahkauv70: The image was already openend at this point. The file is actually opened before any API calls are made to allow for early failure without any changes to the server side. So, normally this shouldn't actually fail anymore, when we got to this point. However to be sure the error is now returned and the function is aborted

internal/cmd/beta/image/create/create.go
bahkauv70: Ok, thought about this one again: A "real" progress indicator would be more useful, so I've added a simple implementation that will show every 2s the percentage already uploaded

internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,401 @@
...
+		Long:  "Creates images.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a named image 'my-new-image' from a raw disk image located in '/my/raw/image'`,
+				`$ stackit beta image create --name my-new-image --disk-format=raw --local-file-path=/my/raw/image`,
+			),
+			examples.NewExample(
+				`Create a named image 'my-new-image' from a qcow2 image read from '/my/qcow2/image' with labels describing its contents`,
+				`$ stackit beta image create --name my-new-image --disk-format=qcow2 --local-file-path=/my/qcow2/image--labels os=linux,distro=alpine,version=3.12`,
```
marceljk, line 98: typo
```suggestion
				`$ stackit beta image create --name my-new-image --disk-format=qcow2 --local-file-path=/my/qcow2/image --labels os=linux,distro=alpine,version=3.12`,
```

internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,401 @@
...
+	secureBootFlag             = "secure-boot"
+	uefiFlag                   = "uefi"
+	videoModelFlag             = "video-model"
+	virtioScsiFlag             = "virtio-scsi"
+
+	labelsFlag = "labels"
+
+	minDiskSizeFlag = "min-disk-size"
+	minRamFlag      = "min-ram"
+	ownerFlag       = "owner"
```
marceljk, line 51: It seems like the ``ownerFlag`` isn't used anywhere. Can it be removed?

internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,401 @@
...
+		Short: "Creates images",
+		Long:  "Creates images.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a named image 'my-new-image' from a raw disk image located in '/my/raw/image'`,
+				`$ stackit beta image create --name my-new-image --disk-format=raw --local-file-path=/my/raw/image`,
+			),
+			examples.NewExample(
+				`Create a named image 'my-new-image' from a qcow2 image read from '/my/qcow2/image' with labels describing its contents`,
```
marceljk, line 97:
```suggestion
				`Create a image with name 'my-new-image' from a qcow2 image read from '/my/qcow2/image' with labels describing its contents`,
```

internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,401 @@
...
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates images",
+		Long:  "Creates images.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a named image 'my-new-image' from a raw disk image located in '/my/raw/image'`,
```
marceljk, line 93:
```suggestion
				`Create a image with name 'my-new-image' from a raw disk image located in '/my/raw/image'`,
```

internal/cmd/beta/image/create/create.go
```diff
@@ -0,0 +1,401 @@
...
+		var ch <-chan int
+		reader, ch = newProgressReader(file)
+		go func() {
+			ticker := time.NewTicker(2 * time.Second)
+			var uploaded int
+			done:
+			for {
+				select {
+				case <-ticker.C:
+					p.Info("uploaded %3.1f%%\n", 100.0/float64(stat.Size())*float64(uploaded))
```
marceljk, line 180: You can replace the `\n` with `\r`, then it doesn't print always a new line. Instead it would replace the previous output
```suggestion
					p.Info("uploaded %3.1f%%\r", 100.0/float64(stat.Size())*float64(uploaded))
```

internal/cmd/beta/image/update/update.go
```diff
@@ -0,0 +1,283 @@
...
+
+func NewCmd(p *print.Printer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("update %s", imageIdArg),
+		Short: "Updates an image",
+		Long:  "Updates an image",
+		Args:  args.SingleArg(imageIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(`Update the name of an image with ID "xxx"`, `$ stackit beta image update xxx --name my-new-name`),
+			examples.NewExample(`Update the labels of image "xxx"`, `$ stackit beta image update xxx --labels label1=value1,label2=value2`),
```
marceljk, line 115:
```suggestion
			examples.NewExample(`Update the labels of an image with ID "xxx"`, `$ stackit beta image update xxx --labels label1=value1,label2=value2`),
```

internal/cmd/beta/image/create/create.go
bahkauv70: removed

internal/cmd/beta/image/delete/delete.go
```diff
@@ -0,0 +1,110 @@
...
+
+			projectLabel, err := projectname.GetProjectName(ctx, p, cmd)
+			if err != nil {
+				p.Debug(print.ErrorLevel, "get project name: %v", err)
+				projectLabel = model.ProjectId
+			}
+
+			imageName, err := iaasUtils.GetImageName(ctx, apiClient, model.ProjectId, model.ImageId)
+			if err != nil {
+				p.Warn("get image name: %v", err)
```
marceljk, line 57:
```suggestion
				p.Debug(print.ErrorLevel, "get image name: %v", err)
```

internal/cmd/beta/image/create/create.go
bahkauv70: fixed

internal/cmd/beta/image/create/create.go
bahkauv70: Yes, removed

internal/cmd/beta/image/create/create.go
bahkauv70: done

internal/cmd/beta/image/delete/delete.go
bahkauv70: Is there a semantic difference between these variants or is this just the canonical style? I did use it in other places as well, so those would have to be changed as well

bahkauv70: What about print.Info? Should this be changed as well to print.Debug(print.InfoLevel) ?

internal/cmd/beta/image/update/update.go
bahkauv70: done

internal/cmd/beta/image/create/create.go
bahkauv70: done

internal/cmd/beta/image/delete/delete.go
marceljk: The p.Warn will always be shown to the user and the p.Debug only if the verbosity is set to debug. So it always depends if the user should see the message or if it's more a debug print

joaopalet: Yes, with the default `--versbosity` (`info`), `p.Warn` is shown

`p.Debug` logs are only shown if the user sets the `--verbosity` to `debug`

joaopalet: More info on that [here](https://github.com/stackitcloud/stackit-cli/blob/main/CONTRIBUTION.md#outputs-prints-and-debug-logs)

bahkauv70: Ok, so we should using print.Debug(print.ErrorLevel,...) when it's an internal and already handled error. It should not concern the user, if he did not opt for verbose mode?

marceljk (approved):

PR #561, opened 2025-01-15
ref 700848 - server os-update

aeter:
adds `stackit beta server os-update ...`  ( Server Update Management API )

Testing Done:
tested locally. ran linters/formatters/tests/doc generators

### DIFF
--- a/internal/cmd/beta/server/os-update/schedule/create/create.go
+++ b/internal/cmd/beta/server/os-update/schedule/create/create.go
@@ -1,0 +1,167 @@
+package create
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/goccy/go-yaml"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
+	cliErr "github.com/stackitcloud/stackit-cli/internal/pkg/errors"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/examples"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/flags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/services/serverosupdate/client"
+
+	"github.com/spf13/cobra"
+	"github.com/stackitcloud/stackit-sdk-go/services/serverupdate"
+)
+
+const (
+	nameFlag              = "name"
+	enabledFlag           = "enabled"
+	rruleFlag             = "rrule"
+	maintenanceWindowFlag = "maintenance-window"
+	serverIdFlag          = "server-id"
+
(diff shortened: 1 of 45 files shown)


internal/cmd/beta/server/os-update/schedule/create/create.go
```diff
@@ -0,0 +1,167 @@
...
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates a Server os-update Schedule",
+		Long:  "Creates a Server os-update Schedule.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create a Server os-update Schedule with name "myschedule"`,
+				`$ stackit beta server os-update schedule create --server-id xxx --name=myschedule`),
+			examples.NewExample(
```
bahkauv70, line 53: An example using a non-trivial recurrence rule would be useful here

internal/cmd/beta/server/os-update/schedule/create/create_test.go
```diff
@@ -0,0 +1,212 @@
...
+	model := &inputModel{
+		GlobalFlagModel: &globalflags.GlobalFlagModel{
+			ProjectId: testProjectId,
+			Verbosity: globalflags.VerbosityDefault,
+		},
+		ServerId:          testServerId,
+		ScheduleName:      "example-schedule-name",
+		Enabled:           defaultEnabled,
+		Rrule:             defaultRrule,
+		MaintenanceWindow: int64(23),
```
bahkauv70, line 52: repeating magic value, a constant would be in order

internal/cmd/beta/server/os-update/schedule/describe/describe.go
```diff
@@ -0,0 +1,149 @@
...
+		table := tables.NewTable()
+		table.AddRow("SCHEDULE ID", *schedule.Id)
+		table.AddSeparator()
+		table.AddRow("SCHEDULE NAME", *schedule.Name)
+		table.AddSeparator()
+		table.AddRow("ENABLED", *schedule.Enabled)
+		table.AddSeparator()
+		table.AddRow("RRULE", *schedule.Rrule)
+		table.AddSeparator()
+		table.AddRow("MAINTENANCE WINDOW", *schedule.MaintenanceWindow)
```
bahkauv70, line 139: the os-update describe command used the place holder "..." for empty values, which is not handled here. Or can those entries never be nil?

bahkauv70 (commented):
commented on some minor formalities

internal/cmd/beta/server/os-update/schedule/describe/describe.go
aeter: This is for the schedule (`os-update schedule describe`)- here they cannot be `nil`. I've handled them as `nil` at the `internal/cmd/beta/server/os-update/describe/describe.go`, for the `os-update describe` command.

Usually, the `os-update create` may take a while in `running` state, thus the need for `nil` handling there (but here, the schedule is created instantly).

bahkauv70 (approved):

bahkauv70 (approved):

PR #569, opened 2025-01-21
Extend output of IaaS commands

marceljk:
- `beta network list`
  - add value: prefixes
- `beta security-group describe`
  - add values
    - stateful
    - createdAt
    - updatedAt
  - add "Rules" table with values:
    - ID
    - description
    - protocol
    - direction
    - etherType
    - portRange
    - ipRange
    - icmpParameters
    - remoteSecurityGroupId
- `beta security-group list`
  - add values:
    - description
    - labels
- `beta security-group rule list`
  - add value: remoteSecurityGroupId
- `beta server describe`
  - add "Maintenance Window" table with values:
    - status
    - details
    - startsAt
    - endsAt
- `beta server list`
  - add values:
    - machineType
    - nicIPv4
    - publicIPs

### DIFF
--- a/internal/cmd/beta/server/list/list.go
+++ b/internal/cmd/beta/server/list/list.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
 
 	"github.com/goccy/go-yaml"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
@@ -140,6 +141,7 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *iaas.APICli
 	if model.LabelSelector != nil {
 		req = req.LabelSelector(*model.LabelSelector)
 	}
+	req = req.Details(true)
 
 	return req
 }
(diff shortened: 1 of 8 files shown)


internal/cmd/beta/server/list/list.go
```diff
@@ -164,11 +166,36 @@ func outputResult(p *print.Printer, outputFormat string, servers []iaas.Server)
...
 			server := servers[i]
-			table.AddRow(*server.Id, *server.Name, *server.Status, *server.AvailabilityZone)
+
+			nicIPv4 := ""
+			publicIPs := ""
+			if server.Nics != nil && len(*server.Nics) > 0 {
+				for i, nic := range *server.Nics {
+					if nic.Ipv4 != nil || nic.PublicIp != nil {
+						nicIPv4 += utils.PtrString(nic.Ipv4)
+						publicIPs = utils.PtrString(nic.PublicIp)
```
Fyusel, line 180:
```suggestion
						publicIPs += utils.PtrString(nic.PublicIp)
```

internal/cmd/beta/server/list/list.go
```diff
@@ -164,11 +166,36 @@ func outputResult(p *print.Printer, outputFormat string, servers []iaas.Server)
...
 
 		for i := range servers {
 			server := servers[i]
-			table.AddRow(*server.Id, *server.Name, *server.Status, *server.AvailabilityZone)
+
+			nicIPv4 := ""
+			publicIPs := ""
+			if server.Nics != nil && len(*server.Nics) > 0 {
+				for i, nic := range *server.Nics {
+					if nic.Ipv4 != nil || nic.PublicIp != nil {
```
Fyusel, line 178: If one of them is nil you will get a nil pointer exception in the following lines

internal/cmd/beta/server/list/list.go
```diff
@@ -164,11 +166,36 @@ func outputResult(p *print.Printer, outputFormat string, servers []iaas.Server)
 		return nil
 	default:
 		table := tables.NewTable()
-		table.SetHeader("ID", "Name", "Status", "Availability Zones")
+		table.SetHeader("ID", "Name", "Status", "Machine Types", "Availability Zones", "Nic IPv4", "Public IPs")
```
Fyusel, line 169:
```suggestion
		table.SetHeader("ID", "Name", "Status", "Machine Type", "Availability Zones", "Nic IPv4", "Public IPs")
```

internal/cmd/beta/server/list/list.go
marceljk: The function `utils.PtrString()` with is used below to access the value of the pointers, checks if the pointers are nil. If that's the case it returns an empty string. So it shouldn't be an issue.

Fyusel: Yes, true. I missed that.

If one of them is nil, "" and "\n" is added to the string (empty line). This is intended, right?
Output would be something like (haven't tested it yet):
```
Nic IP4_1 - Public IP_1
Nic IP_2
Nic IP_3 - Public IP_3
```

marceljk: Yes this is intended. Because if a Nic doesn't have a public ip, it shouldn't show any public ip in the same line. With adding `\n` after each iteration, it makes sure that they doesn't get mixed.

Fyusel: Great :+1:

Fyusel (approved):

PR #580, opened 2025-01-30
feat: Support dns TXT records with more than 255 characters

marceljk:


internal/cmd/dns/record-set/create/create.go
```diff
@@ -152,7 +154,23 @@ func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
 func buildRequest(ctx context.Context, model *inputModel, apiClient *dns.APIClient) dns.ApiCreateRecordSetRequest {
 	records := make([]dns.RecordPayload, 0)
 	for _, r := range model.Records {
-		records = append(records, dns.RecordPayload{Content: utils.Ptr(r)})
+		result := r
+		if len(r) > 255 && model.Type == txtType {
```
Fyusel, line 158: Can we use a const for this value?

internal/cmd/dns/record-set/update/update.go
```diff
@@ -170,7 +179,23 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *dns.APIClie
 	if model.Records != nil {
 		records = utils.Ptr(make([]dns.RecordPayload, 0))
 		for _, r := range *model.Records {
-			records = utils.Ptr(append(*records, dns.RecordPayload{Content: utils.Ptr(r)}))
+			result := r
+			if len(r) > 255 && utils.PtrString(model.Type) == txtType {
```
Fyusel, line 183: Can we extract this into a function since this code is the same for create and update, right?

internal/cmd/dns/record-set/create/create.go
Fyusel: I think there should be also a check if the length is greater than 4096. Since this is the maximum supported length by DNS service.

internal/cmd/dns/record-set/update/update.go
marceljk: Yes you're right. This makes it also better testable. I'll change it

Fyusel (approved):

PR #583, opened 2025-02-03
fix:  Add nil checks for values which are printed to the console

marceljk:


internal/cmd/beta/key-pair/describe/describe.go
```diff
@@ -145,10 +151,10 @@ func outputResult(p *print.Printer, outputFormat string, showOnlyPublicKey bool,
 			return nil
 		}
 		table := tables.NewTable()
-		table.AddRow("KEY PAIR NAME", *keyPair.Name)
+		table.AddRow("KEY PAIR NAME", utils.PtrString(keyPair.Name))
 		table.AddSeparator()
 
-		if *keyPair.Labels != nil && len(*keyPair.Labels) > 0 {
+		if keyPair.Labels != nil && len(*keyPair.Labels) > 0 {
```
bahkauv70, line 157: the nil check for the labels slice is not needed, as  go defines len(nil) as 0

internal/cmd/beta/key-pair/update/update.go
```diff
@@ -123,7 +123,18 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
 }
 
 func outputResult(p *print.Printer, model *inputModel, keyPair *iaas.Keypair) error {
```
bahkauv70, line 125: in this case the model could be passed per value as well, which would save a nil check below. Passing struct values is surprisingly efficient in go

internal/cmd/beta/network-area/describe/describe.go
```diff
@@ -174,31 +174,33 @@ func outputResult(p *print.Printer, outputFormat string, networkArea *iaas.Netwo
...
-		}
-		if networkArea.Ipv4.MinPrefixLen != nil {
-			table.AddRow("MIN PREFIX LENGTH", *networkArea.Ipv4.MinPrefixLen)
-			table.AddSeparator()
+		if networkArea.Ipv4 != nil {
+			if networkArea.Ipv4.TransferNetwork != nil {
+				table.AddRow("TRANSFER RANGE", *networkArea.Ipv4.TransferNetwork)
+				table.AddSeparator()
+			}
+			if networkArea.Ipv4.DefaultNameservers != nil && len(*networkArea.Ipv4.DefaultNameservers) > 0 {
```
bahkauv70, line 182: see previous remark regarding nil slices

internal/cmd/beta/server/command/template/list/list.go
```diff
@@ -143,7 +142,17 @@ func outputResult(p *print.Printer, outputFormat string, templates []runcommand.
 		table.SetHeader("NAME", "OS TYPE", "TITLE")
 		for i := range templates {
 			s := templates[i]
-			table.AddRow(*s.Name, strings.Join(*s.OsType, ","), *s.Title)
+
+			var osType string
+			if s.OsType != nil && len(*s.OsType) > 0 {
```
bahkauv70, line 147: len(nil)=0

internal/cmd/beta/server/command/template/list/list.go
```diff
@@ -143,7 +142,17 @@ func outputResult(p *print.Printer, outputFormat string, templates []runcommand.
...
+				osType = strings.Join(*s.OsType, ",")
```
bahkauv70, line 148: imho it would make sense to create a helper function _JoinStringPointer_ that takes a pointer to a string slice and encapsulates this functionality in a reusable fashion. Something like this:
func JoinStringPointer(vals *[]string, sep string) string {
    if vals == nil {
        return ""
    }
    return strings.Join(*vals,sep)
}

internal/cmd/beta/sqlserverflex/user/list/list.go
```diff
@@ -160,7 +160,7 @@ func outputResult(p *print.Printer, outputFormat string, users []sqlserverflex.I
 		table.SetHeader("ID", "USERNAME")
 		for i := range users {
 			user := users[i]
-			table.AddRow(*user.Id, *user.Username)
+			table.AddRow(utils.PtrString(user.Id), utils.PtrString(user.Username))
```
bahkauv70, line 163:
```suggestion
			table.AddRow(
                utils.PtrString(user.Id), 
                utils.PtrString(user.Username),
            )
```

internal/cmd/load-balancer/describe/describe.go
```diff
@@ -160,9 +161,9 @@ func buildLoadBalancerTable(loadBalancer *loadbalancer.LoadBalancer) tables.Tabl
 		networkId = *networks[0].NetworkId
 	}
 
-	externalAdress := "-"
+	externalAddress := "-"
```
bahkauv70, line 164: As this pattern keeps repeating, a utility function `PtrStringDefault[T any](t T, defaultVal string) string` would probably in order

internal/cmd/postgresflex/backup/describe/describe.go
```diff
@@ -133,13 +133,17 @@ func outputResult(p *print.Printer, cmd *cobra.Command, outputFormat string, bac
...
+		table.AddRow("ID", utils.PtrString(backup.Id))
 		table.AddSeparator()
-		table.AddRow("CREATED AT", *backup.StartTime)
+		table.AddRow("CREATED AT", utils.PtrString(backup.StartTime))
 		table.AddSeparator()
 		table.AddRow("EXPIRES AT", backupExpireDate)
 		table.AddSeparator()
-		table.AddRow("BACKUP SIZE", bytesize.New(float64(*backup.Size)))
+		backupSize := "n/a"
+		if backup.Size != nil {
```
bahkauv70, line 143: This keeps repeating as well. Another helper function suitable?

bahkauv70 (changes requested):
There are still pointer-dereferences remaining in the code. Use the following regular expressions to find them:
* `p\..*?\(.*\*.*\)`
* `table.*\(.*\*.*\)`

internal/cmd/beta/key-pair/describe/describe.go
bahkauv70: Correction from my side: This is a pointer to a slice, so the check is valid!

internal/cmd/beta/network-area/describe/describe.go
bahkauv70: Correction from my side: This is a pointer to a slice, so the check is valid!

bahkauv70 (approved):

PR #592, opened 2025-02-06
feat: region adjustment object storage

marceljk:
- adapted the commands of object-storage to the new region adjustment of `github.com/stackitcloud/stackit-sdk-go/services/objectstorage` v1.0.0

### DIFF
--- a/internal/cmd/object-storage/bucket/create/create.go
+++ b/internal/cmd/object-storage/bucket/create/create.go
@@ -122,7 +122,7 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
 }
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *objectstorage.APIClient) objectstorage.ApiCreateBucketRequest {
-	req := apiClient.CreateBucket(ctx, model.ProjectId, model.GlobalFlagModel.Region, model.BucketName)
+	req := apiClient.CreateBucket(ctx, model.ProjectId, model.Region, model.BucketName)
 	return req
 }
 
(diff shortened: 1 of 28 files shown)


internal/cmd/object-storage/bucket/create/create.go
```diff
@@ -122,7 +122,7 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
 }
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *objectstorage.APIClient) objectstorage.ApiCreateBucketRequest {
-	req := apiClient.CreateBucket(ctx, model.ProjectId, model.BucketName)
+	req := apiClient.CreateBucket(ctx, model.ProjectId, model.GlobalFlagModel.Region, model.BucketName)
```
rubenhoenle, line 125:
```suggestion
	req := apiClient.CreateBucket(ctx, model.ProjectId, model.GlobalFlagModel.Region, model.BucketName)
```

Maybe a Newbie question, but in some `buildRequest` functions you use `model.GlobalFlagModel.Region` and in others `model.Region`.

Is that intentional, and if yes, why? 😄

marceljk: Good point! No it wasn't really intended, but basically you can access the same value with `model.GlobalFlagModel.Region` and `model.Region`.

https://github.com/stackitcloud/stackit-cli/blob/7b8daa40628d0de78966b4c7cab6eb4729133d6a/internal/cmd/object-storage/bucket/create/create.go#L28

This is because of the type embedding in the model type. So you can access the attributes of `GlobalFlagModel` both ways. But I'm going to change this to stay consistent with one way.

rubenhoenle (approved):

PR #604, opened 2025-02-18
fix: `stackit beta server` commands - add nil pointer checks and tests for the outputResult functions

marceljk:
- add nil pointer checks
- add tests for the outputResult functions within the server commands

### DIFF
--- a/internal/cmd/beta/server/backup/create/create_test.go
+++ b/internal/cmd/beta/server/backup/create/create_test.go
@@ -208,3 +208,31 @@ func TestBuildRequest(t *testing.T) {
 		})
 	}
 }
+
+func TestOutputResult(t *testing.T) {
+	type args struct {
+		outputFormat string
+		serverLabel  string
+		resp         serverbackup.BackupJob
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name:    "empty",
+			args:    args{},
+			wantErr: false,
+		},
+	}
+	p := print.NewPrinter()
+	p.Cmd = NewCmd(p)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := outputResult(p, tt.args.outputFormat, tt.args.serverLabel, tt.args.resp); (err != nil) != tt.wantErr {
+				t.Errorf("outputResult() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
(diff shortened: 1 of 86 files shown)


internal/cmd/beta/server/backup/create/create_test.go
```diff
@@ -208,3 +208,31 @@ func TestBuildRequest(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name:    "empty",
+			args:    args{},
+			wantErr: false,
+		},
```
Fyusel, line 227: Should we add more test cases here? Like for each output format? (none, json, yaml)

Fyusel: Also for the other tests

internal/cmd/beta/server/os-update/create/create.go
```diff
@@ -60,8 +62,19 @@ func NewCmd(p *print.Printer) *cobra.Command {
 				return err
 			}
 
+			serverLabel := model.ServerId
+			// Get server name
+			if iaasApiClient, err := iaasClient.ConfigureClient(p); err == nil {
+				serverName, err := iaasUtils.GetServerName(ctx, iaasApiClient, model.ProjectId, model.ServerId)
+				if err != nil {
+					p.Debug(print.ErrorLevel, "get server name: %v", err)
+				} else if serverName != "" {
```
Fyusel, line 71: nitpick: different code style, sometimes "else if" sometimes "only if"

internal/cmd/beta/server/backup/create/create_test.go
marceljk: In the beginning I also added tests for json and yaml, but in general they can handle everything without panic or returning an error. So even if a pointer is nil and the nil check is removed, they return null as a result, which is fine from myside. And the part for json and yaml output is always the same. Only the default output is always different with our custom code, which needs to be tested to make sure there is no nil check missing.
But we can discuss this in our daily.

internal/cmd/beta/server/os-update/create/create.go
marceljk: I take a second look and make this more consistent

marceljk: I updated it.
Please keep aware that there are still two different styles to get the server name.
This is because in some cases the iaasClient is needed anyway, and the command will fail if it can't configure it. And in some cases the iaasClient is only needed to get the server name. If it fails in the second case, then it can be ignored and the server-id will be taken. So the handling is a bit different there

Fyusel (approved):

PR #605, opened 2025-02-19
fix: stackit beta network-interface commands - add nil pointer checks and tests for the outputResult functions

marceljk:
- add nil pointer checks
- add tests for the outputResult functions within the `network-interface` commands

### DIFF
--- a/internal/cmd/beta/network-interface/list/list.go
+++ b/internal/cmd/beta/network-interface/list/list.go
@@ -83,6 +83,9 @@ func NewCmd(p *print.Printer) *cobra.Command {
 					p.Debug(print.ErrorLevel, "get network name: %v", err)
 					networkLabel = *model.NetworkId
 				}
+				if networkLabel == "" {
+					networkLabel = *model.NetworkId
+				}
 				p.Info("No network interfaces found for network %q\n", networkLabel)
 				return nil
 			}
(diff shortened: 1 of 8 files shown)


internal/cmd/beta/network-interface/list/list.go
```diff
@@ -83,6 +83,9 @@ func NewCmd(p *print.Printer) *cobra.Command {
 					p.Debug(print.ErrorLevel, "get network name: %v", err)
 					networkLabel = *model.NetworkId
 				}
+				if networkLabel == "" {
```
rubenhoenle, line 86:
```suggestion
				else if networkLabel == "" {
```

can be an else if, the same value was already assigned inside the previous if statement

rubenhoenle (dismissed):

internal/cmd/beta/network-interface/update/update_test.go
```diff
@@ -291,3 +291,38 @@ func TestBuildRequest(t *testing.T) {
...
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name:    "empty",
+			args:    args{},
+			wantErr: false,
```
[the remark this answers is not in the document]
marceljk: Yes, I forgot to add an error to the outputFunction when nic == nil. I'll update it

internal/cmd/beta/network-interface/list/list.go
marceljk: yes makes sense. I'll update it

rubenhoenle (approved):

rubenhoenle (approved):

Issue #574, opened 2025-01-24
cli: stackit volume delete prints empty name


j-b-e:
Hi,

deleteing a Volume which has no name, (e.g. an automatically created boot volume) prints an empty string on cli.

```
$ stackit beta volume list
 ID                                   │ NAME          │ STATUS    │ SERVER                               │ AVAILABILITY ZONE │ SIZE (GB) 
──────────────────────────────────────┼───────────────┼───────────┼──────────────────────────────────────┼───────────────────┼───────────
 8f197065-6c3b-4029-be47-6df3358177f7 │               │ AVAILABLE │                                      │ eu01-m            │        20
```

```
$ stackit beta volume delete 8f197065-6c3b-4029-be47-6df3358177f7 -y
Deleting volume ✓
Deleted volume ""
```
I suggest to fallback to the Id.

```
$ stackit --version
STACKIT CLI (beta)
Version: 0.21.2 (2025-01-17)
```

bahkauv70:
This is an already known issue in the CLI, which will be addressed soon. Please use the JSON output as a workaround for now

PR #623, opened 2025-02-21
chore(volume): Add nil pointer checks and tests for outputResult func…

Fyusel:
fixes: #574

### DIFF
--- a/internal/cmd/beta/volume/list/list_test.go
+++ b/internal/cmd/beta/volume/list/list_test.go
@@ -4,13 +4,12 @@ import (
 	"context"
 	"testing"
 
-	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
-
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/uuid"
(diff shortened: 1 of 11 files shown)


internal/cmd/beta/volume/list/list_test.go
```diff
@@ -202,3 +201,30 @@ func TestBuildRequest(t *testing.T) {
...
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name:    "empty",
+			args:    args{},
+			wantErr: false,
+		},
+	}
```
marceljk, line 220: Can you add here a test case with an empty volume within the volumes slice?
```
		{
			name:    "set empty volume",
			args:    args{
				volumes: []iaas.Volume{{}},
			},
			wantErr: false,
		},
```

internal/cmd/beta/volume/performance-class/list/list_test.go
```diff
@@ -202,3 +201,30 @@ func TestBuildRequest(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name:    "empty",
+			args:    args{},
+			wantErr: false,
+		},
```
marceljk, line 219: And also here

marceljk:
https://github.com/stackitcloud/stackit-cli/blob/31084ad3c4e1446d0b07349dae7759d618e137da/internal/cmd/beta/volume/delete/delete.go#L59-L64
Could you add an additional check here if the volume name is empty? If it's empty, it should use the volumeId as label
This would solve this issue #574

marceljk (approved):

PR #626, opened 2025-02-21
chore (postgresflex): Fix potential nil pointer and add test cases

bahkauv70:


internal/cmd/postgresflex/instance/update/update.go
```diff
@@ -307,8 +307,15 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient PostgreSQLFl
 	return req, nil
 }
 
-func outputResult(p *print.Printer, model *inputModel, instanceLabel string, resp *postgresflex.PartialUpdateInstanceResponse) error {
-	switch model.OutputFormat {
+func outputResult(p *print.Printer, outputFormat string, async bool, model *inputModel, instanceLabel string, resp *postgresflex.PartialUpdateInstanceResponse) error {
```
Fyusel, line 310: model parameter can be removed

internal/cmd/postgresflex/options/options.go
```diff
@@ -166,10 +166,10 @@ func buildAndExecuteRequest(ctx context.Context, p *print.Printer, model *inputM
 		}
 	}
 
-	return outputResult(p, model, flavors, versions, storages)
+	return outputResult(p, model.OutputFormat, *model, flavors, versions, storages)
```
Fyusel, line 169: I would not additionally pass _model.OutputFormat_ when the whole model is passed and needed.

internal/cmd/postgresflex/instance/update/update.go
bahkauv70: Removed

Fyusel (approved):

PR #652, opened 2025-03-17
chore(loadbalancer): regional adjustment

rubenhoenle:
relates to STACKITCLI-127


marceljk (commented):
Looks in general good! The only thing what needs to be changed is the following:

From:
https://github.com/stackitcloud/stackit-cli/blob/526183ac2336f94d449850bd1e7962bb27a9ee35/internal/pkg/services/load-balancer/client/client.go#L30-L31
to: 
```go
} else {
	cfgOptions = append(cfgOptions, authCfgOption)
}
```

Because with the current state, the region config will be used to define the region within the client. This isn't possible after the region adjustment and throws an error

marceljk (approved):

PR #661, opened 2025-03-21
replace STACKIT logo on login page

marceljk:


internal/pkg/auth/user_login.go
```diff
@@ -232,6 +233,20 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 		}
 	})
 
+	mux.HandleFunc(logoPath, func(w http.ResponseWriter, _ *http.Request) {
+		defer cleanup(server)
+
+		img, err := htmlContent.ReadFile(path.Join(htmlTemplatesPath, logoSVGFilePath))
```
rubenhoenle, line 238: Why not just use a go embed? Less error-prone, no error handling needed, ...

```go
//go:embed templates/login-successful.html
var htmlTemplateContent string

//go:embed templates/stackit_nav_logo_light.svg
var logoSvgContent []byte
```

internal/pkg/auth/user_login.go
```diff
@@ -258,6 +260,13 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 	return nil
 }
 
+// base64Encode encodes a []byte to a base64 representation as string
+func base64Encode(message []byte) string {
```
rubenhoenle, line 264: Very nice solution overall! One small nitpick, this belongs into a global util pkg IMO because it's not tied to the single login page use-case. What do you think?

rubenhoenle (dismissed):

internal/pkg/auth/user_login.go
marceljk: yes, makes sense. I'll update this

rubenhoenle (approved):

Issue #664, opened 2025-03-24
Feature Request: add `--disable-writing` to `auth` commands & add `--api-token` as a global flag


mardonner:
Hey,
during my deployment pipelines, I'm trying to keep ALL my secrets in the environment variables. With this self-imposed restriction, I'm having trouble with `stackit auth` under these circumstances.

To illustrate, imagine something like this as the last step of a initial deployment pipeline:

```bash
stackit config set --project-id $(STACKIT_PROJECT_ID)
stackit auth activate-service-account --service-account-key-path <(echo $STACKIT_SERVICE_ACCOUNT_KEY)

KUBECONFIG=$(stackit ske kubeconfig create <mycluster> --disable-writing --output-format json -y) \
helm install argocd argo-cd \
--kubeconfig <(echo $KUBECONFIG) \
--repo <private-artifactory-url> \
--username $(artifactory_technical_user_name) \
--password $(artifactory_technical_user_token) \
-f <(envsubst < myvalues.yaml.template)
# myvalues contains secrets like git token and imagepullsecret in the 'extraObjects:' section in helm the chart
```

This authenticates against STACKIT, creates a kubeconfig and deploys a minimal argoCD instance via helm, that has just enough configuration in order to connect to our gitops repo where the applicationsets and an application for self-managed argo live.
I'm explicitly trying to avoid using the helm and argocd terraform providers. I'm also not using `helm repo add` so that I don't store credentials in `~/.config/helm/repositories.yaml`. I'm also using process substitution to pass the service account key to stackit-cli and the kubeconfig to helm, since these expect file path and as I said, don't want to write files to disk.

---
Now to how I would like this to work:
In my world, this script above has one downside. At the moment there is no way around `stackit auth` writing the api token to disk. Something I'm trying to avoid. I would like to have the option to not write the token into `~/.config/stackit/cli-auth-storage.txt` but instead either to stdout or into a environment variable. If written to stdout, I would use something like `export STACKIT_API_TOKEN=$(stackit auth activate-service-account --service-account-key-path <(echo $STACKIT_SERVICE_ACCOUNT_KEY) --disable-writing)`.

The second thing I would need for this to work is for `stackit`-cli to either respect this environment variable or have an extra flag like `--api-token` so that I can get my kubeconfig like this:
```bash
KUBECONFIG=$(stackit ske kubeconfig create <mycluster> --disable-writing --output-format json -y --api-token $STACKIT_API_TOKEN)
```

[...]

PR #682, opened 2025-04-04
feat(auth): Introduce possibility to use an environment variable to a…

Fyusel:
…uthenticate

If STACKIT_ACCESS_TOKEN is set this environment variable is used instead of stored tokens. Additionally the activate-service-account command is extended in order to only print the token but not store them in the keyring or in a file on the disk.

## Description

relates to #664

## Checklist

- [x] Issue was linked above
- [x] Docs are up-to-date: `make generate-docs`

### DIFF
--- a/AUTHENTICATION.md
+++ b/AUTHENTICATION.md
@@ -13,6 +13,8 @@ $ stackit auth activate-service-account
 
 You can also configure the service account credentials directly in the CLI. To get help and to get a list of the available options run the command with the `-h` flag.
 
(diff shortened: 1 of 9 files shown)


AUTHENTICATION.md
```diff
@@ -13,6 +13,8 @@ $ stackit auth activate-service-account
 
 You can also configure the service account credentials directly in the CLI. To get help and to get a list of the available options run the command with the `-h` flag.
 
+**_Note:_** There is an optional flag `--only-print-access-token` which can be used to only obtain the access token which prevents writing the credentials to the keyring or into `~/.config/stackit/cli-auth-storage.txt`. This access token can be stored as environment variable (STACKIT_ACCESS_TOKEN) in order to be used for all subsequent commands by default.
```
marceljk, line 16: On mac there is the path a bit different when writing to keyring fails. In this case the path is ~/Library/Application Support/stackit/cli-auth-storage.txt

Fyusel: Good point :+1:

marceljk (approved):

PR #778, opened 2025-06-04
fix(cli): List mongodb storages returns empty result 

Benjosh95:
## Description

The issue is that the condition len(*options.Storages.Storages.StorageClasses) == 0 is incorrect - it's actually preventing the table from being displayed when there are storage classes available. It should be checking if the length is greater than 0, not equal to 0.

Jira-Ticket: stackitcli-200

## Checklist

- [x] Issue was linked above

## Testing-Instructions

$ stackit mongodbflex options --flavor-id 1.2 --storages 
returning a table with the requested content.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -207,7 +207,7 @@ require (
 	go.uber.org/zap v1.24.0 // indirect
 	golang.org/x/exp/typeparams v0.0.0-20250210185358-939b2ce775ac // indirect
 	golang.org/x/sync v0.15.0 // indirect
-	golang.org/x/tools v0.33.0 // indirect
+	golang.org/x/tools v0.34.0 // indirect
 	google.golang.org/protobuf v1.36.6 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	honnef.co/go/tools v0.6.1 // indirect


Fyusel (dismissed):

internal/cmd/mongodbflex/options/options.go
```diff
@@ -225,7 +225,7 @@ func outputResultAsTable(p *print.Printer, model *inputModel, options *options)
 	if model.Versions && len(*options.Versions) != 0 {
 		content = append(content, buildVersionsTable(*options.Versions))
 	}
-	if model.Storages && options.Storages.Storages != nil && len(*options.Storages.Storages.StorageClasses) == 0 {
```
rubenhoenle, line 228: Please add a test case for this which fails without your fix and succeeds with your fix :)

thorbenbelow: Hi @rubenhoenle, slightly off-topic: I just ran into the same issue with postgres. 
https://github.com/stackitcloud/stackit-cli/blob/main/internal/cmd/postgresflex/options/options.go#L225

I'm not familiar with the codebase, but from a first look, the lack of a test for this seems consistent with the rest of it.  
From what I saw in this and the other test files, `Test_outputResult` mainly checks for errors but doesn’t verify the actual output content, and it looks like `print.Printer` doesn’t currently make that very straightforward either.  
  
So I was wondering, do you expect the output to be validated for completeness here, or is the idea that it should just return an error if, for example, the output is empty?

rubenhoenle: sorry, my bad. Fine then :smile:

rubenhoenle (dismissed):

internal/cmd/mongodbflex/options/options.go
rubenhoenle: Hi @thorbenbelow,

thanks for the report for the postgres options, we will also fix this within this PR.

For the tests you are right, there's currently not really an option to test this with our `print.Printer` setup. When I wrote this comment I didn't have this in mind, sorry for the confusion. So we won't add further testcases here for now.

rubenhoenle:
- [ ] Fix also the postgres option before merge: https://github.com/stackitcloud/stackit-cli/pull/778#discussion_r2128617896

PR #774, opened 2025-06-02
feat(cli): onboard iaas snapshot api

Benjosh95:
## Description

Onboarding the snapshot API https://docs.api.stackit.cloud/documentation/iaas/version/v1#tag/Snapshots to manage the snapshot resources via the cli.

## Checklist

- [x] Issue was linked above
- [x] Code format was applied: `make fmt`
- [x] Examples were added / adjusted (see e.g. [here](https://github.com/stackitcloud/stackit-cli/blob/ef291d1683ca5b0d719ec0a26ecb999a32685117/internal/cmd/ske/cluster/create/create.go#L49-L63))
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI)
- [x] Unit tests got implemented or updated
- [x] Unit tests are passing: `make test` (will be checked by CI)
- [x] No linter issues: `make lint` (will be checked by CI) 

## Testing-Instructions

1. Create a volume that will be used for snapshot testing.  
   stackit volume create \
     --availability-zone eu01-1 \
     --size 10 \
     --name "snapshot-test-vol" \
     --labels purpose=snapshot-testing

2. Create a snapshot of the volume above
   - Required `volume-id` flag
   - Optional name and labels
   - Async operation  
   stackit volume snapshot create \
     --volume-id $VOLUME_ID \
     --name "test-snap-1" \
     --labels purpose=testing,test-type=minimal

3. List snapshot(s) 
   stackit volume snapshot list --limit 3

4. Describe snapshot  
   stackit volume snapshot describe $SNAPSHOT_ID

5. Update snapshot   
   stackit volume snapshot update $SNAPSHOT_ID \
     --name "test-snap-1-updated" \
     --labels purpose=testing-updated

6. Verify update  
   stackit volume snapshot describe $SNAPSHOT_ID

7. Delete the snapshot
   stackit volume snapshot delete $SNAPSHOT_ID 

8. Delete the volume  
   stackit volume delete $VOLUME_ID

PR #773, opened 2025-06-01
feat(cli): onboard iaas volume backup api 

Benjosh95:
## Description

Onboarding the volume backup API https://docs.api.stackit.cloud/documentation/iaas/version/v1#tag/Backups to manage the backup resources via the cli.

## Checklist

- [x] Issue was linked above
- [x] Code format was applied: `make fmt`
- [x] Examples were added / adjusted (see e.g. [here](https://github.com/stackitcloud/stackit-cli/blob/ef291d1683ca5b0d719ec0a26ecb999a32685117/internal/cmd/ske/cluster/create/create.go#L49-L63))
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI)
- [x] Unit tests got implemented or updated
- [x] Unit tests are passing: `make test` (will be checked by CI)
- [x] No linter issues: `make lint` (will be checked by CI) 

## Testing-Instructions

1. Create a test volume to have something to backup
stackit volume create --name backup-test-vol --size 10 --performance-class storage_premium_perf0 --availability-zone eu01-1 --labels initial=true,purpose=backup-testing 

2. Create backup of the volume with name and labels
stackit volume backup create --source-id <VOLUME_ID> --source-type volume --name first-backup --labels state=original,test=true

3. List backups with limit and label selector
stackit volume backup list --limit 5 --label-selector state=original

4. Update backup with new name and labels
stackit volume backup update <BACKUP_ID> --name updated-backup --labels state=modified,test=true

5. Describe backup to verify changes
stackit volume backup describe <BACKUP_ID>

6. Restore backup
stackit volume backup restore <BACKUP_ID> 

7. Clean up by deleting backup
stackit volume backup delete <BACKUP_ID>

8. Clean up by deleting volume
stackit volume delete <VOLUME_ID>

PR #790, opened 2025-06-10
refactor(cli): add ConvertStringMapToInterfaceMap util function 

Benjosh95:
## Description

Codebase contained multiple implementations of converting `map[string]string` to `*map[string]interface{}`. This pattern appears in many command files, particularly in the IaaS service commands (volume, backup, network, etc.) when handling label conversions. The code got extracted into util function `ConvertStringMapToInterfaceMap` and refactored.

Jira Ticket: [https://jira.schwarz/browse/STACKITCLI-203](url)

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/server/update/update.go
+++ b/internal/cmd/server/update/update.go
@@ -131,7 +131,7 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *iaas.APICli
 
 	payload := iaas.UpdateServerPayload{
 		Name:   model.Name,
-		Labels: utils.ConvertStringMapToInterfaceMap(*model.Labels),
+		Labels: utils.ConvertStringMapToInterfaceMap(model.Labels),
 	}
 
 	return req.UpdateServerPayload(payload)
--- a/internal/cmd/server/update/update.go
+++ b/internal/cmd/server/update/update.go
@@ -129,18 +129,9 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
 func buildRequest(ctx context.Context, model *inputModel, apiClient *iaas.APIClient) iaas.ApiUpdateServerRequest {
 	req := apiClient.UpdateServer(ctx, model.ProjectId, model.ServerId)
 
-	var labelsMap *map[string]interface{}
-	if model.Labels != nil && len(*model.Labels) > 0 {
-		// convert map[string]string to map[string]interface{}
-		labelsMap = utils.Ptr(map[string]interface{}{})
-		for k, v := range *model.Labels {
-			(*labelsMap)[k] = v
-		}
-	}
-
 	payload := iaas.UpdateServerPayload{
 		Name:   model.Name,
-		Labels: labelsMap,
+		Labels: utils.ConvertStringMapToInterfaceMap(*model.Labels),
 	}
 
 	return req.UpdateServerPayload(payload)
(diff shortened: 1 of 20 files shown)


Benjosh95:
- [ ] Needs to refactor conversions in https://github.com/stackitcloud/stackit-cli/pull/774 and https://github.com/stackitcloud/stackit-cli/pull/773 aswell in this PR when merge completed

internal/pkg/utils/utils.go
```diff
@@ -129,3 +129,16 @@ func Base64Encode(message []byte) string {
 func UserAgentConfigOption(cliVersion string) sdkConfig.ConfigurationOption {
 	return sdkConfig.WithUserAgent(fmt.Sprintf("stackit-cli/%s", cliVersion))
 }
+
+// ConvertStringMapToInterfaceMap converts a map[string]string to a pointer to map[string]interface{}.
+// Returns nil if the input map is empty.
+func ConvertStringMapToInterfaceMap(m map[string]string) *map[string]interface{} {
```
marceljk:
```suggestion
func ConvertStringMapToInterfaceMap(m map[string]string) *map[string]interface{} { //nolint:gocritic
```

marceljk: Can you add some testcases for this function?

internal/cmd/server/update/update.go
```diff
@@ -129,18 +129,9 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
-		labelsMap = utils.Ptr(map[string]interface{}{})
-		for k, v := range *model.Labels {
-			(*labelsMap)[k] = v
-		}
-	}
-
 	payload := iaas.UpdateServerPayload{
 		Name:   model.Name,
-		Labels: labelsMap,
+		Labels: utils.ConvertStringMapToInterfaceMap(*model.Labels),
```
marceljk: In all cases the model.Labels is a pointer. In the current way this could lead to some nil pointer exceptions and some existing tests fail because of this.
I would suggest, to change the utils function, that it receive a pointer to a map and then you can add in the utils function the check, if the pointer is nil.
```suggestion
		Labels: utils.ConvertStringMapToInterfaceMap(model.Labels),
```

rubenhoenle: +1 from me for this, Marcel got a good point here

marceljk (dismissed):
Looks good and works well! Can be merged

internal/pkg/utils/utils.go
```diff
@@ -129,3 +129,18 @@ func Base64Encode(message []byte) string {
 func UserAgentConfigOption(cliVersion string) sdkConfig.ConfigurationOption {
 	return sdkConfig.WithUserAgent(fmt.Sprintf("stackit-cli/%s", cliVersion))
 }
+
+// ConvertStringMapToInterfaceMap converts a map[string]string to a pointer to map[string]interface{}.
+// Returns nil if the input map is empty.
+//
+//nolint:gocritic
```
rubenhoenle: Maybe we could have some short comment which answers the question
- why does the linter complain?
- why do we ignore the linter and use this code anyways?

See e.g. https://github.com/stackitcloud/terraform-provider-stackit/blob/281d31f615c91196a5d8ff648f1e05b0fd496668/stackit/internal/services/iaas/utils/util.go#L36

Will make future contributors more comfortable with using this util func. :smile:

rubenhoenle (dismissed):

rubenhoenle (approved):

PR #811, opened 2025-06-18
fix: replaced deprecated brews config in goreleaser

marceljk:
## Description

The `brews` config is deprecated and needs to be replaced by homebrew_casks

relates to #1234

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -32,6 +32,7 @@ builds:
       - amd64
     hooks:
       post:
+        # Signing
         - cmd: codesign -s "{{.Env.APPLE_APPLICATION_IDENTITY}}" -f -v --options=runtime "dist/macos-builds_{{.Target}}/{{.Name}}"
           output: true
         - cmd: codesign -vvv --deep --strict "dist/macos-builds_{{.Target}}/{{.Name}}"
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -46,12 +47,29 @@ builds:
           output: true
         - cmd: spctl -a -t open --context context:primary-signature -v dist/{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}.dmg
           output: true
+        # Completion files
+        - cmd: mkdir -p dist/macos-builds_{{.Target}}/completions
+        - cmd: sh -c './dist/macos-builds_{{.Target}}/{{.Name}} completion zsh > ./dist/macos-builds_{{.Target}}/completions/stackit.zsh'
```
rubenhoenle, line 52: why do you need the `sh -c` here?

marceljk: I tested it first without `sh -c` but then it doesn't pipe the content into the files. And with adding `sh -c` it works

rubenhoenle: Did you try this one?
```suggestion
        - cmd: "./dist/macos-builds_{{.Target}}/{{.Name}} completion zsh > ./dist/macos-builds_{{.Target}}/completions/stackit.zsh"
```

marceljk: I just tested it and it doesn't work, so the files are not created

Fyusel (approved):

PR #933, opened 2025-08-20
fix: remove access_token prefix from YAML output in get-access-token

Benjosh95:
## Description

relates to https://jira.schwarz/browse/STACKITCLI-236

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/auth/get-access-token/get_access_token.go
+++ b/internal/cmd/auth/get-access-token/get_access_token.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"github.com/goccy/go-yaml"
 	"github.com/spf13/cobra"
 	"github.com/stackitcloud/stackit-cli/internal/cmd/params"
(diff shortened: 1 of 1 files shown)


internal/cmd/auth/get-access-token/get_access_token.go
```diff
@@ -61,13 +60,7 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
...
 				return nil
 			case print.YAMLOutputFormat:
-				details, err := yaml.MarshalWithOptions(map[string]string{
-					"access_token": accessToken,
-				}, yaml.IndentSequence(true), yaml.UseJSONMarshaler())
-				if err != nil {
-					return fmt.Errorf("marshal image list: %w", err)
-				}
-				params.Printer.Outputln(string(details))
+				params.Printer.Outputln(accessToken)
```
Fyusel: nitpick: That is the same as "default" so we can get rid of the Yaml case?

PR #954, opened 2025-09-04
fix: server yaml output of userdata field displayed as bytearray instead of base64

Benjosh95:
## Description

Jira: https://jira.schwarz/browse/STACKITCLI-219

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/server/describe/describe.go
+++ b/internal/cmd/server/describe/describe.go
@@ -30,6 +30,26 @@ type inputModel struct {
 	ServerId string
 }
 
+// convertServerForYAML creates a map with UserData as base64 string for YAML output
+func convertServerForYAML(server *iaas.Server) map[string]interface{} {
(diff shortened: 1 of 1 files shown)


internal/cmd/server/describe/describe.go
```diff
@@ -30,6 +30,26 @@ type inputModel struct {
...
+	}
+
+	var serverMap map[string]interface{}
+	if err := json.Unmarshal(jsonData, &serverMap); err != nil {
+		return nil
+	}
+
+	return serverMap
+}
+
```
marceljk, line 52: I think it would be better to move this into the utils package and make the function more generic. So that it's not specific for `iaas.Server` and can be used by other types as well. In `$ stackit server list -o yaml`, we have the same issue and with a more generic utils function, we could reuse it there

PR #981, opened 2025-09-10
fix: use refresh tokens in curl commands for user token flow

Benjosh95:
## Description

Use existing GetValidAccessToken in curl commands to automatically refresh expired tokens for user token flows

Jira: https://jira.schwarz/browse/STACKITCLI-22

## Checklist

- [ ] Issue was linked above


internal/pkg/auth/auth.go
```diff
@@ -202,7 +202,7 @@ func GetValidAccessToken(p *print.Printer) (string, error) {
 	// Refresh the tokens
 	err = refreshTokens(utf)
 	if err != nil {
-		return "", fmt.Errorf("refresh access token: %w", err)
+		return "", fmt.Errorf("access token expired and refresh token is invalid. Please run 'stackit auth login' to re-authenticate: %w", err)
```
marceljk: I think we could improve the error handling here, because the code in `pkg` should be independent of the CLI. So that potential other tools can use it as well.
Maybe we can just return the error here as suggested
```suggestion
		return "", fmt.Errorf("access token and refresh token expired: %w", err)
```
and at the place where this function is called, we print a info that the session is expired and the user should re-authenticate, but don't return the error. Because usually for the users it doesn't matter what's the exact error.  Something like this:
```go
accessToken, err := auth.GetValidAccessToken(p)
if err != nil {
		p.Debug(print.ErrorLevel, "get valid access token: %v", err)
		return "", &errors.SessionExpiredError{}
}
```
And in case a user execute the command with `--verbosity debug`, he can see what's the error.

Issue #893, opened 2025-07-28
cli: --output-format flag is not always honored


lilioid:
## Description

The `--output-format` flag is not always honored. 
For example, if one runs a `stackit ske cluster list -o json` on a project that has SKE enabled but where no clusters yet exist, the CLI only outputs a plaintext error.
With the current behavior, it is not easily possible to use the STACKIT cli in scripts because one cannot always parse the output correctly.

## Steps to reproduce

1. Run `stackit ske cluster list --project-id=… -o json` on a project where SKE is enabled but where no cluster has yet been created

## Actual behavior

The CLI exits with exit code 0 (which is fine) and the following plaintext output:
```
No clusters found for project "…"
```

## Expected behavior

The CLI should exit with code 0 and print `[]` to stdout.

## Environment
 - OS: Linux
 - Version of STACKIT CLI (see `stackit --version`): `v0.32.0`

Fyusel:
Hi @lilioid,
thank you for reporting this issue. We will have a look at it.

rubenhoenle:
Hi @lilioid,

just wanted to let you know that we fixed this for SKE (https://github.com/stackitcloud/stackit-cli/commit/a0eee4ef7bbed38042fa10dc99d0533166671142) and we're working on fixing it for all other list commands also :)

Best regards
Ruben

KevinGimbel:
Hello! It seems this is also the case with `stackit auth login` - the `-o` flag does nothing there.

KevinGimbel:
I expected to get a credentials.json which i could store into a file -- I guess `get-access-token` is what I should be using instead!

rubenhoenle:
Hi @lilioid,

thanks again for your bug report. We fixed this now across all subcommands of our STACKIT CLI. Therefore we will close this issue. In case you encounter any further problems feel free to open up another issue.

Best regards
Ruben from STACKIT Developer Tools team

PR #993, opened 2025-09-22
feat: provide functionality to list public ip-ranges

h3adex:
## Description

Adds functionality to list STACKIT public IP ranges.

The `-o pretty` flag does not render the output as a table, as this format wouldn't be useful. Instead, it joins all IP ranges with commas. This is helpful if you need to send the IP ranges to a customer or colleague.

E.g.
Screenshot 2025-09-22 at 10 16 19

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/pkg/utils/utils_test.go
+++ b/internal/pkg/utils/utils_test.go
@@ -580,6 +580,7 @@ func TestGetSliceFromPointer(t *testing.T) {
 			result := GetSliceFromPointer(tt.input)
 
 			if result == nil {
+				t.Errorf("GetSliceFromPointer() = %v, want %v", result, tt.expected)
 				return
 			}
 


rubenhoenle:
> E.g. Screenshot 2025-09-22 at 10 16 19

Could we have this output with line breaks please?

This makes sense for users which want to work with tools like `grep` or e.g. `fzf`.

For reference, I personally love to work with CLIs like that: `stackit public-ip ip-ranges list | fzf` 
And I think I'm not the only one :smile:

h3adex:
Sorry! Accidentally amended the changes.

internal/cmd/public-ip/ip-ranges/ip-ranges.go
```diff
@@ -0,0 +1,26 @@
+package ipranges
```
Benjosh95: Please rename all occurences of "ip-ranges" into "ranges" (docs, directory, package and imports) 

Desired command naming "stackit public-ip ranges list"

Benjosh95 (changes requested):

internal/cmd/public-ip/ip-ranges/list/list.go
```diff
@@ -0,0 +1,122 @@
...
+		if err != nil {
+			return fmt.Errorf("marshal public IP: %w", err)
+		}
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		var publicIps []string
+		for _, item := range *networkListResponse.Items {
+			if item.Cidr != nil || *item.Cidr != "" {
```
Benjosh95: Can panic if item.Cidr is null

```suggestion
			if item.Cidr != nil && *item.Cidr != "" {
```

Benjosh95 (changes requested):

internal/cmd/public-ip/ip-ranges/list/list.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+	}
+	return cmd
+}
+
+func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
+	globalFlags := globalflags.Parse(p, cmd)
+	if globalFlags.ProjectId == "" {
+		return nil, &errors.ProjectIdError{}
+	}
```
Benjosh95: The project ID shouldn't be required since it's a global command and isn't tied to a specific project

```suggestion
```

internal/cmd/public-ip/ranges/ip-ranges.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmd(params *params.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ip-ranges",
+		Short: "Provides functionality for STACKIT Cloud public ip-ranges",
```
Benjosh95: Please also rename all occurences of **_STACKIT Cloud_** to _**STACKIT**_ to be more consistent with the codebase.

```suggestion
		Short: "Provides functionality for STACKIT public-ip ranges",
```

internal/cmd/public-ip/ip-ranges/list/list.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+}
+
+func NewCmd(params *params.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "Lists all STACKIT cloud public IP ranges.",
```
Benjosh95: The "Short" usually has no trailing period

```suggestion
		Short: "Lists all STACKIT public-ip ranges",
```

internal/cmd/public-ip/ranges/ip-ranges.go
```diff
@@ -0,0 +1,26 @@
...
+		Long:  "Provides functionality for STACKIT Cloud public ip-ranges",
```
Benjosh95: The "Long" usually has a trailing period
```suggestion
		Long:  "Provides functionality for STACKIT Cloud public ip-ranges.",
```

Benjosh95 (changes requested):

internal/cmd/public-ip/ranges/list/list.go
```diff
@@ -0,0 +1,153 @@
...
+			// Call API
+			req := apiClient.ListPublicIPRanges(ctx)
+			resp, err := req.Execute()
+
+			if err != nil {
+				return fmt.Errorf("list public IP ranges: %w", err)
+			}
+
+			if resp.Items == nil || len(*resp.Items) == 0 {
+				params.Printer.Info("No public IP ranges found\n")
```
rubenhoenle: Invalid JSON/YAML output. See https://github.com/stackitcloud/stackit-cli/issues/893 and e.g. https://github.com/stackitcloud/stackit-cli/commit/a0eee4ef7bbed38042fa10dc99d0533166671142 for reference.

internal/cmd/public-ip/ranges/list/list.go
```diff
@@ -0,0 +1,153 @@
...
+			return fmt.Errorf("marshal public IP ranges: %w", err)
+		}
+		p.Outputln(string(details))
+
+		return nil
+	default:
+		var cidrs []string
+		for _, item := range publicIpRanges {
+			if item.Cidr != nil && *item.Cidr != "" {
+				cidrs = append(cidrs, *item.Cidr)
```
rubenhoenle:
```suggestion
				p.Outputln(*item.Cidr)
```

Why create a new `cidrs` slice if you can just output it here? :smile:

internal/cmd/public-ip/ranges/list/list.go
```diff
@@ -0,0 +1,148 @@
...
+				return err
+			}
+
+			// Call API
+			req := apiClient.ListPublicIPRanges(ctx)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("list public IP ranges: %w", err)
+			}
+			publicIpRanges := *resp.Items
```
rubenhoenle: Now you're doing a potential nil pointer dereference here without any additional checks :smiling_face_with_tear:

rubenhoenle: Feel free to add something like this to the utils package

```golang
func GetSliceFromPointer[T any](s *[]T) []T {
    if s == nil {
        return []T{} 
    }
    return *s
}
```

Then you can do 
```golang
publicIpRanges := utils.GetSliceFromPointer(resp.Items)
```
and it's nil-safe.

internal/pkg/utils/utils.go
```diff
@@ -250,3 +250,11 @@ func ConvertToBase64PatchedServers(servers []iaas.Server) []Base64PatchedServer
 
 	return result
 }
+
+// GetSliceFromPointer returns the value of a pointer to a slice of type T. If the pointer is nil, it returns an empty slice.
+func GetSliceFromPointer[T any](s *[]T) []T {
```
rubenhoenle, line 256: Could we have some tests for that one? :smile:

internal/cmd/public-ip/ranges/list/list.go
```diff
@@ -0,0 +1,149 @@
...
+				return err
+			}
+
+			// Call API
+			req := apiClient.ListPublicIPRanges(ctx)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("list public IP ranges: %w", err)
+			}
+			publicIpRanges := utils.GetSliceFromPointer(resp.Items)
```
rubenhoenle, line 70: Nice one! Keep this one in mind for future code reviews @Benjosh95 @marceljk @Fyusel

internal/pkg/utils/utils_test.go
```diff
@@ -544,3 +544,61 @@ func TestBase64Bytes_MarshalYAML(t *testing.T) {
...
+			expected:  []string{"item1", "item2"},
+			expectNil: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := GetSliceFromPointer(tt.input)
+
+			if tt.expectNil {
```
rubenhoenle: Whats's the point of `expectNil`? Don't get it tbh. The func should never return nil.

internal/pkg/utils/utils_test.go
```diff
@@ -544,3 +544,61 @@ func TestBase64Bytes_MarshalYAML(t *testing.T) {
...
+			expectNil: false,
+		},
+		{
+			name: "pointer to nil slice",
+			input: func() *[]string {
+				var s []string
+				return &s
+			}(),
+			expected:  nil,
+			expectNil: true,
```
marceljk: Nice edge case, but the expected output of the utils function should never be nil. Instead an empty slice should be returned
```suggestion
			name: "pointer to nil slice",
			input: func() *[]string {
				var s []string
				return &s
			}(),
			expected:  []string{},
			expectNil: true,
```
To make this test case successful, you need to extend the utils function:
```go
func GetSliceFromPointer[T any](s *[]T) []T {
	if s == nil || *s == nil {
		return []T{}
	}
	return *s
}
```

With this you can probably also remove the attribute `expectNil`

internal/pkg/utils/utils_test.go
```diff
@@ -544,3 +544,53 @@ func TestBase64Bytes_MarshalYAML(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := GetSliceFromPointer(tt.input)
+
+			if result == nil && tt.expected == nil {
+				return
+			}
```
marceljk: The expected result should never be nil, because we always want to get an slice. So you can change the if clause to only check if the result == nil and if that's the case, return an error, e.g.:
```suggestion
			if result == nil {
			    t.Errorf("GetSliceFromPointer() = %v, want %v", result, tt.expected)
				return
			}
```

marceljk (approved):

PR #998, opened 2025-09-23
fix: yaml output of servers userdata is bytearray instead of base64

Benjosh95:
## Description

Problem:
The commands for server **list** and **describe** returns the userData as byteArray instead of base64 string.

Jira-Tickets: 
- https://jira.schwarz/browse/STACKITSDK-139
- https://jira.schwarz/browse/STACKITCLI-219

Testing instruction:
1. Create a server with userData ("test") via TF config looking something like this: 

```
            resource "stackit_server" "server-01" {
              project_id = var.project_id
              name       = "user-data-test-server"
              boot_volume = {
                source_type           = "image"
                size                  = 32
                source_id             = "a2c127b2-b1b5-4aee-986f-41cd11b41279" // Ubuntu 24.04 image
                delete_on_termination = true
              }
              machine_type = "t1.1"
              availability_zone  = "eu01-1"
              user_data = "test"
            `}`
```
              
2. Build the CLI for this branch
3. Run the two commands **_stackit server describe <SERVER-ID> -o yaml_** and  **_stackit server list -o yaml_** to see if userData is correctly output as base64

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/pkg/utils/utils.go
+++ b/internal/pkg/utils/utils.go
@@ -194,6 +194,9 @@ type Base64PatchedServer struct {
 }
 
 // ConvertToBase64PatchedServer converts an iaas.Server to Base64PatchedServer
+// This is a temporary workaround to get the desired base64 encoded yaml output for userdata
+// and will be replaced by a fix in the Go-SDK
+// ref: https://jira.schwarz/browse/STACKITSDK-246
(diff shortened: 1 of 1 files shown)


internal/pkg/utils/utils.go
marceljk, line 1: Can you add a comment that this is a temporary workaround until this is fixed in the Go SDK. Also add the ID of the follow up story to the comment

PR #1000, opened 2025-09-25
feat: add project validation (only for service SKE yet) 

Benjosh95:
## Description

This change implements validation checks for the project ID, project existence, and user access rights. It also simplifies the project label retrieval logic by reducing boilerplate code.

Jira-Ticket: https://jira.schwarz/browse/STACKITCLI-61

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/ske/describe/describe.go
+++ b/internal/cmd/ske/describe/describe.go
@@ -43,7 +43,7 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
 			}
 
 			// Validate project ID (exists and user has access)
-			_, err = validation.ValidateProject(ctx, params.Printer, params.CliVersion, cmd, model.ProjectId)
+			_, err = validation.ValidateAndGetProjectLabel(ctx, params.Printer, params.CliVersion, cmd, model.ProjectId)
 			if err != nil {
 				return err
 			}
--- a/internal/cmd/ske/describe/describe.go
+++ b/internal/cmd/ske/describe/describe.go
@@ -42,7 +42,6 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
 				return err
 			}
 
-			// Validate project ID (exists and user has access)
 			_, err = validation.ValidateAndGetProjectLabel(ctx, params.Printer, params.CliVersion, cmd, model.ProjectId)
 			if err != nil {
 				return err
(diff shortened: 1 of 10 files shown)


internal/pkg/validation/project.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/projectname"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/services/resourcemanager/client"
+
+	"github.com/spf13/cobra"
+	"github.com/stackitcloud/stackit-sdk-go/core/oapierror"
+)
+
+// ValidateProject validates that the project ID is not empty, exists, and the user has access to it.
+// It returns the project name for display purposes.
+func ValidateProject(ctx context.Context, p *print.Printer, cliVersion string, cmd *cobra.Command, projectId string) (string, error) {
```
rubenhoenle: Feels wrong that a func named `ValidateProject` returns a string for the project label. Choose a suitable name please.

internal/cmd/ske/describe/describe.go
```diff
@@ -41,6 +41,13 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			// Validate project ID (exists and user has access)
```
rubenhoenle: Didn't we agree in the refinement that we want to reduce boilerplate code within the implementation of this story? Instead we'll now have even more of it with your current approach.

internal/pkg/validation/project.go
```diff
@@ -0,0 +1,61 @@
...
+					projectLabel = projectName
+				}
+				return "", &errors.ProjectNotFoundError{ProjectId: projectId, ProjectLabel: projectLabel}
+			case http.StatusForbidden:
+				// Try to get project name for better error message
+				projectLabel := projectId
+				if projectName, nameErr := projectname.GetProjectName(ctx, p, cliVersion, cmd); nameErr == nil {
+					projectLabel = projectName
+				}
+				return "", &errors.ProjectAccessDeniedError{ProjectId: projectId, ProjectLabel: projectLabel}
```
marceljk: I would summarize this cases to one and always print the same error message, that either the project couldn't be found or permissions are missing.
The resource manager response in both cases with 403, because this prevents that some can find project Ids which are valid but he doesn't have the needed permissions.

PR #1016, opened 2025-10-08
refactor: use a func to print input model debug output

rubenhoenle:
## Description

I replaced some boilerplate code with a new printer func. 

Old: 

```golang
if p.IsVerbosityDebug() {
    modelStr, err := print.BuildDebugStrFromInputModel(model)
    if err != nil {
	    p.Debug(print.ErrorLevel, "convert model to string for debugging: %v", err)
    } else {
	    p.Debug(print.DebugLevel, "parsed input values: %s", modelStr)
    }
}
```

New: 
```golang
p.DebugInputModel(model)
```

New func is located here: `internal/pkg/print/print.go`

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -53,7 +53,7 @@ Please remember to run `make generate-docs` after your changes to keep the comma
 
 Below is a typical structure of a CLI command:
 
-https://github.com/stackitcloud/stackit-cli/blob/85ce44cd18d11169f2548d8657031b5fc6f94740/.github/docs/contribution-guide/cmd.go#L23-L156
+https://github.com/stackitcloud/stackit-cli/blob/main/.github/docs/contribution-guide/cmd.go
 
 Please remember to always add unit tests for `parseInput`, `buildRequest` (in `bar_test.go`), and any other util functions used.
 
@@ -87,7 +87,7 @@ If you want to add a command that uses a STACKIT service `foo` that was not yet 
     1.  This is done in `internal/pkg/services/foo/client/client.go`
     2.  Below is an example of a typical `client.go` file structure:
 
-https://github.com/stackitcloud/stackit-cli/blob/85ce44cd18d11169f2548d8657031b5fc6f94740/.github/docs/contribution-guide/client.go#L12-L35
+https://github.com/stackitcloud/stackit-cli/blob/main/.github/docs/contribution-guide/client.go
 
 ### Local development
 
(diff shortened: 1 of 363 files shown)


CONTRIBUTION.md
```diff
@@ -53,7 +53,7 @@ Please remember to run `make generate-docs` after your changes to keep the comma
 
 Below is a typical structure of a CLI command:
 
-https://github.com/stackitcloud/stackit-cli/blob/85ce44cd18d11169f2548d8657031b5fc6f94740/.github/docs/contribution-guide/cmd.go#L23-L156
+https://github.com/stackitcloud/stackit-cli/blob/main/.github/docs/contribution-guide/cmd.go
```
marceljk, line 56: This isn't rendered anymore

CONTRIBUTION.md
```diff
@@ -87,7 +87,7 @@ If you want to add a command that uses a STACKIT service `foo` that was not yet
     1.  This is done in `internal/pkg/services/foo/client/client.go`
     2.  Below is an example of a typical `client.go` file structure:
 
-https://github.com/stackitcloud/stackit-cli/blob/85ce44cd18d11169f2548d8657031b5fc6f94740/.github/docs/contribution-guide/client.go#L12-L35
+https://github.com/stackitcloud/stackit-cli/blob/main/.github/docs/contribution-guide/client.go
```
marceljk, line 90: And this isn't rendered anymore as well

CONTRIBUTION.md
rubenhoenle: I know, but with the current approach we would have to adjust the commit with every change to the contribution guide, don't know if that's a good idea

CONTRIBUTION.md
rubenhoenle: see my other comment

CONTRIBUTION.md
marceljk: From a user perspective, the current approach would be better readable, because all informations are on one page. But I see the issue, that we might could forget to update these code references. 
I'm open for both approaches, but we should than might keep it consistent. Here the old approach is still used: https://github.com/stackitcloud/stackit-cli/blob/72d190f93bb27e18fb2725136c3b148af8622037/CONTRIBUTION.md?plain=1#L64

rubenhoenle: Yes, that's intended. The link is pinned to a commit so the line reference will always be valid in the future. Else the line reference is worthless and the readers possibly won't find the correct place to register their new subcommand.

PR #1050, opened 2025-10-23
feat(logs): improve output for server logs

h3adex:
## Description

This PR improves the output of server logs to make it more human readable. The screenshot shows old and new output

Screenshot 2025-10-23 at 09 41 36

## Checklist
- [x] Issue was linked above

### DIFF
--- a/internal/cmd/server/log/log_test.go
+++ b/internal/cmd/server/log/log_test.go
@@ -184,9 +184,8 @@ func TestBuildRequest(t *testing.T) {
 
 func TestOutputResult(t *testing.T) {
 	type args struct {
-		outputFormat string
-		serverLabel  string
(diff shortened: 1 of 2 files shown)


internal/cmd/server/log/log_test.go
```diff
@@ -180,7 +180,7 @@ func TestOutputResult(t *testing.T) {
 	type args struct {
 		outputFormat string
 		serverLabel  string
-		log          string
+		logs         []string
```
rubenhoenle:
```suggestion
		logLines         []string
```

Please use the param names from the func you're testing if possible

marceljk:
When using before the output-format "pretty", the output is the same like with this change.

PR #1056, opened 2025-10-28
feat(iaas): Region adjustments and migration to v2 API

marceljk:
## Description

relates to STACKITCLI-227

## Checklist

- [x] Issue was linked above

### DIFF
--- a/docs/stackit_network-area.md
+++ b/docs/stackit_network-area.md
@@ -35,6 +35,7 @@ stackit network-area [flags]
 * [stackit network-area describe](./stackit_network-area_describe.md)	 - Shows details of a STACKIT Network Area
 * [stackit network-area list](./stackit_network-area_list.md)	 - Lists all STACKIT Network Areas (SNA) of an organization
 * [stackit network-area network-range](./stackit_network-area_network-range.md)	 - Provides functionality for network ranges in STACKIT Network Areas
+* [stackit network-area region](./stackit_network-area_region.md)	 - Provides functionality for regional configuration of STACKIT Network Area (SNA)
 * [stackit network-area route](./stackit_network-area_route.md)	 - Provides functionality for static routes in STACKIT Network Areas
 * [stackit network-area update](./stackit_network-area_update.md)	 - Updates a STACKIT Network Area (SNA)
 
--- a/docs/stackit_network-area_create.md
+++ b/docs/stackit_network-area_create.md
@@ -13,32 +13,20 @@ stackit network-area create [flags]
 ### Examples
 
 ```
-  Create a network area with name "network-area-1" in organization with ID "xxx" with network ranges and a transfer network
-  $ stackit network-area create --name network-area-1 --organization-id xxx --network-ranges "1.1.1.0/24,192.123.1.0/24" --transfer-network "192.160.0.0/24"
+  Create a network area with name "network-area-1" in organization with ID "xxx"
+  $ stackit network-area create --name network-area-1 --organization-id xxx"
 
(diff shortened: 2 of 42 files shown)


internal/cmd/network/describe/describe.go
```diff
@@ -88,44 +88,60 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
-		}
-
-		var ipv6nameservers []string
-		if network.NameserversV6 != nil {
-			ipv6nameservers = append(ipv6nameservers, *network.NameserversV6...)
+		// IPv4
+		var ipv4Nameservers []string
+		var ipv4Prefixes []string
+		var publicIp *string
+		var ipv4Gateway *string
```
Fyusel: NItpick: Can be written in two lines

internal/cmd/network/describe/describe.go
```diff
@@ -88,44 +88,60 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
+			}
 		}
 
-		var ipv6prefixes []string
-		if network.PrefixesV6 != nil {
-			ipv6prefixes = append(ipv6prefixes, *network.PrefixesV6...)
+		// IPv6
+		var ipv6Nameservers []string
+		var ipv6Prefixes []string
+		var ipv6Gateway *string
```
Fyusel, line 119: Same here

internal/cmd/network/list/list.go
```diff
@@ -143,18 +143,22 @@ func outputResult(p *print.Printer, outputFormat string, networks []iaas.Network
 		table.SetHeader("ID", "NAME", "STATUS", "PUBLIC IP", "PREFIXES", "ROUTED")
 
 		for _, network := range networks {
-			publicIp := utils.PtrString(network.PublicIp)
+			var publicIp string
+			var prefixes string
```
Fyusel: Same here

internal/cmd/server/create/create_test.go
```diff
@@ -168,7 +173,6 @@ func TestParseInput(t *testing.T) {
 				delete(flagValues, bootVolumePerformanceClassFlag)
 				delete(flagValues, bootVolumeDeleteOnTerminationFlag)
 				delete(flagValues, keypairNameFlag)
-				delete(flagValues, networkIdFlag)
```
Fyusel, line 171: In this PR are some changes which are not belonging to the region adjustment. I am not sure if those are changed in the other PRs or are made additionally here?

marceljk: The iaas v2 api doesn't only introduced the region adjustment, but also some generic api changes. E.g. for the server it's now required to set either a network ID or network interface ID. This was an issue in the past, that it was possible to create a server without being connected to a network, which results in some Support Tickets for the iaas team. So they set it now to required.

In the CLI, I set it as well as required one of:
https://github.com/stackitcloud/stackit-cli/blob/225d9e4d90d200b31d1b1ff531188d095a288ad1/internal/cmd/server/create/create.go#L187-L188

Setting it as required breaks tests of the server create command and I had to adjust this.

While adjusting all tests, I also changed some `var string` to `const string`, if it was possible, to prevent further changes. And I also replaced, that an additional const is created for `globalflags.ProjectIdFlag` and `globalflags.RegionFlag` and  used the const itself.

Fyusel: Ok, got it. I thought that those changes were made in the other PRs and this only includes the "generic" adjustment regarding the region parameter which are needed in the commands :+1:

Fyusel (dismissed):

PR #1057, opened 2025-10-28
feat(iaas): Region adjustment and v2 API migration for network-area commands

marceljk:
## Description

relates to STACKITCLI-227

## Checklist

- [x] Issue was linked above

### DIFF
--- a/docs/stackit_network-area.md
+++ b/docs/stackit_network-area.md
@@ -35,6 +35,7 @@ stackit network-area [flags]
 * [stackit network-area describe](./stackit_network-area_describe.md)	 - Shows details of a STACKIT Network Area
 * [stackit network-area list](./stackit_network-area_list.md)	 - Lists all STACKIT Network Areas (SNA) of an organization
 * [stackit network-area network-range](./stackit_network-area_network-range.md)	 - Provides functionality for network ranges in STACKIT Network Areas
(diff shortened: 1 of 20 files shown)


internal/cmd/network-area/route/describe/describe.go
```diff
@@ -100,18 +100,44 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
-		table.AddRow("NEXTHOP", utils.PtrString(route.Nexthop))
+		if destination := route.Destination; destination != nil {
+			if destination.DestinationCIDRv4 != nil {
+				table.AddRow("DESTINATION", utils.PtrString(destination.DestinationCIDRv4.Value))
+				table.AddSeparator()
+			}
+			if destination.DestinationCIDRv6 != nil {
+				table.AddRow("DESTINATION", utils.PtrString(destination.DestinationCIDRv6.Value))
+				table.AddSeparator()
+			}
```
Fyusel, line 123: As far as I can see it is only possible that one is set at a time?

If I am wrong I would change at least the DESTINATION description to differentiate between those two.

marceljk: Yes, always only one of both will be set, depending of the type (see [here](https://internal-docs.api.qa.stackit.cloud/documentation/iaas/version/v2#tag/Area/operation/v2GetRouteOfArea) the api docs).

I'll change it to an `else if`, so that it's clear in the code, that only one of both will be used

PR #1058, opened 2025-10-28
chore(client): Refactor ConfigureClient() function to use a generic f…

Fyusel:
## Description

The ConfigureClient() method was implemented for all clients with almost the same code. In order to reduce this code duplication a generic method was defined. The ConfigureClient() method is still present but is only a wrapper which means that the normal code does not need to be changed at all. This generic function still uses the region as parameter for some clients. This will be adjusted when all services have implemented the region adjustment.

For testing just execute the list command for every service (except the git service).

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/.github/docs/contribution-guide/client.go
+++ b/.github/docs/contribution-guide/client.go
@@ -2,46 +2,12 @@ package client
 
 import (
 	"github.com/spf13/viper"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/auth"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/config"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
+	genericclient "github.com/stackitcloud/stackit-cli/internal/pkg/generic-client"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
-	sdkConfig "github.com/stackitcloud/stackit-sdk-go/core/config"
 	"github.com/stackitcloud/stackit-sdk-go/services/foo"
 	// (...)
 )
 
 func ConfigureClient(p *print.Printer, cliVersion string) (*foo.APIClient, error) {
-	authCfgOption, err := auth.AuthenticationConfig(p, auth.AuthorizeUser)
-	if err != nil {
-		return nil, &errors.AuthError{}
-	}
-
-	region := viper.GetString(config.RegionKey)
-	cfgOptions := []sdkConfig.ConfigurationOption{
-		utils.UserAgentConfigOption(cliVersion),
-		sdkConfig.WithRegion(region), // Configuring region is needed if "foo" is a regional API
-		authCfgOption,
-	}
-
-	customEndpoint := viper.GetString(config.fooCustomEndpointKey)
-
-	if customEndpoint != "" {
-		cfgOptions = append(cfgOptions, sdkConfig.WithEndpoint(customEndpoint))
-	}
-
-	if p.IsVerbosityDebug() {
-		cfgOptions = append(cfgOptions,
-			sdkConfig.WithMiddleware(print.RequestResponseCapturer(p, nil)),
-		)
-	}
-
-	apiClient, err := foo.NewAPIClient(cfgOptions...)
-	if err != nil {
-		p.Debug(print.ErrorLevel, "create new API client: %v", err)
-		return nil, &errors.AuthError{}
-	}
-
-	return apiClient, nil
+	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.fooCustomEndpointKey), false, genericclient.CreateApiClient[*foo.APIClient](foo.NewAPIClient))
 }
--- a/internal/pkg/generic-client/generic_client.go
+++ b/internal/pkg/generic-client/generic_client.go
@@ -1,0 +1,50 @@
+package genericclient
+
+import (
+	"github.com/spf13/viper"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/auth"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/config"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	sdkConfig "github.com/stackitcloud/stackit-sdk-go/core/config"
+)
+
+type CreateApiClient[T any] func(opts ...sdkConfig.ConfigurationOption) (T, error)
+
+// ConfigureClientGeneric contains the generic code which needs to be executed in order to configure the api client.
+func ConfigureClientGeneric[T any](p *print.Printer, cliVersion, customEndpoint string, useRegion bool, createApiClient CreateApiClient[T]) (T, error) {
+	// return value if an error happens
+	var zero T
+	authCfgOption, err := auth.AuthenticationConfig(p, auth.AuthorizeUser)
+	if err != nil {
+		p.Debug(print.ErrorLevel, "configure authentication: %v", err)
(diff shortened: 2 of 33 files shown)


internal/pkg/generic-client/GenericClient.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	sdkConfig "github.com/stackitcloud/stackit-sdk-go/core/config"
+)
+
+type CreateApiClient[T any] func(opts ...sdkConfig.ConfigurationOption) (T, error)
+
+// ConfigureClientGeneric contains the generic code which needs to be executed in order to configure the api client.
+// TODO: region parameter will be removed when every API implemented the new region
```
marceljk: I think there will be always some service which won't have the regional requests, because they provide only a global api, e.g. the cost service or resourcemanager.

marceljk (approved):

.github/docs/contribution-guide/client.go
```diff
@@ -2,46 +2,13 @@ package client
...
-	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
+	genericclient "github.com/stackitcloud/stackit-cli/internal/pkg/generic-client"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
-	sdkConfig "github.com/stackitcloud/stackit-sdk-go/core/config"
 	"github.com/stackitcloud/stackit-sdk-go/services/foo"
 	// (...)
 )
 
+// TODO: region parameter will be removed when every service implemented the region adjustment
```
marceljk: I think this shouldn't be part of the contribution guide

internal/pkg/generic-client/GenericClient.go
```diff
@@ -0,0 +1,49 @@
...
+	cfgOptions := []sdkConfig.ConfigurationOption{
+		utils.UserAgentConfigOption(cliVersion),
+		authCfgOption,
+	}
+
+	if customEndpoint != "" {
+		cfgOptions = append(cfgOptions, sdkConfig.WithEndpoint(customEndpoint))
+	}
+
+	// TODO: this will be removed when every API implemented the new region
```
marceljk: Here is one more TODO leftover

Fyusel: removed

internal/pkg/services/kms/client/client.go
```diff
@@ -1,45 +1,14 @@
...
-	}
-
-	apiClient, err := kms.NewAPIClient(cfgOptions...)
-	if err != nil {
-		p.Debug(print.ErrorLevel, "create new API client: %v", err)
-		return nil, &errors.AuthError{}
-	}
-
-	return apiClient, nil
+	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.KMSCustomEndpointKey), viper.GetString(config.RegionKey), genericclient.CreateApiClient[*kms.APIClient](kms.NewAPIClient))
```
marceljk: The previous client configuration of kms, didn't used the region
```suggestion
	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.KMSCustomEndpointKey), "", genericclient.CreateApiClient[*kms.APIClient](kms.NewAPIClient))
```

internal/pkg/services/load-balancer/client/client.go
```diff
@@ -1,44 +1,14 @@
...
-	}
-
-	apiClient, err := loadbalancer.NewAPIClient(cfgOptions...)
-	if err != nil {
-		p.Debug(print.ErrorLevel, "create new API client: %v", err)
-		return nil, &errors.AuthError{}
-	}
-
-	return apiClient, nil
+	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.LoadBalancerCustomEndpointKey), viper.GetString(config.RegionKey), genericclient.CreateApiClient[*loadbalancer.APIClient](loadbalancer.NewAPIClient))
```
marceljk: Same here
```suggestion
	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.LoadBalancerCustomEndpointKey), "", genericclient.CreateApiClient[*loadbalancer.APIClient](loadbalancer.NewAPIClient))
```

internal/pkg/services/runcommand/client/client.go
```diff
@@ -1,46 +1,14 @@
...
-	}
-
-	apiClient, err := runcommand.NewAPIClient(cfgOptions...)
-	if err != nil {
-		p.Debug(print.ErrorLevel, "create new API client: %v", err)
-		return nil, &errors.AuthError{}
-	}
-
-	return apiClient, nil
+	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.RunCommandCustomEndpointKey), viper.GetString(config.RegionKey), genericclient.CreateApiClient[*runcommand.APIClient](runcommand.NewAPIClient))
```
marceljk: here

internal/pkg/services/serverosupdate/client/client.go
```diff
@@ -1,46 +1,14 @@
...
-	}
-
-	apiClient, err := serverupdate.NewAPIClient(cfgOptions...)
-	if err != nil {
-		p.Debug(print.ErrorLevel, "create new API client: %v", err)
-		return nil, &errors.AuthError{}
-	}
-
-	return apiClient, nil
+	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.ServerOsUpdateCustomEndpointKey), viper.GetString(config.RegionKey), genericclient.CreateApiClient[*serverupdate.APIClient](serverupdate.NewAPIClient))
```
marceljk: here

internal/pkg/services/sqlserverflex/client/client.go
```diff
@@ -1,45 +1,14 @@
...
-	}
-
-	apiClient, err := sqlserverflex.NewAPIClient(cfgOptions...)
-	if err != nil {
-		p.Debug(print.ErrorLevel, "create new API client: %v", err)
-		return nil, &errors.AuthError{}
-	}
-
-	return apiClient, nil
+	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.SQLServerFlexCustomEndpointKey), viper.GetString(config.RegionKey), genericclient.CreateApiClient[*sqlserverflex.APIClient](sqlserverflex.NewAPIClient))
```
marceljk: here

.github/docs/contribution-guide/client.go
```diff
@@ -2,46 +2,12 @@ package client
...
-	}
-
-	apiClient, err := foo.NewAPIClient(cfgOptions...)
-	if err != nil {
-		p.Debug(print.ErrorLevel, "create new API client: %v", err)
-		return nil, &errors.AuthError{}
-	}
-
-	return apiClient, nil
+	return genericclient.ConfigureClientGeneric(p, cliVersion, viper.GetString(config.fooCustomEndpointKey), "", genericclient.CreateApiClient[*foo.APIClient](foo.NewAPIClient))
```
rubenhoenle: Feels wrong to me to pass empty strings here...

Fyusel: Will change it to boolean

internal/pkg/generic-client/generic_client.go
marceljk, line 1: I think snake case is more widely used in go for naming go files than CamelCase. We also use it e.g. in "service_account.go" or "user_login.go". https://github.com/stackitcloud/stackit-cli/tree/cfd42a05e19069bbb067e2a9ce1cce0679e9a603/internal/pkg/auth

So I would suggest that we rename `GenericClient.go` to `generic_client.go`

PR #1063, opened 2025-10-29
feat(iaas): onboard network-area region commands

marceljk:
## Description

relates to STACKITCLI-227

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/network-area/region/create/create_test.go
+++ b/internal/cmd/network-area/region/create/create_test.go
@@ -1,0 +1,307 @@
+package create
+
+import (
+	"context"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/uuid"
+	"github.com/stackitcloud/stackit-cli/internal/cmd/params"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/testutils"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
+)
+
+const (
+	testRegion                    = "eu01"
+	testDefaultPrefixLength int64 = 25
+	testMaxPrefixLength     int64 = 29
+	testMinPrefixLength     int64 = 24
+	testTransferNetwork           = "192.168.2.0/24"
+)
+
+type testCtxKey struct{}
+
+var testCtx = context.WithValue(context.Background(), testCtxKey{}, "foo")
+var testClient = &iaas.APIClient{}
+
+var (
+	testAreaId             = uuid.NewString()
+	testOrgId              = uuid.NewString()
+	testDefaultNameservers = []string{"8.8.8.8", "8.8.4.4"}
+	testNetworkRanges      = []string{"192.168.0.0/24", "10.0.0.0/24"}
+)
+
+func fixtureFlagValues(mods ...func(flagValues map[string]string)) map[string]string {
+	flagValues := map[string]string{
+		globalflags.RegionFlag: testRegion,
+
+		networkAreaIdFlag:           testAreaId,
+		organizationIdFlag:          testOrgId,
+		ipv4DefaultNameservers:      strings.Join(testDefaultNameservers, ","),
+		ipv4DefaultPrefixLengthFlag: strconv.FormatInt(testDefaultPrefixLength, 10),
+		ipv4MaxPrefixLengthFlag:     strconv.FormatInt(testMaxPrefixLength, 10),
+		ipv4MinPrefixLengthFlag:     strconv.FormatInt(testMinPrefixLength, 10),
+		ipv4NetworkRangesFlag:       strings.Join(testNetworkRanges, ","),
+		ipv4TransferNetworkFlag:     testTransferNetwork,
+	}
+	for _, mod := range mods {
+		mod(flagValues)
+	}
+	return flagValues
+}
+
+func fixtureInputModel(mods ...func(model *inputModel)) *inputModel {
+	model := &inputModel{
+		GlobalFlagModel: &globalflags.GlobalFlagModel{
+			Region:    testRegion,
+			Verbosity: globalflags.VerbosityDefault,
+		},
+		OrganizationId:          testOrgId,
+		NetworkAreaId:           testAreaId,
+		IPv4DefaultNameservers:  utils.Ptr(testDefaultNameservers),
+		IPv4DefaultPrefixLength: utils.Ptr(testDefaultPrefixLength),
+		IPv4MaxPrefixLength:     utils.Ptr(testMaxPrefixLength),
+		IPv4MinPrefixLength:     utils.Ptr(testMinPrefixLength),
+		IPv4NetworkRanges:       testNetworkRanges,
+		IPv4TransferNetwork:     testTransferNetwork,
+	}
+	for _, mod := range mods {
+		mod(model)
+	}
+	return model
+}
+
+func fixtureRequest(mods ...func(request *iaas.ApiCreateNetworkAreaRegionRequest)) iaas.ApiCreateNetworkAreaRegionRequest {
+	request := testClient.CreateNetworkAreaRegion(testCtx, testOrgId, testAreaId, testRegion)
+	request = request.CreateNetworkAreaRegionPayload(fixturePayload())
+	for _, mod := range mods {
+		mod(&request)
+	}
+	return request
+}
+
+func fixturePayload(mods ...func(payload *iaas.CreateNetworkAreaRegionPayload)) iaas.CreateNetworkAreaRegionPayload {
+	var networkRange []iaas.NetworkRange
+	if len(testNetworkRanges) > 0 {
+		networkRange = make([]iaas.NetworkRange, len(testNetworkRanges))
+		for i := range testNetworkRanges {
+			networkRange[i] = iaas.NetworkRange{
+				Prefix: utils.Ptr(testNetworkRanges[i]),
+			}
+		}
+	}
+
+	payload := iaas.CreateNetworkAreaRegionPayload{
+		Ipv4: &iaas.RegionalAreaIPv4{
+			DefaultNameservers: utils.Ptr(testDefaultNameservers),
+			DefaultPrefixLen:   utils.Ptr(testDefaultPrefixLength),
+			MaxPrefixLen:       utils.Ptr(testMaxPrefixLength),
+			MinPrefixLen:       utils.Ptr(testMinPrefixLength),
+			NetworkRanges:      utils.Ptr(networkRange),
+			TransferNetwork:    utils.Ptr(testTransferNetwork),
+		},
+	}
+	for _, mod := range mods {
+		mod(&payload)
+	}
+	return payload
+}
+
+func TestParseInput(t *testing.T) {
+	tests := []struct {
+		description   string
+		argValues     []string
+		flagValues    map[string]string
+		isValid       bool
+		expectedModel *inputModel
+	}{
+		{
+			description:   "base",
+			flagValues:    fixtureFlagValues(),
+			isValid:       true,
+			expectedModel: fixtureInputModel(),
+		},
+		{
+			description: "no values",
+			flagValues:  map[string]string{},
+			isValid:     false,
+		},
+		{
+			description: "area id missing",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				delete(flagValues, networkAreaIdFlag)
+			}),
+			isValid: false,
+		},
+		{
+			description: "area id invalid 1",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[networkAreaIdFlag] = ""
+			}),
+			isValid: false,
+		},
+		{
+			description: "area id invalid 2",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[networkAreaIdFlag] = "invalid-uuid"
+			}),
+			isValid: false,
+		},
+		{
+			description: "org id missing",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				delete(flagValues, organizationIdFlag)
+			}),
+			isValid: false,
+		},
+		{
+			description: "org id invalid 1",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[organizationIdFlag] = ""
+			}),
+			isValid: false,
+		},
+		{
+			description: "org id invalid 2",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[organizationIdFlag] = "invalid-uuid"
+			}),
+			isValid: false,
+		},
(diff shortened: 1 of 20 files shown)


internal/cmd/network-area/region/create/create_test.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/cmd/params"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/testutils"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
+)
+
+const (
+	regionFlag = globalflags.RegionFlag
```
rubenhoenle: What's the point of duplicating one constant into another one? Why not just use `globalflags.RegionFlag` directly? :smile:

internal/cmd/network-area/region/delete/delete_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/uuid"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/testutils"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
+)
+
+const (
+	regionFlag = globalflags.RegionFlag
```
rubenhoenle: same here, I gues you're doing it in multiple tests :sweat_smile:

internal/cmd/network-area/region/list/list.go
```diff
@@ -0,0 +1,154 @@
...
+func NewCmd(params *params.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "Lists all configured regions for a STACKIT Network Area (SNA)",
+		Long:  "Lists all configured regions for a STACKIT Network Area (SNA).",
+		Args:  args.SingleArg(networkAreaArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(
+				`List all configured region for a STACKIT Network Area with ID "xxx" in organization with ID "yyy"`,
+				`$ stackit network-area region list xxx  --organization-id yyy`,
```
rubenhoenle:
```suggestion
				`$ stackit network-area region list xxx --organization-id yyy`,
```

internal/cmd/network-area/region/create/create_test.go
marceljk: removed the regionFlag and just used `globalflags.RegionFlag` now. I also adjusted it in all other iaas commands in the separated PRs 😄

internal/cmd/network-area/region/delete/delete_test.go
marceljk: updated

internal/cmd/network-area/create/create.go
```diff
@@ -109,8 +127,38 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
...
+					return fmt.Errorf("create network area: empty response")
+				}
+				reqNetworkArea := buildRequestNetworkAreaRegion(ctx, model, *resp.Id, apiClient)
+				respNetworkArea, err := reqNetworkArea.Execute()
+				if err != nil {
+					return fmt.Errorf("create network area region: %w", err)
+				}
+				if !model.AssumeYes {
+					s := spinner.New(params.Printer)
+					s.Start("Create network area")
```
Fyusel:
```suggestion
					s.Start("Create network area region")
```

internal/cmd/network-area/create/create.go
```diff
@@ -109,8 +127,38 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
...
+					_, err = wait.CreateNetworkAreaRegionWaitHandler(ctx, apiClient, model.OrganizationId, *resp.Id, model.Region).WaitWithContext(ctx)
+					if err != nil {
+						return fmt.Errorf("wait for completing SKE credentials rotation %w", err)
```
Fyusel: copy past error "ske credentials rotation"

internal/cmd/network-area/create/create.go
```diff
@@ -148,44 +235,70 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel,
 		Labels:              flags.FlagToStringToStringPointer(p, cmd, labelFlag),
 	}
 
+	hasAllRequiredRegionalAreaFieldsSet := model.NetworkRanges != nil || model.TransferNetwork != nil
```
Fyusel: This is only present for the testing, right? Since the _cmd.MarkFlagsRequiredTogether(networkRangesFlag, transferNetworkFlag)_ prevents to get here.

Nevertheless I think there is something wrong. The unit tests are currently passing but the test cases are wrong in my opinion e.g.
```
description: "deprecated network ranges missing",
				delete(flagValues, networkRangesFlag)
			}),
			isValid: true,
			expectedModel: fixtureInputModel(func(model *inputModel) {
				model.NetworkRanges = nil
			}),
```
By looking at the fixtureFlagValues() the only fields which are set are: name, organization and labels so the removal of _delete(flagValues, networkRangesFlag)_ does nothing. When debugging this test case you can see that the networkRanges and also the transferNetwork is nil. I think we should add tests for having both set and removing one in order to see that we run into this issue here.

internal/cmd/network-area/create/create_test.go
```diff
@@ -142,18 +181,24 @@ func TestParseInput(t *testing.T) {
 			isValid: false,
 		},
 		{
-			description: "network ranges missing",
+			description: "deprecated network ranges missing",
 			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
 				delete(flagValues, networkRangesFlag)
```
Fyusel: This does nothing since networkRangesFlag is not set.

internal/cmd/network-area/create/create_test.go
```diff
@@ -142,18 +181,24 @@ func TestParseInput(t *testing.T) {
...
+			isValid: true,
+			expectedModel: fixtureInputModel(func(model *inputModel) {
+				model.NetworkRanges = nil
+			}),
 		},
 		{
-			description: "transfer network missing",
+			description: "deprecated transfer network missing",
 			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
 				delete(flagValues, transferNetworkFlag)
```
Fyusel: Same here

internal/cmd/network-area/region/list/list.go
```diff
@@ -0,0 +1,153 @@
...
+			if err != nil {
+				params.Printer.Debug(print.ErrorLevel, "get network area name: %v", err)
+				networkAreaLabel = model.NetworkAreaId
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("create network range: %w", err)
```
Fyusel: wrong error message

internal/cmd/network-area/region/update/update.go
```diff
@@ -0,0 +1,165 @@
...
+				if err != nil {
+					return err
+				}
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("create network area region: %w", err)
```
Fyusel: wrong error message

internal/cmd/network-area/route/create/create.go
```diff
@@ -93,7 +116,27 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
...
-			route, err := iaasUtils.GetRouteFromAPIResponse(*model.Prefix, *model.Nexthop, resp.Items)
+			var destination string
+			var nexthop string
+			if model.DestinationV4 != nil {
+				destination = *model.DestinationV4
+			} else if model.DestinationV6 != nil {
+				destination = *model.DestinationV6
+			}
+
+			if model.NexthopV4 != nil {
```
Fyusel, line 127: Where to get the prioritization of those? If there is a prio this should be documented (also in the --help of the command).

Fyusel: Ah, I see some lines below those flags are marked exclusively so a customer cannot set more than one -> I am fine with that.

internal/cmd/network-area/route/create/create.go
```diff
@@ -109,38 +152,146 @@ func configureFlags(cmd *cobra.Command) {
...
-				Labels:  utils.ConvertStringMapToInterfaceMap(model.Labels),
+				Destination: &iaas.RouteDestination{
+					DestinationCIDRv4: destinationV4,
+					DestinationCIDRv6: destinationV6,
+				},
+				Nexthop: &iaas.RouteNexthop{
+					NexthopIPv4:      nexthopIPv4,
+					NexthopIPv6:      nexthopIPv6,
+					NexthopBlackhole: nexthopBlackhole,
+					NexthopInternet:  nexthopInternet,
```
Fyusel, line 292: Eh, you need to provide all of them even if only one is allowed to set :-(

internal/cmd/network-area/update/update.go
```diff
@@ -99,7 +117,29 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
 				return fmt.Errorf("update network area: %w", err)
 			}
 
-			return outputResult(params.Printer, model.OutputFormat, orgLabel, *resp)
+			if resp == nil || resp.Id == nil {
+				return fmt.Errorf("create network area: empty response")
```
Fyusel: wrong error message

internal/cmd/network-area/update/update.go
```diff
@@ -99,7 +117,29 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
...
+			}
+
+			responses := NetworkAreaResponses{
+				NetworkArea: *resp,
+			}
+
+			if hasDeprecatedFlagsSet(model) {
+				deprecatedFlags := getConfiguredDeprecatedFlags(model)
+				params.Printer.Warn("the flags %q are deprecated and will be removed after April 2026. Use `$ stackit network-area region` to configure these options for a network area.\n", strings.Join(deprecatedFlags, ","))
+				if resp == nil || resp.Id == nil {
```
Fyusel: Is this needed here? The same check is some lines above

internal/cmd/network-area/create/create.go
marceljk: Good catch! I had it in the beginning a bit different and the fixtureFlagValues and fixtureInputModel had the deprecated fields set. I removed it afterwards, because the default values of them, shouldn't contain any deprecated flags and I missed then to adjust the tests as well. I'll adjust it, so that it will be tested correctly.

The check you mentioned first in your comment isn't only for testing. The name of the bool `hasAllRequiredRegionalAreaFieldsSet` may doesn't fit 100% and I will rename it. What is should actual check here is, if any of the deprecated fields is set, which are **optional** for the request, and if no one of the **required** fields for the regional area is set, then it should throw an error, because the request can not be executed. I'll also add a small comment in the code to explain this

internal/cmd/network-area/create/create_test.go
marceljk: updated

internal/cmd/network-area/create/create_test.go
marceljk: updated

internal/cmd/network-area/route/create/create.go
marceljk: all nexthop types are initialized as `var nexthopIPv4 *iaas.NexthopIPv4` so they are all nil in the beginning. If they were intialized as `nexthopIPv4 := &iaas.NexthopIPv4{}`, they would be an actual instance of the struct and assigning them directly to `iaas.RouteNexthop` would result in unexpected behavior. Then I would need to do some additional checks, to set the correct nexthop type

With the current approach, the nexthop in the payload could be also set with additional if checks like:
```go
	nexthop := &iaas.RouteNexthop{}
	if nexthopIPv4 != nil {
		nexthop.NexthopIPv4 = nexthopIPv4
	} else if nexthopIPv6 != nil {
		nexthop.NexthopIPv6 = nexthopIPv6
	} else if nexthopBlackhole != nil {
		nexthop.NexthopBlackhole = nexthopBlackhole
	} else if nexthopInternet != nil {
		nexthop.NexthopInternet = nexthopInternet
	}
```
But the outcome would be the same and I think the current solution is cleaner and requires less code

internal/cmd/network-area/update/update.go
marceljk: No it's not needed. Removed it

internal/cmd/network-area/create/create.go
marceljk: Update: I left the name of the bool like it was, because there is already this check `cmd.MarkFlagsRequiredTogether(networkRangesFlag, transferNetworkFlag)` as mentioned by you and instead I updated the condition to `model.NetworkRanges != nil && model.TransferNetwork != nil`. Added also a small comment to this check.
I also extended the unit tests with a few more cases like, only one of the required deprecated fields is set, both are set, only the optional deprecated flags are set.

internal/cmd/network-area/region/delete/delete.go
```diff
@@ -0,0 +1,129 @@
...
+				if err != nil {
+					return err
+				}
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("create network range: %w", err)
```
Fyusel: wrong error message

internal/cmd/network-area/region/delete/delete.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			if !model.Async {
+				s := spinner.New(params.Printer)
+				s.Start("Delete network area region")
+				_, err = wait.DeleteNetworkAreaRegionWaitHandler(ctx, apiClient, model.OrganizationId, model.NetworkAreaId, model.Region).WaitWithContext(ctx)
+				if err != nil {
+					return fmt.Errorf("wait for network area deletion: %w", err)
```
Fyusel: wrong error message

internal/cmd/network-area/region/describe/describe.go
```diff
@@ -0,0 +1,169 @@
...
+				params.Printer.Debug(print.ErrorLevel, "get network area name: %v", err)
+				// Set explicit the networkAreaName to empty string and not to the ID, because this is used for the table output
+				networkAreaName = ""
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("create network range: %w", err)
```
Fyusel: wrong error message

internal/cmd/network-area/region/list/list.go
```diff
@@ -0,0 +1,153 @@
...
+			if err != nil {
+				params.Printer.Debug(print.ErrorLevel, "get network area name: %v", err)
+				networkAreaLabel = model.NetworkAreaId
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("list network range: %w", err)
```
Fyusel: wrong error message

PR #1094, opened 2025-11-19
feat(config): add ignore-existing for profile create

dergeberl:
## Description

This PR adds `--ignore-existing`  as a flag for `stackit config profile create` in order to call the create idempotent and don not get a error in case the profile already exists. That make it possible to use profiles in automation/tool. 

Example: I am currently integrating stackit(-cli) into the [gardenctl](https://github.com/gardener/gardenctl-v2) (https://github.com/gardener/gardenctl-v2/pull/669). 

## Checklist

- [ ] Issue was linked above


internal/cmd/config/profile/create/create.go
```diff
@@ -85,6 +88,7 @@ func NewCmd(params *params.CmdParams) *cobra.Command {
 
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(noSetFlag, false, "Do not set the profile as the active profile")
+	cmd.Flags().Bool(ignoreExistingFlag, false, "Suppress the error it the profile exists already")
```
cgoetz-inovex: typo
```suggestion
	cmd.Flags().Bool(ignoreExistingFlag, false, "Suppress the error if the profile exists already")
```

internal/pkg/config/profiles.go
```diff
@@ -80,7 +80,7 @@ func GetProfileFromEnv() (string, bool) {
 // If emptyProfile is true, it creates an empty profile. Otherwise, copies the config from the current profile to the new profile.
 // If setProfile is true, it sets the new profile as the active profile.
 // If the profile already exists, it returns an error.
```
cgoetz-inovex:
```suggestion
// If the profile already exists and ignoreExisting is false, it returns an error.
```

internal/pkg/config/profiles.go
```diff
@@ -98,6 +98,9 @@ func CreateProfile(p *print.Printer, profile string, setProfile, emptyProfile bo
 	// Error if the profile already exists
 	_, err = os.Stat(configFolderPath)
 	if err == nil {
+		if ignoreExising {
+			return nil
```
cgoetz-inovex, line 108:
```suggestion
			if setProfile {
			    err = SetProfile(p, profile)
			    if err != nil {
			        return fmt.Errorf("set profile: %w", err)
			    }
			}
			return nil
```
Currently the value of `setProfile` is ignored, when this branch is taken.

When creating a profile with `--ignore-existing`, that already exists but isn't active at the moment, the CLI reports:
> Successfully created and set active profile to "qa"
'qa' isn't activated in this case.

cgoetz-inovex (commented):
tested locally, lgtm

internal/pkg/config/profiles.go
```diff
@@ -79,8 +79,8 @@ func GetProfileFromEnv() (string, bool) {
 // CreateProfile creates a new profile.
 // If emptyProfile is true, it creates an empty profile. Otherwise, copies the config from the current profile to the new profile.
 // If setProfile is true, it sets the new profile as the active profile.
-// If the profile already exists, it returns an error.
-func CreateProfile(p *print.Printer, profile string, setProfile, emptyProfile bool) error {
+// If the profile already exists and ignoreExisting is false, it returns an error.
+func CreateProfile(p *print.Printer, profile string, setProfile, ignoreExising, emptyProfile bool) error {
```
marceljk: nitpick: typo 
```suggestion
func CreateProfile(p *print.Printer, profile string, setProfile, ignoreExisting, emptyProfile bool) error {
```

internal/pkg/config/profiles.go
```diff
@@ -98,6 +98,15 @@ func CreateProfile(p *print.Printer, profile string, setProfile, emptyProfile bo
 	// Error if the profile already exists
 	_, err = os.Stat(configFolderPath)
 	if err == nil {
+		if ignoreExising {
```
marceljk: What should the flag ignoreExisting do? For me it's currently a bit confusing, because I assumed that it's like an overwrite of the existing profile with the same name. Usually when a user runs `stackit config profile create test`, it creates a copy of the current active profile, stores it in a new profile and sets the profile to active.

When I now set the flag `--ignore-existing`, the CLI just sets the existing profile to active, but doesn't copy the previous active profile. And in case the flag `--no-set` is set, the command would do nothing, just end without an error.

dergeberl: It should not throw an error when the profile already exists and not overwrite an existing profile. In our use-case we create a profile dynamically with the `gardenctl` for different environments. We do not want to overwrite the config as they need to set manually if that envs needs different API endpoints.

If the name `--ignore-existing` is confusing we maybe could to something like ``--ignore-existing-error`.

WDYT?

marceljk: Alright, I think I got it and the flag name should be fine. To make this clear for every user, can you please extend the description of the flag with the information, that when the flag is set, it does not modifies existing profiles.

marceljk (approved):

cgoetz-inovex (approved):

PR #1100, opened 2025-11-19
feat(cdn): add cdn client, config, list command

cgoetz-inovex:
## Description

Issue: STACKITCLI-70

## Checklist

- [x] Issue was linked above

### DIFF
--- a/docs/stackit_beta.md
+++ b/docs/stackit_beta.md
@@ -46,6 +46,7 @@ stackit beta [flags]
 * [stackit beta edge-cloud](./stackit_beta_edge-cloud.md)	 - Provides functionality for edge services.
 * [stackit beta intake](./stackit_beta_intake.md)	 - Provides functionality for intake
 * [stackit beta kms](./stackit_beta_kms.md)	 - Provides functionality for KMS
+* [stackit beta logs](./stackit_beta_logs.md)	 - Provides functionality for Logs
 * [stackit beta sfs](./stackit_beta_sfs.md)	 - Provides functionality for SFS (stackit file storage)
 * [stackit beta sqlserverflex](./stackit_beta_sqlserverflex.md)	 - Provides functionality for SQLServer Flex
 
--- a/docs/stackit_beta_logs.md
+++ b/docs/stackit_beta_logs.md
@@ -1,0 +1,34 @@
+## stackit beta logs
+
+Provides functionality for Logs
+
+### Synopsis
+
+Provides functionality for Logs.
+
+```
+stackit beta logs [flags]
+```
+
+### Options
+
+```
+  -h, --help   Help for "stackit beta logs"
+```
+
+### Options inherited from parent commands
+
+```
+  -y, --assume-yes             If set, skips all confirmation prompts
+      --async                  If set, runs the command asynchronously
+  -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
+  -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
+      --verbosity string       Verbosity of the CLI, one of ["debug" "info" "warning" "error"] (default "info")
+```
+
+### SEE ALSO
+
+* [stackit beta](./stackit_beta.md)	 - Contains beta STACKIT CLI commands
+* [stackit beta logs instance](./stackit_beta_logs_instance.md)	 - Provides functionality for Logs instances
+
--- a/docs/stackit_beta_logs_instance.md
+++ b/docs/stackit_beta_logs_instance.md
@@ -1,0 +1,38 @@
+## stackit beta logs instance
+
+Provides functionality for Logs instances
+
+### Synopsis
+
+Provides functionality for Logs instances.
+
+```
+stackit beta logs instance [flags]
+```
+
+### Options
+
+```
+  -h, --help   Help for "stackit beta logs instance"
+```
+
+### Options inherited from parent commands
+
+```
+  -y, --assume-yes             If set, skips all confirmation prompts
+      --async                  If set, runs the command asynchronously
+  -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
+  -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
+      --verbosity string       Verbosity of the CLI, one of ["debug" "info" "warning" "error"] (default "info")
+```
+
+### SEE ALSO
+
+* [stackit beta logs](./stackit_beta_logs.md)	 - Provides functionality for Logs
+* [stackit beta logs instance create](./stackit_beta_logs_instance_create.md)	 - Creates a Logs instance
+* [stackit beta logs instance delete](./stackit_beta_logs_instance_delete.md)	 - Deletes the given Logs instance
+* [stackit beta logs instance describe](./stackit_beta_logs_instance_describe.md)	 - Shows details of a Logs instance
+* [stackit beta logs instance list](./stackit_beta_logs_instance_list.md)	 - Lists Logs instances
+* [stackit beta logs instance update](./stackit_beta_logs_instance_update.md)	 - Updates a Logs instance
+
--- a/docs/stackit_beta_logs_instance_create.md
+++ b/docs/stackit_beta_logs_instance_create.md
@@ -1,0 +1,50 @@
+## stackit beta logs instance create
+
+Creates a Logs instance
+
+### Synopsis
+
+Creates a Logs instance.
+
+```
+stackit beta logs instance create [flags]
+```
+
+### Examples
+
+```
+  Create a Logs instance with name "my-instance" and retention time 10 days
+  $ stackit beta logs instance create --display-name "my-instance" --retention-days 10
+
+  Create a Logs instance with name "my-instance", retention time 10 days, and a description
+  $ stackit beta logs instance create --display-name "my-instance" --retention-days 10 --description "Description of the instance"
+
+  Create a Logs instance with name "my-instance", retention time 10 days, and restrict access to a specific range of IP addresses.
+  $ stackit beta logs instance create --display-name "my-instance" --retention-days 10 --acl 1.2.3.0/24
+```
+
+### Options
+
+```
+      --acl strings           Access control list
+      --description string    Description
+      --display-name string   Display name
+  -h, --help                  Help for "stackit beta logs instance create"
+      --retention-days int    The days for how long the logs should be stored before being cleaned up
+```
+
+### Options inherited from parent commands
+
+```
+  -y, --assume-yes             If set, skips all confirmation prompts
+      --async                  If set, runs the command asynchronously
+  -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
+  -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
+      --verbosity string       Verbosity of the CLI, one of ["debug" "info" "warning" "error"] (default "info")
+```
+
+### SEE ALSO
+
+* [stackit beta logs instance](./stackit_beta_logs_instance.md)	 - Provides functionality for Logs instances
+
--- a/docs/stackit_beta_logs_instance_delete.md
+++ b/docs/stackit_beta_logs_instance_delete.md
@@ -1,0 +1,40 @@
+## stackit beta logs instance delete
+
+Deletes the given Logs instance
+
+### Synopsis
+
+Deletes the given Logs instance.
+
+```
+stackit beta logs instance delete INSTANCE_ID [flags]
+```
+
+### Examples
+
+```
+  Delete a Logs instance with ID "xxx"
+  $ stackit beta logs instance delete "xxx"
+```
+
+### Options
+
+```
+  -h, --help   Help for "stackit beta logs instance delete"
+```
+
+### Options inherited from parent commands
+
+```
+  -y, --assume-yes             If set, skips all confirmation prompts
+      --async                  If set, runs the command asynchronously
+  -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
+  -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
+      --verbosity string       Verbosity of the CLI, one of ["debug" "info" "warning" "error"] (default "info")
+```
+
+### SEE ALSO
+
+* [stackit beta logs instance](./stackit_beta_logs_instance.md)	 - Provides functionality for Logs instances
+
--- a/docs/stackit_beta_logs_instance_describe.md
+++ b/docs/stackit_beta_logs_instance_describe.md
@@ -1,0 +1,43 @@
+## stackit beta logs instance describe
+
+Shows details of a Logs instance
+
+### Synopsis
+
+Shows details of a Logs instance
+
+```
+stackit beta logs instance describe INSTANCE_ID [flags]
+```
+
+### Examples
+
+```
+  Get details of a Logs instance with ID "xxx"
+  $ stackit beta logs instance describe xxx
+
+  Get details of a Logs instance with ID "xxx" in JSON format
+  $ stackit beta logs instance describe xxx --output-format json
+```
+
+### Options
+
+```
+  -h, --help   Help for "stackit beta logs instance describe"
+```
+
+### Options inherited from parent commands
+
+```
+  -y, --assume-yes             If set, skips all confirmation prompts
+      --async                  If set, runs the command asynchronously
+  -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
+  -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
+      --verbosity string       Verbosity of the CLI, one of ["debug" "info" "warning" "error"] (default "info")
+```
+
+### SEE ALSO
+
+* [stackit beta logs instance](./stackit_beta_logs_instance.md)	 - Provides functionality for Logs instances
+
--- a/docs/stackit_beta_logs_instance_list.md
+++ b/docs/stackit_beta_logs_instance_list.md
@@ -1,0 +1,44 @@
+## stackit beta logs instance list
+
+Lists Logs instances
+
+### Synopsis
+
+Lists Logs instances within the project.
+
+```
+stackit beta logs instance list [flags]
+```
+
+### Examples
+
+```
+  List all Logs instances
+  $ stackit beta logs instance list
+
+  List the first 10 Logs instances
+  $ stackit beta logs instance list --limit=10
+```
+
+### Options
+
+```
+  -h, --help        Help for "stackit beta logs instance list"
+      --limit int   Limit the output to the first n elements
+```
+
+### Options inherited from parent commands
+
+```
+  -y, --assume-yes             If set, skips all confirmation prompts
+      --async                  If set, runs the command asynchronously
+  -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
+  -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
+      --verbosity string       Verbosity of the CLI, one of ["debug" "info" "warning" "error"] (default "info")
+```
+
+### SEE ALSO
+
+* [stackit beta logs instance](./stackit_beta_logs_instance.md)	 - Provides functionality for Logs instances
+
--- a/docs/stackit_beta_logs_instance_update.md
+++ b/docs/stackit_beta_logs_instance_update.md
@@ -1,0 +1,50 @@
+## stackit beta logs instance update
+
+Updates a Logs instance
+
+### Synopsis
+
+Updates a Logs instance.
+
+```
+stackit beta logs instance update INSTANCE_ID [flags]
+```
+
+### Examples
+
+```
+  Update the display name of the Logs instance with ID "xxx"
+  $ stackit beta logs instance update xxx --display-name new-name
+
+  Update the retention time of the Logs instance with ID "xxx"
+  $ stackit beta logs instance update xxx --retention-days 40
+
+  Update the ACL of the Logs instance with ID "xxx"
+  $ stackit beta logs instance update xxx --acl 1.2.3.0/24
+```
+
+### Options
+
+```
+      --acl strings           Access control list
+      --description string    Description
+      --display-name string   Display name
+  -h, --help                  Help for "stackit beta logs instance update"
+      --retention-days int    The days for how long the logs should be stored before being cleaned up
+```
+
+### Options inherited from parent commands
+
+```
+  -y, --assume-yes             If set, skips all confirmation prompts
+      --async                  If set, runs the command asynchronously
+  -o, --output-format string   Output format, one of ["json" "pretty" "none" "yaml"]
+  -p, --project-id string      Project ID
+      --region string          Target region for region-specific requests
+      --verbosity string       Verbosity of the CLI, one of ["debug" "info" "warning" "error"] (default "info")
+```
+
+### SEE ALSO
+
+* [stackit beta logs instance](./stackit_beta_logs_instance.md)	 - Provides functionality for Logs instances
+
--- a/docs/stackit_config_set.md
+++ b/docs/stackit_config_set.md
@@ -42,6 +42,7 @@ stackit config set [flags]
       --kms-custom-endpoint string                                 KMS API base URL, used in calls to this API
       --load-balancer-custom-endpoint string                       Load Balancer API base URL, used in calls to this API
       --logme-custom-endpoint string                               LogMe API base URL, used in calls to this API
+      --logs-custom-endpoint string                                Logs API base URL, used in calls to this API
       --mariadb-custom-endpoint string                             MariaDB API base URL, used in calls to this API
       --mongodbflex-custom-endpoint string                         MongoDB Flex API base URL, used in calls to this API
       --object-storage-custom-endpoint string                      Object Storage API base URL, used in calls to this API
--- a/docs/stackit_config_unset.md
+++ b/docs/stackit_config_unset.md
@@ -40,6 +40,7 @@ stackit config unset [flags]
       --kms-custom-endpoint                                 KMS API base URL. If unset, uses the default base URL
       --load-balancer-custom-endpoint                       Load Balancer API base URL. If unset, uses the default base URL
       --logme-custom-endpoint                               LogMe API base URL. If unset, uses the default base URL
+      --logs-custom-endpoint                                Logs API base URL. If unset, uses the default base URL
       --mariadb-custom-endpoint                             MariaDB API base URL. If unset, uses the default base URL
       --mongodbflex-custom-endpoint                         MongoDB Flex API base URL. If unset, uses the default base URL
       --object-storage-custom-endpoint                      Object Storage API base URL. If unset, uses the default base URL
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/stackitcloud/stackit-sdk-go/services/git v0.10.1
 	github.com/stackitcloud/stackit-sdk-go/services/iaas v1.3.0
 	github.com/stackitcloud/stackit-sdk-go/services/intake v0.4.1
+	github.com/stackitcloud/stackit-sdk-go/services/logs v0.4.0
 	github.com/stackitcloud/stackit-sdk-go/services/mongodbflex v1.5.5
 	github.com/stackitcloud/stackit-sdk-go/services/opensearch v0.24.3
 	github.com/stackitcloud/stackit-sdk-go/services/postgresflex v1.3.2
--- a/go.sum
+++ b/go.sum
@@ -624,6 +624,10 @@ github.com/stackitcloud/stackit-sdk-go/services/loadbalancer v1.7.0 h1:ZyaB4jL71
 github.com/stackitcloud/stackit-sdk-go/services/loadbalancer v1.7.0/go.mod h1:dYmNdSNDKUG+E0SwuFWu+c8CuMBF/l6w1bdzAHxQao0=
 github.com/stackitcloud/stackit-sdk-go/services/logme v0.25.3 h1:fUQLWs2WsXFh+FtFDYOm1kv/gJrGBZLjhVOXJOuYfFY=
 github.com/stackitcloud/stackit-sdk-go/services/logme v0.25.3/go.mod h1:305j9bvzJ+3c4csOw4SUfLSSxRbkpL0osbvqMI89FeM=
+github.com/stackitcloud/stackit-sdk-go/services/logs v0.3.0 h1:N1gerABK2vH7/PBkxZeaWYJ7dz3rjeCHuto+FAuGx3w=
+github.com/stackitcloud/stackit-sdk-go/services/logs v0.3.0/go.mod h1:m4IjH1/RtJOF072kjAB0E/ejoIc++myrKmIahphfO6Q=
+github.com/stackitcloud/stackit-sdk-go/services/logs v0.4.0 h1:EOUVSKvu/m5N+psxeB69IIpANev/jw6HIw2yfh/HO7w=
+github.com/stackitcloud/stackit-sdk-go/services/logs v0.4.0/go.mod h1:m4IjH1/RtJOF072kjAB0E/ejoIc++myrKmIahphfO6Q=
 github.com/stackitcloud/stackit-sdk-go/services/mariadb v0.25.3 h1:Y5Ct3Zi5UcIOwjKMWpKl0nrqiq7psTf4NJv0IKgwTkc=
 github.com/stackitcloud/stackit-sdk-go/services/mariadb v0.25.3/go.mod h1:TMl5WcpjzUiAlLWaxMKbu9ysDzFziSPgg4xLxj9jjfY=
 github.com/stackitcloud/stackit-sdk-go/services/mongodbflex v1.5.5 h1:tPISli81nuvLc5DPqgpvYPSjTySV0wXtMtkfdNXG4CU=
--- a/internal/cmd/beta/beta.go
+++ b/internal/cmd/beta/beta.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/edge"
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/intake"
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/kms"
+	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/logs"
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/sfs"
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/sqlserverflex"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
@@ -48,5 +49,6 @@ func addSubcommands(cmd *cobra.Command, params *types.CmdParams) {
 	cmd.AddCommand(edge.NewCmd(params))
 	cmd.AddCommand(intake.NewCmd(params))
 	cmd.AddCommand(kms.NewCmd(params))
+	cmd.AddCommand(logs.NewCmd(params))
 	cmd.AddCommand(cdn.NewCmd(params))
 }
--- a/internal/cmd/beta/logs/instance/create/create.go
+++ b/internal/cmd/beta/logs/instance/create/create.go
@@ -1,0 +1,176 @@
+package create
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/stackitcloud/stackit-cli/internal/pkg/projectname"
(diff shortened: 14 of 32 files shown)


internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+func NewCmd(params *params.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List CDN distributions",
+		Long:  "List all CDN distributions in your account.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`List all CDN distributions`,
+				`$ stackit beta dns distribution list`,
```
rubenhoenle:
```suggestion
				`$ stackit beta cdn distribution list`,
```

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+			),
+			examples.NewExample(
+				`List all CDN distributions sorted by id`,
+				`$ stackit beta dns distribution list --sort-by=id`,
```
rubenhoenle:
```suggestion
				`$ stackit beta cdn distribution list --sort-by=id`,
```

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+
+			return outputResult(params.Printer, model.OutputFormat, distributions)
+		},
+	}
+
+	configureFlags(cmd)
+	return cmd
+}
+
+var sortByFlagOptions = []string{"id", "created", "updated", "origin-url", "status"}
```
rubenhoenle: Please directly use directly the API values: "id" "updatedAt" "createdAt" "originUrl" "status" "originUrlRelated"

This was wrong in the Jira story.

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+	case "status":
+		return "status"
+	default:
+		panic("invalid sortBy value, programmer error")
+	}
+}
+
+func outputResult(p *print.Printer, outputFormat string, distributions []cdn.Distribution) error {
+	if distributions == nil {
+		distributions = make([]cdn.Distribution, 0) // otherwise prints null in json output
```
rubenhoenle, line 121: That's a good catch indeed! Any chance we can solve this in a central place for all commands?

cgoetz-inovex: Nice idea, I'll investigate this a bit, don't know much about go reflection yet.

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+		for i := range distributions {
+			d := &distributions[i]
+			regions := make([]string, 0, len(*d.Config.Regions))
+			for _, r := range *d.Config.Regions {
+				regions = append(regions, string(r))
+			}
+			joinedRegions := strings.Join(regions, ", ")
+			table.AddRow(
+				utils.PtrString(d.Id),
+				joinedRegions,
```
rubenhoenle, line 138:
```suggestion
				strings.Join(sdkUtils.EnumSliceToStringSlice(*d.Config.Regions), ","),
```

try this one please :smile: 

https://github.com/stackitcloud/stackit-sdk-go/blob/0437d95e48ffe7fea9ffe08f152229ff401419e3/core/utils/utils.go#L17-L29

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,424 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/testutils"
+	sdkConfig "github.com/stackitcloud/stackit-sdk-go/core/config"
+	"github.com/stackitcloud/stackit-sdk-go/services/cdn"
+)
+
+var testProjectId = uuid.NewString()
+var testClient = &cdn.APIClient{}
+var testCtx = context.WithValue(context.Background(), "foo", "foo")
+var testNextPageID = "next-page-id-123"
```
rubenhoenle: could be const I guess :sweat_smile: (yeah I know, nitpick)

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,424 @@
...
+var testTime = time.Now()
+var testID = "dist-1"
```
rubenhoenle: same here

internal/cmd/beta/cdn/distribution/list/list.go
rubenhoenle: Reflection is considered a no-no, see https://principles.schwarz/principles/engineering-principles/go/avoid-using-reflection

internal/cmd/beta/cdn/distribution/list/list.go
rubenhoenle: Maybe there's also a enum generated in the SDK which holds these values. Then this would be the way to go because it will be kept up-to-date in the future without any efforts on our side.

internal/cmd/beta/cdn/distribution/list/list_test.go
cgoetz-inovex: We need to take the address of `testNextPageID` in some tests to pass it to the generated API -> const does not work.

internal/cmd/beta/cdn/distribution/list/list_test.go
cgoetz-inovex: We also need the address of this one.

internal/cmd/beta/cdn/distribution/list/list.go
cgoetz-inovex: Searched for `sortBy` and `originUrl` etc. could not find an enum for this. The spec looks like it should generate an enum: https://github.com/stackitcloud/stackit-api-specifications/blob/83214868e6dbdc53d053cc07654542d8f34b5491/services/cdn/v1beta2/cdn.json#L1486

How could I start to investigate this further?

internal/cmd/beta/cdn/distribution/list/list.go
cgoetz-inovex: With reflection I'd add this in `Printer.OutputResult()`:

```go
	val := reflect.ValueOf(output)
	if val.IsNil() {
		// when passing a nil slice or map, JSON and YAML marshal to "null"
		// so we need to create an empty slice or map to have "[]" or "{}" output instead
		switch val.Kind() {
		case reflect.Slice:
			output = make([]any, 0)
		case reflect.Map:
			output = make(map[any]any)
		default:
			// do nothing
		}
	}
```

The linked document suggests type switches, which would not work here. 

Another solution without reflection, that comes to mind, would be to introduce specialized `OutputResult` functions, like `OutputResultSlice/Map`.

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+		distributions = make([]cdn.Distribution, 0) // otherwise prints null in json output
+	}
+	return p.OutputResult(outputFormat, distributions, func() error {
+		if len(distributions) == 0 {
+			p.Outputln("No CDN distributions found")
+			return nil
+		}
+
+		table := tables.NewTable()
+		table.SetHeader("ID", "REGIONS", "STATUS")
```
cgoetz-inovex, line 130: @rubenhoenle, when using the `table` output we only get these columns. When using the JSON format we would get all fields of `cdn.Distribution`. Should I introduce here custom struct that only contains `ID`, `REGIONS` and `STATUS` in JSON?

internal/cmd/beta/cdn/distribution/list/list_test.go
cgoetz-inovex: nvm. just discovered `utils.Ptr()` for this.

internal/cmd/beta/cdn/distribution/list/list.go
rubenhoenle: No, please don't. Just build a thin layer around the API and print what the API returns.

internal/pkg/utils/strings.go
```diff
@@ -26,6 +27,18 @@ func JoinStringKeysPtr(m map[string]any, sep string) string {
 	return JoinStringKeys(m, sep)
 }
 
+// JoinStringMap concatenates the key-value pairs of a string map, key and value separated by kvSep, key value pairs separated by sep.
+func JoinStringMap(m map[string]string, kvSep, sep string) string {
```
rubenhoenle: it's fine, can stay here in this pkg

rubenhoenle:
```suggestion
func JoinStringMap(m map[string]string, keyValueSeparator, separator string) string {
```

internal/pkg/utils/strings.go
```diff
@@ -26,6 +27,18 @@ func JoinStringKeysPtr(m map[string]any, sep string) string {
 	return JoinStringKeys(m, sep)
 }
 
+// JoinStringMap concatenates the key-value pairs of a string map, key and value separated by kvSep, key value pairs separated by sep.
```
rubenhoenle:
```suggestion
// JoinStringMap concatenates the key-value pairs of a string map, key and value separated by keyValueSeparator, key value pairs separated by separator.
```

internal/pkg/utils/strings.go
rubenhoenle: I think it's more readable if you write it out, isn't it?

internal/pkg/utils/strings_test.go
```diff
@@ -30,3 +30,39 @@ func TestTruncate(t *testing.T) {
 		})
 	}
 }
+
+func TestJoinStringMap(t *testing.T) {
+	tests := []struct {
+		name string
+		m    map[string]string
```
rubenhoenle:
```suggestion
		input    map[string]string
```

Sorry, I'm nitpicking. But please use proper names.

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+				`List all CDN distributions`,
+				`$ stackit beta cdn distribution list`,
+			),
+			examples.NewExample(
+				`List all CDN distributions sorted by id`,
+				`$ stackit beta cdn distribution list --sort-by=id`,
+			),
+		),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background() // should this be cancellable?
```
rubenhoenle:
```suggestion
			ctx := context.Background()
```

We have it like this everywhere. And as of now I don't see any reason to change that. Any objections?

internal/cmd/beta/cdn/distribution/list/list.go
rubenhoenle: :+1: 

Could be an option, please feel free to create a Jira ticket and propose a solution :blush:

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	return p.OutputResult(outputFormat, distributions, func() error {
+		if len(distributions) == 0 {
+			p.Outputln("No CDN distributions found")
+			return nil
+		}
+
+		table := tables.NewTable()
+		table.SetHeader("ID", "REGIONS", "STATUS")
+		for i := range distributions {
```
rubenhoenle:
```suggestion
		for i, d := range distributions {
```

Saves yourself the `d := &distributions[i]` line below, not?

internal/cmd/beta/cdn/distribution/list/list.go
```diff
@@ -0,0 +1,173 @@
...
+			d := &distributions[i]
+			joinedRegions := strings.Join(sdkUtils.EnumSliceToStringSlice(*d.Config.Regions), ", ")
```
rubenhoenle: Take care of the potential nil pointer reference here please

internal/pkg/flags/flag_to_value.go
```diff
@@ -47,6 +47,20 @@ func FlagToStringSliceValue(p *print.Printer, cmd *cobra.Command, flag string) [
 	return nil
 }
 
+// Returns the flag's value as a []string.
+// Returns nil if flag is not set, if its value can not be converted to []string, or if the flag does not exist.
+func FlagToStringArrayValue(p *print.Printer, cmd *cobra.Command, flag string) []string {
```
rubenhoenle, line 52:
```suggestion
func FlagToStringSliceValue(p *print.Printer, cmd *cobra.Command, flag string) []string {
```

won't we call it slice in golang?

rubenhoenle, line 52: Please add some unit tests for this new func

internal/pkg/flags/flag_to_value.go
```diff
@@ -75,6 +89,20 @@ func FlagToInt64Pointer(p *print.Printer, cmd *cobra.Command, flag string) *int6
 	return nil
 }
 
+// Returns a pointer to the flag's value.
+// Returns nil if the flag is not set, if its value can not be converted to int64, or if the flag does not exist.
+func FlagToInt32Pointer(p *print.Printer, cmd *cobra.Command, flag string) *int32 {
```
rubenhoenle, line 94: same here, please add some unit tests

internal/cmd/beta/cdn/distribution/update/update_test.go
```diff
@@ -0,0 +1,365 @@
...
+					m.MonthlyLimitBytes = utils.Ptr(testMonthlyLimitBytes)
+					m.Optimizer = utils.Ptr(true)
+				},
+			),
+		},
+		{
+			description: "max config http backend",
+			argValues:   []string{testDistributionID},
+			flagValues: fixtureFlagValues(
+				func(m map[string]string) {
```
rubenhoenle:
```suggestion
				func(flagValues map[string]string) {
```

same here, I would prefer proper names if possible

internal/cmd/beta/cdn/distribution/update/update_test.go
```diff
@@ -0,0 +1,365 @@
...
+					m[flagHTTP] = "true"
+					m[flagHTTPOriginURL] = "https://origin.example.com"
+					m[flagHTTPOriginRequestHeaders] = "X-Example-Header: example-value, X-Another-Header: another-value"
+					m[flagHTTPGeofencing] = "https://dach.example.com DE,AT,CH"
+				},
+			),
+			isValid: true,
+			expected: fixtureInputModel(
+				func(m *inputModel) {
```
rubenhoenle:
```suggestion
				func(model *inputModel) {
```
same here, I would prefer proper names if possible

internal/cmd/beta/cdn/distribution/update/update_test.go
rubenhoenle: Applies to all tests in this table test

internal/cmd/beta/cdn/distribution/update/update_test.go
rubenhoenle: Applies to all tests in this table test

internal/cmd/beta/cdn/distribution/update/update_test.go
```diff
@@ -0,0 +1,365 @@
...
+		DistributionID: testDistributionID,
+		Regions:        []cdn.Region{},
+	}
+	for _, mod := range mods {
+		mod(model)
+	}
+	return model
+}
+
+func fixtureRequest(mods ...func(p *cdn.PatchDistributionPayload)) cdn.ApiPatchDistributionRequest {
```
rubenhoenle:
```suggestion
func fixtureRequest(mods ...func(payload *cdn.PatchDistributionPayload)) cdn.ApiPatchDistributionRequest {
```

internal/cmd/beta/cdn/distribution/update/update_test.go
```diff
@@ -0,0 +1,365 @@
...
+	req := testClient.PatchDistribution(testCtx, testProjectId, testDistributionID)
+	if p := fixturePayload(mods...); p != nil {
+		req = req.PatchDistributionPayload(*fixturePayload(mods...))
+	}
+	return req
+}
+
+func fixturePayload(mods ...func(p *cdn.PatchDistributionPayload)) *cdn.PatchDistributionPayload {
```
rubenhoenle:
```suggestion
func fixturePayload(mods ...func(payload *cdn.PatchDistributionPayload)) *cdn.PatchDistributionPayload {
```

internal/cmd/beta/cdn/distribution/update/update.go
```diff
@@ -0,0 +1,372 @@
...
+		MonthlyLimitBytes:    monthlyLimit,
+		Loki:                 loki,
+		Optimizer:            optimizer,
+	}
+
+	p.DebugInputModel(model)
+	return &model, nil
+}
+
+// TODO both parseGeofencing and parseOriginRequestHeaders copied from create.go, move to another package and make public?
```
rubenhoenle: To resolve this TODO: we have some service-specific utils packages usually to not duplicate such stuff.

E.g. here's the one from IaaS: https://github.com/stackitcloud/stackit-cli/blob/main/internal/pkg/services/iaas/utils/utils.go

Move the funcs there so they can be re-used.

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+	flagValues := map[string]string{
+		globalflags.ProjectIdFlag: testProjectId,
+	}
+	for _, mod := range mods {
+		mod(flagValues)
+	}
+	return flagValues
+}
+
+func flagSortBy(sortBy string) func(m map[string]string) {
```
rubenhoenle: Not really a fan of these additional funcs... They make the test setup more complex than it is already.

Why not just stick with this:

```golang

		    flagValues[sortByFlag] = "value"
		}),
```

I'm not seeing the point of adding such a func to set single string flags. 

For a complex group of flags, it can make sense. But for such simple cases it isn't worth it IMHO :sweat_smile:

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+	return func(m map[string]string) {
+		m[sortByFlag] = sortBy
+	}
+}
+
+func flagProjectId(id *string) func(m map[string]string) {
```
rubenhoenle: same here

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+		},
+		SortBy: "createdAt",
+	}
+	for _, mod := range mods {
+		mod(m)
+	}
+	return m
+}
+
+func inputSortBy(sortBy string) func(m *inputModel) {
```
rubenhoenle: same here

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+	return func(m *inputModel) {
+		m.SortBy = sortBy
+	}
+}
+
+func fixtureRequest(mods ...func(r cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest {
```
rubenhoenle:
```suggestion
func fixtureRequest(mods ...func(request cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest {
```

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+	r = r.PageSize(100)
+	r = r.SortBy("createdAt")
+	for _, mod := range mods {
+		r = mod(r)
+	}
+	return r
+}
+
+func requestSortBy(sortBy string) func(r cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest {
+	return func(r cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest {
```
rubenhoenle: same here

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+		return r.SortBy(sortBy)
+	}
+}
+
+func requestNextPageID(nextPageID string) func(r cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest {
+	return func(r cdn.ApiListDistributionsRequest) cdn.ApiListDistributionsRequest {
```
rubenhoenle: same here

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+		statusCode: 200,
+	}
+	for _, mod := range mods {
+		mod(&r)
+	}
+	return r
+}
+
+func responseStatus(statusCode int) func(r *testResponse) {
+	return func(r *testResponse) {
```
rubenhoenle: same here

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+		r.statusCode = statusCode
+	}
+}
+
+func responseNextPageID(nextPageID *string) func(r *testResponse) {
+	return func(r *testResponse) {
```
rubenhoenle: same here

internal/cmd/beta/cdn/distribution/list/list_test.go
```diff
@@ -0,0 +1,469 @@
...
+						Regions: &[]cdn.Region{
+							cdn.REGION_EU,
+							cdn.REGION_AF,
+						},
+					},
+					Status: utils.Ptr(testStatus),
+				},
+			},
+			expected: `
+ ID     │ REGIONS │ STATUS
```
rubenhoenle, line 444: some question I have in my mind here: Would it make sense maybe to create some abstraction for table outputs (or mocks) so we can properly test the table output without having to copy formatted tables into our unit tests.

**Don't get me wrong, I absolutely (!!) love the idea that we will test the table output for all subcommands in the future**, but these formatted tables copied into our unit test files could become a pain to maintain :smile:

internal/cmd/beta/cdn/distribution/update/update.go
rubenhoenle: > Move the funcs there so they can be re-used.

Edit: What I really wanted to tell: Create a util package for cdn and move the funcs there. Don't move them into the IaaS util package please :joy:

internal/cmd/beta/cdn/distribution/describe/describe_test.go
```diff
@@ -0,0 +1,406 @@
...
+			args:        []string{testDistributionID},
+			flags: map[string]string{
+				globalflags.ProjectIdFlag: "invalid-uuid",
+			},
+			isValid: false,
+		},
+		{
+			description: "with WAF",
+			args:        []string{testDistributionID},
+			flags: fixtureFlagValues(func(m map[string]string) {
```
rubenhoenle:
```suggestion
			flags: fixtureFlagValues(func(flagValues map[string]string) {
```

internal/cmd/beta/cdn/distribution/describe/describe_test.go
```diff
@@ -0,0 +1,406 @@
...
+				m[flagWithWaf] = "true"
+			}),
+			isValid: true,
+			expected: fixtureInputModel(func(m *inputModel) {
```
rubenhoenle:
```suggestion
			expected: fixtureInputModel(func(model *inputModel) {
```

internal/cmd/beta/cdn/distribution/describe/describe_test.go
```diff
@@ -0,0 +1,406 @@
...
+		expected    cdn.ApiGetDistributionRequest
+	}{
+		{
+			description: "base",
+			model:       fixtureInputModel(),
+			expected:    testClient.GetDistribution(testCtx, testProjectID, testDistributionID).WithWafStatus(false),
+		},
+		{
+			description: "with WAF",
+			model: fixtureInputModel(func(m *inputModel) {
```
rubenhoenle:
```suggestion
			model: fixtureInputModel(func(model *inputModel) {
```

internal/cmd/beta/cdn/distribution/list/list.go
cgoetz-inovex: General question:
Is there a general approach? So should we check every pointer for nil?
The API spec for the response marks  every step of `distributions.config.regions` as required.

I'll add nil checks but would be interested how our general approach is.

internal/pkg/flags/flag_to_value.go
cgoetz-inovex: `FlagToStringSliceValue` already exists.
The difference is in the first line:
```go
func FlagToStringSliceValue(p *print.Printer, cmd *cobra.Command, flag string) []string {
	value, err := cmd.Flags().GetStringSlice(flag)
```
vs.
```go
func FlagToStringArrayValue(p *print.Printer, cmd *cobra.Command, flag string) []string {
	value, err := cmd.Flags().GetStringArray(flag)
```
--> the difference is which function on `Flags` is called.
from [StringArray Docs](https://pkg.go.dev/github.com/spf13/pflag#StringArray):
> StringArray defines a string flag with specified name, default value, and usage string. The return value is the address of a []string variable that stores the value of the flag. The value of each argument will not try to be separated by comma. Use a StringSlice for that.

I've declared:
```go
	cmd.Flags().StringArray(flagHTTPGeofencing, []string{}, "Geofencing rules for HTTP backend in the format 'https://example.com US,DE'. URL and countries have to be quoted. Repeatable.")
```
As a string array. This allows us to configure multiple geofencing rules like this:
```
$ stackit .... \
--http-geofencing="https://foo.example.com DE,CH"
--http-geofencing="https://bar.example.com AT"
```
If I'd use `StringSlice` instead, the pflag library would try to split the first `--http-geofencing` flag into two values: `https://foo.example.com DE` and `CH`.

TLDR: naming is here consistent with the usage of pflag library.

internal/cmd/beta/cdn/distribution/list/list_test.go
cgoetz-inovex: Changed this.

Some arguments in favor of such funcs:
- less typing when writing tests (although completion models alleviate this a bit)
- less reading 
- easier to refactor if the specifics change
- could be considered idiomatic: https://golang.cafe/blog/golang-functional-options-pattern

internal/cmd/beta/cdn/distribution/list/list_test.go
cgoetz-inovex: Also thought about that, especially that the strings are padded with spaces is a bit painful.
For maintaining the current solution I used a breakpoint in the failure branch and evaluated`buf.String()` in the debugger, copying the result into the test and adding placeholders.

I'd add this point to our discussion to also test the run functions. There are a lot of possibilities how we could solve this. We could write a matcher library for this output or could go as far as the stuff mentioned in https://research.swtch.com/testing . Or something in between.

internal/pkg/flags/flag_to_value.go
rubenhoenle: Thanks, didn't know that. Fine then :smile:

internal/cmd/beta/cdn/distribution/describe/describe.go
```diff
@@ -0,0 +1,220 @@
...
+	table.AddSeparator()
+	table.AddRow("DEFAULT CACHE DURATION", defaultCacheDuration)
+	table.AddSeparator()
+	table.AddRow("LOG SINK PUSH URL", logSinkPushUrl)
+	table.AddSeparator()
+	table.AddRow("MONTHLY LIMIT (BYTES)", monthlyLimitBytes)
+	table.AddSeparator()
+	table.AddRow("OPTIMIZER ENABLED", optimizerEnabled)
+	table.AddSeparator()
+	// TODO config has yet another WAF block, left it out because the docs say to use the WAF block at the top level to determine enabled rules. There's also mode and type fields here, both left out.
```
rubenhoenle: open TODO

internal/cmd/beta/cdn/distribution/create/create.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			model, err := parseInput(params.Printer, cmd, args)
+			if err != nil {
+				return err
+			}
+			if model.Bucket != nil {
+				pw, err := params.Printer.PromptForPassword("enter your secret access key for the object storage bucket: ")
```
rubenhoenle, line 122: How does this work with scripts? Could you add an example please which shows this use-case? :sweat_smile:

internal/cmd/beta/cdn/distribution/create/create.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	return req.CreateDistributionPayload(*payload)
+}
+
+func outputResult(p *print.Printer, outputFormat, projectLabel string, resp *cdn.CreateDistributionResponse) error {
+	if resp == nil {
+		return fmt.Errorf("create distribution response is nil")
+	}
+	return p.OutputResult(outputFormat, resp, func() error {
+		p.Outputf("Created CDN distribution for %q. Id: %s\n", projectLabel, utils.PtrString(resp.Distribution.Id))
```
rubenhoenle:
```suggestion
		p.Outputf("Created CDN distribution for %q. ID: %s\n", projectLabel, utils.PtrString(resp.Distribution.Id))
```

image

During the update it's upper-case. Please stay consistent within your implementation :sweat_smile:

internal/cmd/beta/cdn/distribution/delete/delete.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			_, err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("delete loadbalancer: %w", err)
+			}
+
+			params.Printer.Outputf("CDN distribution %q deleted.", model.DistributionID)
```
rubenhoenle: I get some weird `%` at the end of the output when testing this.

image

And in addition, it's possible to delete non-existent distributions. I just took some random uuid from https://www.uuidgenerator.net/  and got the message that the distribution was deleted. I would expect an HTTP 404 error in that case from the API. Could you check this please with the service team?

internal/cmd/beta/cdn/distribution/create/create.go
cgoetz-inovex: nice catch, `Print.PromptForPassword` would need some changes:
```
@@ -172,11 +172,16 @@ func (p *Printer) PromptForEnter(prompt string) error {
 func (p *Printer) PromptForPassword(prompt string) (string, error) {
        p.Cmd.PrintErr(prompt)
        defer p.Outputln("")
-       bytePassword, err := term.ReadPassword(int(syscall.Stdin))
-       if err != nil {
-               return "", fmt.Errorf("read password: %w", err)
+       if term.IsTerminal(syscall.Stdin) {
+               bytePassword, err := term.ReadPassword(int(syscall.Stdin))
+               if err != nil {
+                       return "", fmt.Errorf("read password: %w", err)
+               }
+               return string(bytePassword), nil
        }
-       return string(bytePassword), nil
+       // Fallback for non-terminal environments
+       reader := bufio.NewReader(p.Cmd.InOrStdin())
+       return reader.ReadString('\n')
 }
 ```
 With this a script like:
 ```
 #!/usr/bin/env bash

cat secret | bin/stackit beta cdn distribution create -y --project-id xxx --bucket --bucket-url https://bucket.example.com \
  --bucket-credentials-access-key-id yyyy --bucket-region EU \
  --regions AF,EU
 ```
 works.
 
 Should we change `Print.PromptForPassword` accordingly?
 
 Other commands (observability-credentials/update; loadbalancer/add) read passwords as command line arguments, which can be read by other processes.
 
 Imo we should use env vars and/or stdin.

File: internal/cmd/beta/cdn/distribution/delete/delete.go (hunk 37 above)
Comment (cgoetz-inovex) [reply internal/cmd/beta/cdn/distribution/delete/delete.go]: [Checked the docs](https://docs.api.stackit.cloud/documentation/cdn/version/v1#tag/Distribution/operation/DeleteDistribution):
> Once synchronized, the distribution is deleted. Further attempts to delete this response will return an empty response.
--> this looks like by design.

regarding the `%` char: fixed by adding a line break at the end of the message.

File: internal/cmd/beta/cdn/distribution/create/create.go (hunk 35 above)
Comment (rubenhoenle) [reply internal/cmd/beta/cdn/distribution/create/create.go:122]: I would prefer stdin, else we have a mess with hundreds of env variables some day :sweat_smile:

Comment (rubenhoenle) [reply internal/cmd/beta/cdn/distribution/create/create.go:122]: > Should we change Print.PromptForPassword accordingly?

yes, please. If you're confident to do that do it within this PR. If not please create a Jira issue for it and bring it up in the refinement @cgoetz-inovex  :sweat_smile:

File: internal/cmd/beta/cdn/distribution/delete/delete.go (hunk 37 above)
Comment (rubenhoenle) [reply internal/cmd/beta/cdn/distribution/delete/delete.go]: Nice, thanks

Comment (rubenhoenle) [reply internal/cmd/beta/cdn/distribution/delete/delete.go]: > --> this looks like by design.

Ok, then that's none of our business I guess.

File: internal/cmd/beta/cdn/distribution/create/create.go (hunk 35 above)
Comment (rubenhoenle) [reply internal/cmd/beta/cdn/distribution/create/create.go:122]: > Other commands (observability-credentials/update; loadbalancer/add) read passwords as command line arguments, which can be read by other processes.

Also please feel free to open Jira issues for these args so we can deprecate them in the future and replace them with appropriate input methods. :wink:

Comment (cgoetz-inovex) [reply internal/cmd/beta/cdn/distribution/create/create.go:122]: pushed changes to PromptForPassword and added tests.
Also added an example to `cdn/distribution/create` for this

File: internal/pkg/print/print.go [hunk 38]
```diff
@@ -182,11 +182,20 @@ func (p *Printer) PromptForEnter(prompt string) error {
...
+	}
+	// Fallback for non-terminal environments
+	reader := bufio.NewReader(p.Cmd.InOrStdin())
+	pw, err := reader.ReadString('\n')
 	if err != nil {
-		return "", fmt.Errorf("read password: %w", err)
+		return "", fmt.Errorf("read password from non-terminal: %w", err)
 	}
-	return string(bytePassword), nil
+	return pw[:len(pw)-1], nil // remove trailing newline
```
rubenhoenle, line 198: Do we need to take care so we don't run out of bounds here?

cgoetz-inovex: Thought about adding a len check. [docs](https://pkg.go.dev/bufio#Reader.ReadString)
> ... returning a string containing the data up to and including the delimiter. ...
> ReadString returns err != nil if and only if the returned data does not end in delim.

--> if we passed the err check we always have `len(pw)` > 0 here

rubenhoenle: Ok nice, didn't know that. Thanks :sweat_smile:

internal/cmd/beta/cdn/distribution/delete/delete.go
```diff
@@ -0,0 +1,94 @@
...
+				return err
+			}
+
+			projectLabel, err := projectname.GetProjectName(ctx, params.Printer, params.CliVersion, cmd)
+			if err != nil {
+				params.Printer.Debug(print.ErrorLevel, "get project name: %v", err)
+				projectLabel = model.ProjectId
+			}
+
+			if !model.AssumeYes {
```
rubenhoenle: I think this one isn't needed anymore after integrating your other PR (#1217) :sweat_smile:

internal/cmd/beta/cdn/distribution/update/update.go
```diff
@@ -0,0 +1,338 @@
...
+			if err != nil {
+				return err
+			}
+
+			projectLabel, err := projectname.GetProjectName(ctx, params.Printer, params.CliVersion, cmd)
+			if err != nil {
+				params.Printer.Debug(print.ErrorLevel, "get project name: %v", err)
+				projectLabel = model.ProjectId
+			}
+			if !model.AssumeYes {
```
rubenhoenle: I think this one isn't needed anymore after integrating your other PR (#1217) :sweat_smile:

internal/cmd/beta/cdn/distribution/delete/delete.go
cgoetz-inovex: nice catch

PR #1130, opened 2025-11-27
Refactor to support separate auth for the SDK (and thus terraform) + small bugfixes

franklouwers:
### Add API Authentication for Terraform Provider Integration

This PR implements the CLI authentication feature proposed in [terraform-provider-stackit#880](https://github.com/stackitcloud/terraform-provider-stackit/discussions/880), enabling the Terraform Provider and SDK to use CLI user credentials instead of requiring service accounts for local development. (See https://github.com/stackitcloud/terraform-provider-stackit/issues/719 for the issue that triggered this as well)

#### Overview

Currently, the STACKIT Terraform Provider and SDK only support service account authentication (Key Flow and Token Flow), requiring users to create and manage service account credentials even for local development. This PR adds a new `stackit auth api` command group that allows external tools to leverage the CLI's convenient OAuth2 user authentication.

I also fixed a bug with token refresh expiration times.

#### Changes

This implementation follows the RFC's preferred approach: **keychain storage with file-based fallback**.

##### 1. Storage Context System

Refactored the storage layer to support multiple independent credential contexts:
- **CLI context**: For `stackit auth login` (existing behavior unchanged)
- **API context**: For `stackit auth api login` (new, isolated storage)

**Storage locations:**
- Keychain: `stackit-cli-api` (macOS Keychain, Windows Credential Manager, Linux Secret Service)
- File fallback: `~/.stackit/cli-api-auth-storage.txt` (base64-encoded)

##### 2. Token Refresh Fix

**Critical improvement**: Fixed token refresh mechanism to use actual JWT expiration instead of session expiration:
- **Before**: Stored session expiry (default 2h from config), causing SDK to think tokens were valid when they'd already expired
- **After**: Parse and store JWT `exp` claim (typically 15min), enabling proper automatic token refresh

This fix applies to both `stackit auth login` and `stackit auth api login`.

##### 3. New Commands

Added `stackit auth api` subcommand group:

```bash
# Authenticate for Terraform/SDK (opens browser)
stackit auth api login

# Check authentication status
stackit auth api status

# Get access token with automatic refresh
stackit auth api get-access-token

# Remove API credentials
stackit auth api logout
```

#### Benefits

- **No service accounts required** for local Terraform development
- **Automatic token refresh** with bidirectional storage sync
- **Independent credentials**: CLI and API can use different accounts simultaneously
- **Profile support**: Each profile has independent CLI and API authentication
- **Secure storage**: System keychain with automatic file fallback

#### Usage Example

```bash
# User authenticates once
$ stackit auth api login
# Opens browser, stores credentials in isolated API context

# Terraform Provider can now use these credentials
# (Requires provider update to support CLI auth - separate PR)
```

#### Implementation Notes

- Maintains full backward compatibility with existing auth flows
- No changes to `stackit auth login` behavior
- Storage contexts prevent credential conflicts
- Test coverage

#### Related

- RFC: https://github.com/stackitcloud/terraform-provider-stackit/discussions/880
- Terraform provider issue: https://github.com/stackitcloud/terraform-provider-stackit/issues/719
- Follow-up PR needed in terraform-provider-stackit to consume these credentials

## Checklist

- [x] Issue was linked above

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@v44.2.0
+        uses: renovatebot/github-action@v44.2.3
         with:
           configurationFile: .github/renovate.json
           token: ${{ secrets.RENOVATE_TOKEN }}
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/inhies/go-bytesize v0.0.0-20220417184213-4913239db9cf
-	github.com/jedib0t/go-pretty/v6 v6.7.7
+	github.com/jedib0t/go-pretty/v6 v6.7.8
 	github.com/lmittmann/tint v1.1.2
 	github.com/mattn/go-colorable v0.1.14
 	github.com/spf13/cobra v1.10.2
@@ -17,7 +17,7 @@ require (
 	github.com/spf13/viper v1.21.0
 	github.com/stackitcloud/stackit-sdk-go/core v0.20.1
 	github.com/stackitcloud/stackit-sdk-go/services/alb v0.7.3
-	github.com/stackitcloud/stackit-sdk-go/services/authorization v0.10.1
+	github.com/stackitcloud/stackit-sdk-go/services/authorization v0.11.0
 	github.com/stackitcloud/stackit-sdk-go/services/dns v0.17.3
 	github.com/stackitcloud/stackit-sdk-go/services/git v0.10.1
 	github.com/stackitcloud/stackit-sdk-go/services/iaas v1.3.0
@@ -261,7 +261,7 @@ require (
 	github.com/sourcegraph/conc v0.3.1-0.20240121214520-5f936abd7ae8 // indirect
 	github.com/spf13/afero v1.15.0 // indirect
 	github.com/spf13/cast v1.10.0 // indirect
-	github.com/stackitcloud/stackit-sdk-go/services/kms v1.1.2
+	github.com/stackitcloud/stackit-sdk-go/services/kms v1.2.0
 	github.com/stackitcloud/stackit-sdk-go/services/loadbalancer v1.6.2
 	github.com/stackitcloud/stackit-sdk-go/services/logme v0.25.3
 	github.com/stackitcloud/stackit-sdk-go/services/mariadb v0.25.3
--- a/go.sum
+++ b/go.sum
@@ -371,8 +371,8 @@ github.com/inconshreveable/mousetrap v1.1.0 h1:wN+x4NVGpMsO7ErUn/mUI3vEoE6Jt13X2
 github.com/inconshreveable/mousetrap v1.1.0/go.mod h1:vpF70FUmC8bwa3OWnCshd2FqLfsEA9PFc4w1p2J65bw=
 github.com/inhies/go-bytesize v0.0.0-20220417184213-4913239db9cf h1:FtEj8sfIcaaBfAKrE1Cwb61YDtYq9JxChK1c7AKce7s=
 github.com/inhies/go-bytesize v0.0.0-20220417184213-4913239db9cf/go.mod h1:yrqSXGoD/4EKfF26AOGzscPOgTTJcyAwM2rpixWT+t4=
-github.com/jedib0t/go-pretty/v6 v6.7.7 h1:Y1Id3lJ3k4UB8uwWWy3l8EVFnUlx5chR5+VbsofPNX0=
-github.com/jedib0t/go-pretty/v6 v6.7.7/go.mod h1:YwC5CE4fJ1HFUDeivSV1r//AmANFHyqczZk+U6BDALU=
+github.com/jedib0t/go-pretty/v6 v6.7.8 h1:BVYrDy5DPBA3Qn9ICT+PokP9cvCv1KaHv2i+Hc8sr5o=
+github.com/jedib0t/go-pretty/v6 v6.7.8/go.mod h1:YwC5CE4fJ1HFUDeivSV1r//AmANFHyqczZk+U6BDALU=
 github.com/jgautheron/goconst v1.8.2 h1:y0XF7X8CikZ93fSNT6WBTb/NElBu9IjaY7CCYQrCMX4=
 github.com/jgautheron/goconst v1.8.2/go.mod h1:A0oxgBCHy55NQn6sYpO7UdnA9p+h7cPtoOZUmvNIako=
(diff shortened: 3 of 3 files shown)


franklouwers:
Keep

internal/cmd/auth/api/provider.go
```diff
@@ -0,0 +1,35 @@
...
+
+func NewCmd(params *params.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "api",
+		Short: "Manages authentication for the STACKIT Terraform Provider and SDK",
+		Long: `Manages authentication for the STACKIT Terraform Provider and SDK.
+
+These commands allow you to authenticate with your personal STACKIT account
+and share the credentials with the STACKIT Terraform Provider and SDK.
+This provides an alternative to using service accounts for local development.`,
```
cgoetz-inovex: Maybe we could add some words about 'Storage Contexts' and locations here. Sth. along the lines of this maybe?
```suggestion
This provides an alternative to using service accounts for local development.

Tokens are stored separately from tokens received from "stackit auth login", this allows using separate accounts when using the cli directly vs. using it to manage credentials for the SDK and TF provider.

Tokens are stored in the OS keychain with a fallback to local storage.`,
```

internal/cmd/auth/api/get-access-token/get_access_token.go
```diff
@@ -0,0 +1,89 @@
...
+				params.Printer.Outputln(accessToken)
+
+				return nil
+			}
+		},
+	}
+
+	// hide project id flag from help command because it could mislead users
+	cmd.SetHelpFunc(func(command *cobra.Command, strings []string) {
+		_ = command.Flags().MarkHidden(globalflags.ProjectIdFlag) // nolint:errcheck // there's no chance to handle the error here
```
cgoetz-inovex: just a note, since this is copied verbatim from `stackit auth get-access-token` and failure seems almost impossible: we could use `cobra.CheckErr()` here to handle an error.
Maybe it's worth to do this (and also in `auth get-access-token`) so if it fails we would get an error.

internal/pkg/auth/auth.go
```diff
@@ -208,3 +251,53 @@ func GetValidAccessToken(p *print.Printer) (string, error) {
 	// Return the new access token
 	return utf.accessToken, nil
 }
+
+// debugHTTPRequest logs the raw HTTP request details for debugging purposes
+func debugHTTPRequest(p *print.Printer, req *http.Request) {
+	if p == nil || req == nil {
+		return
+	}
```
cgoetz-inovex, line 267:
```suggestion
	}
	if !p.IsVerbosityDebug() {
		return
	}
```
we could save us some work here

internal/pkg/auth/auth.go
```diff
@@ -208,3 +251,53 @@ func GetValidAccessToken(p *print.Printer) (string, error) {
...
+		}
+	}
+	p.Debug(print.DebugLevel, "===================")
+}
+
+// debugHTTPResponse logs the raw HTTP response details for debugging purposes
+func debugHTTPResponse(p *print.Printer, resp *http.Response) {
+	if p == nil || resp == nil {
+		return
+	}
```
cgoetz-inovex, line 288:
```suggestion
	}
	if !p.IsVerbosityDebug() {
		return
	}
```
we could save us some work here

internal/cmd/auth/api/get-access-token/get_access_token.go
```diff
@@ -0,0 +1,89 @@
...
+					return fmt.Errorf("marshal access token: %w", err)
+				}
+				params.Printer.Outputln(string(details))
+
+				return nil
+			default:
+				params.Printer.Outputln(accessToken)
+
+				return nil
+			}
```
cgoetz-inovex:
```suggestion
			result := map[string]string{
				"access_token": accessToken,
			}
			return params.Printer.OutputResult(model.OutputFormat, result, func() error {
				params.Printer.Outputln(accessToken)
				return nil
			})
```
this would also handle `print.YAMLOutputFormat`

internal/cmd/auth/api/status/status.go
```diff
@@ -0,0 +1,109 @@
...
+			if status.Email != "" {
+				p.Outputf("Email: %s\n", status.Email)
+			}
+			p.Outputf("Auth Flow: %s\n", status.AuthFlow)
+		} else {
+			p.Outputln("API Authentication Status: Not authenticated")
+			p.Outputln("\nTo authenticate, run: stackit auth api login")
+		}
+		return nil
+	}
```
cgoetz-inovex:
```suggestion
	return p.OutputResult(model.OutputFormat, status, func() error {
		if status.Authenticated {
			p.Outputln("API Authentication Status: Authenticated")
			if status.Email != "" {
				p.Outputf("Email: %s\n", status.Email)
			}
			p.Outputf("Auth Flow: %s\n", status.AuthFlow)
		} else {
			p.Outputln("API Authentication Status: Not authenticated")
			p.Outputln("\nTo authenticate, run: stackit auth api login")
		}
		return nil
	})
```
this one would also handle YAML output

internal/cmd/auth/api/get-access-token/get_access_token.go
franklouwers: resolved in [696ce71](https://github.com/stackitcloud/stackit-cli/pull/1130/commits/696ce713f798dd111131013a72b76939866be559)

internal/cmd/auth/api/logout/logout.go
```diff
@@ -0,0 +1,35 @@
...
+				`Log out from the STACKIT Terraform Provider and SDK`,
+				"$ stackit auth api logout"),
+		),
+		RunE: func(_ *cobra.Command, _ []string) error {
+			err := auth.LogoutUserWithContext(auth.StorageContextAPI)
+			if err != nil {
+				return fmt.Errorf("log out failed: %w", err)
+			}
+
+			p.Printer.Info("Successfully logged out from STACKIT Terraform Provider and SDK.\n")
```
marceljk: Probably you just copied it from the existing logout command, but here `Outputln` should be used. 
```suggestion
			p.Printer. Outputln("Successfully logged out from STACKIT Terraform Provider and SDK.\n")
```
Info prints it to sterr and Outputln to stdout.

Can you adjust it for the existing logout command as well?

cgoetz-inovex:
@franklouwers, CI failing because of out-of-date docs. Please run `make generate-docs`.

internal/cmd/auth/api/logout/logout.go
franklouwers: Updated in [7ee08bb](https://github.com/stackitcloud/stackit-cli/pull/1130/commits/7ee08bb390762e3f60ef273f1c3c58df1f977625)

franklouwers:
@cgoetz-inovex done in [0ee843f](https://github.com/stackitcloud/stackit-cli/pull/1130/commits/0ee843f4feb1c359534b6be73a6794b4a32d0e48)

PR #1155, opened 2025-12-07
display projects correctly by using resource manager tree

hown3d:
## Description

The current implementation of the `project list` command does not return all projects a user has permission to see. It just returns the projects a user is directly a member of, not if he inherited the permissions from e.g. an organization or folder.

This PR uses the authorization and resourcemanager APIs to correctly query all projects a user has permission to see.
The implementation is a tree based data structure that uses recursion to view all projects of a folder, nested folders and projects directly under a organization.

For printing the Organization name, as well as the folder path is additionally printed. I decided against querying the resourcemanager API for project details as that returned in 500's from the API if queried to fast.

The existing flag filters do not apply as good anymore to the implementation, so they could be removed if the implementation itself is accepted.

Tests will be implemented once a first review has been done, as this also takes some time. I tested the change manually.

## Checklist

- [ ] Issue was linked above
- [ ] Unit tests are passing: `make test` (will be checked by CI)

### DIFF
--- a/docs/stackit_project_list.md
+++ b/docs/stackit_project_list.md
@@ -31,6 +31,7 @@ stackit project list [flags]
 ```
       --creation-time-after string   Filter by creation timestamp, in a date-time with the RFC3339 layout format, e.g. 2023-01-01T00:00:00Z. The list of projects that were created after the given timestamp will be shown
   -h, --help                         Help for "stackit project list"
+      --lifecycle-state string       Filter by lifecycle state (default "active")
       --limit int                    Maximum number of entries to list
       --member string                Filter by member. The list of projects of which the member is part of will be shown
       --page-size int                Number of items fetched in each API call. Does not affect the number of items in the command output (default 50)
--- a/internal/cmd/project/list/errors.go
+++ b/internal/cmd/project/list/errors.go
@@ -1,0 +1,19 @@
+package list
+
+import (
+	"errors"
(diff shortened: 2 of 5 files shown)


hown3d:
Would love to know your opinion :) @stackitcloud/developer-tools

vanto-tech49:
Is there any update on that PR from STACKIT side?

hown3d:
I'm happy to resolve the conflicts once we get some eyes on it :)

rubenhoenle (changes requested):
Hi @hown3d ,

sadly we won't merge this feature in it's current state. There are multiple reasons for it, in the end it all boils down to one problem: This logic should be implemented on API side instead of client side.

1. Performance and Rate limiting: This implementation does a lot of API requests to the resourcemanager to build the tree. This potentially puts more load than needed on the resourcemanager service. Furthermore (and this is more important) we're likely to run into rate limits sooner or later depending how many projects the user has access to.
2. A lot of edge cases: Users can have access to a project, but not to the folder it's located in and so on. This brings me to my next point:
3. There must be already such logic implemented somewhere. If I got it correctly the Portal is basically showing the same tree, so maybe we can talk to the Portal collegues to find out how they handle it.

Feel free to ping me via DM for further discussions, I would be happy to help bringing this feature forward.

Best regards
Ruben from STACKIT Developer Tools team

PR #1226, opened 2026-01-13
escape html encoding to marshal json output completely

GokceGK:
## Description

In the json output format, the used marshal function will escape some characters and replace with the unicode. 
This was causing some issues in the output of some commands like mongodbflex user creation uri. 

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/pkg/print/print.go
+++ b/internal/pkg/print/print.go
@@ -262,10 +262,10 @@ func (p *Printer) OutputResult(outputFormat string, output any, prettyOutputFunc
 		encoder.SetEscapeHTML(false)
 		encoder.SetIndent("", "  ")
 		err := encoder.Encode(output)
-		details := buffer.Bytes()
 		if err != nil {
 			return fmt.Errorf("marshal json: %w", err)
 		}
+		details := buffer.Bytes()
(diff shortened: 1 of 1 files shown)


internal/pkg/print/print.go
```diff
@@ -256,7 +257,12 @@ func (p *Printer) DebugInputModel(model any) {
 func (p *Printer) OutputResult(outputFormat string, output any, prettyOutputFunc func() error) error {
 	switch outputFormat {
 	case JSONOutputFormat:
-		details, err := json.MarshalIndent(output, "", "  ")
+		buffer := &bytes.Buffer{}
+		encoder := json.NewEncoder(buffer)
+		encoder.SetEscapeHTML(false)
+		encoder.SetIndent("", "  ")
+		err := encoder.Encode(output)
+		details := buffer.Bytes()
```
marceljk: I would move this line after the err != nil check. Not sure what will happen when the encode throws an error and we try to read the bytes of the buffer, but this can prevent maybe potential nil pointer expections

GokceGK: applies the requested change

Issue #1250, opened 2026-01-23
stackit auth login fails if something already runs on localhost:8000


KevinGimbel:
## Description

When running something on `localhost:8000` the `stackit auth login` command gets stuck.

## Steps to reproduce

1. Run something on `localhost:8000` 
2. Run `stackit auth login`
3. Browser opens, goes to the app already running on `localhost:8000`, stackit cli is stuck

## Actual behavior

The terminal reads the following and nothing happens:

```
Your browser has been opened to visit:

https://accounts.stackit.cloud/oauth/v2/authorize?<loads-of-stuff>&redirect_uri=http%3A%2F%2Flocalhost%3A8000&<more-stuff>

```

This doesn't always happen which is curious.

I tried to block the port with a simple python server by running `python3 -m http.server`, but then stackit CLI used another port (`8001`) as I would've expected.

When I run a static site generator such as [Zensical](https://zensical.org), the oauth flow gets stuck as described above, and no alternative port is chosen. 

## Expected behavior

I expect stackit cli to reliably fallback to a different port if `8000` is occupied.

## Environment
 - OS: MacOS 26
 - Version of STACKIT CLI (see `stackit --version`): `0.51.0`

PR #1251, opened 2026-01-23
auth: login: explicit check for ipv6 port bindings

svetlyopet:
## Description

Fixes [1250](https://github.com/stackitcloud/stackit-cli/issues/1250) 

This PR tries to address the issue of the net.Listen() not being able to see that the requested port is already taken on MacOS.

This happens specifically on MacOS, when there is already a service bound to a port on the IPv6 localhost interface, but the port is free on the IPv4 localhost interface. The net.Listen() method sees that the port is open on the 127.0.0.1 interface, and binds to that one. When the user gets redirected to localhost in the browser, it can redirect to either one of the services using the same port on the different interfaces(seems like localhost resolving to the IPv6 is preferred).

## Testing

I tested by running the following python service locally, and executing `stackit auth login` after that which replicates the issue and can be observed.

```python
#!/usr/bin/env python3
from http.server import HTTPServer, SimpleHTTPRequestHandler
import socket

HOST = "::1"
PORT = 8000

class IPv6HTTPServer(HTTPServer):
    address_family = socket.AF_INET6

server = IPv6HTTPServer((HOST, PORT), SimpleHTTPRequestHandler)
print(f"Serving on http://[{HOST}]:{PORT}")
server.serve_forever()
```

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/pkg/auth/user_login.go
+++ b/internal/pkg/auth/user_login.go
@@ -85,6 +85,8 @@ func AuthorizeUser(p *print.Printer, authConfig UserAuthConfig) error {
 	var redirectURL string
 	var listener net.Listener
 	var listenerErr error
+	var ipv6Listener net.Listener
+	var ipv6ListenerErr error
 	var port int
 	startingPort := defaultPort
 	portRange := configuredPortRange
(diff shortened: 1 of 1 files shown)


internal/pkg/auth/user_login.go
```diff
@@ -95,9 +95,15 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
...
-		portString := fmt.Sprintf(":%s", strconv.Itoa(port))
+		addr4 := fmt.Sprintf("127.0.0.1:%d", port)
+		addr6 := fmt.Sprintf("[::1]:%d", port)
 		p.Debug(print.DebugLevel, "trying to bind port %d for login redirect", port)
-		listener, listenerErr = net.Listen("tcp", portString)
+		ipv6Listener, ipv6ListenerErr := net.Listen("tcp6", addr6)
+		if ipv6ListenerErr != nil {
+			continue
+		}
+		_ = ipv6Listener.Close()
```
cgoetz-inovex: potential race condition:
If after this line another service binds to `[::1]:{port}` we have the same situation as before.

internal/pkg/auth/user_login.go
```diff
@@ -106,7 +112,7 @@ func AuthorizeUser(p *print.Printer, isReauthentication bool) error {
 		p.Debug(print.DebugLevel, "unable to bind port %d for login redirect: %s", port, listenerErr)
 	}
 	if listenerErr != nil {
```
marceljk, line 118: this should be extended with a check if ipv6ListenerErr is not nil. It's probably unlikely, but if someone uses already all the ports from 8000-8020, it will continue here without an error and try the authentication flow without an redirect url and the CLI will stop with a panic.
 
 I extended your python code, to simulate this case
 ```python
 #!/usr/bin/env python3
import threading
from http.server import HTTPServer, SimpleHTTPRequestHandler
import socket

HOST = "::1"
PORT = 8000


class IPv6HTTPServer(HTTPServer):
    address_family = socket.AF_INET6

for i in range(0, 21):
    server = IPv6HTTPServer((HOST, PORT + i), SimpleHTTPRequestHandler)
    print(f"Serving on http://[{HOST}]:{PORT+i}")
    t = threading.Thread(target=server.serve_forever)
    t.start()
 ```

internal/pkg/auth/user_login.go
svetlyopet: moved the ipv6Listener close call to after successfully binding on ipv4

internal/pkg/auth/user_login.go
svetlyopet: Added a check for ipv6ListenerErr with the same message for binding ports error, couldn't think of a more appropriate error message.

internal/pkg/auth/user_login.go
cgoetz-inovex: still racy, but the window is very small -> fine by me -> merging

marceljk:
Thanks! This will be part of our next release

PR #1023, opened 2025-10-10
feat(routing-table): provide rt and routes functionality

h3adex:
## Description

This PR adds functionality to list and describe routing tables, as well as perform full CRUD operations on routes within those tables. It does not include support for creating or attaching routing tables, as those operations are currently intended to be managed exclusively through Terraform. 

This implementation is primarily aimed at enabling users to inspect and debug routes created via Terraform. Once the routing table feature reaches GA, support for creating routing tables and attaching them to networks will be added to the CLI.

Screenshot 2025-10-10 at 16 22 40

## Checklist

- [x] Issue was linked above
- [x] Code format was applied: `make fmt`
- [x] Examples were added / adjusted (see e.g. [here](https://github.com/stackitcloud/stackit-cli/blob/ef291d1683ca5b0d719ec0a26ecb999a32685117/internal/cmd/ske/cluster/create/create.go#L49-L63))
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI)
- [x] Unit tests got implemented or updated
- [x] Unit tests are passing: `make test` (will be checked by CI)
- [x] No linter issues: `make lint` (will be checked by CI)

PR #1266, opened 2026-02-06
Feat/add rt support

h3adex:
## Description

reopened https://github.com/stackitcloud/stackit-cli/pull/1023

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/routingtable/route/create/create.go
+++ b/internal/cmd/routingtable/route/create/create.go
@@ -1,0 +1,268 @@
+package create
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/examples"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/flags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/services/iaas/client"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/types"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
+)
+
+const (
+	destinationTypeFlag  = "destination-type"
+	destinationValueFlag = "destination-value"
+	labelFlag            = "labels"
+	networkAreaIdFlag    = "network-area-id"
+	nextHopTypeFlag      = "nexthop-type"
+	nextHopValueFlag     = "nexthop-value"
+	organizationIdFlag   = "organization-id"
+	routingTableIdFlag   = "routing-table-id"
+
+	// Destination Type Constants
+	destTypeCIDRv4 = "cidrv4"
+	destTypeCIDRv6 = "cidrv6"
+
+	// NextHop Type Constants
+	nextHopTypeIPv4      = "ipv4"
+	nextHopTypeIPv6      = "ipv6"
+	nextHopTypeInternet  = "internet"
+	nextHopTypeBlackhole = "blackhole"
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	DestinationType  string
+	DestinationValue *string
+	Labels           *map[string]string
+	NetworkAreaId    string
+	NextHopType      string
+	NextHopValue     *string
+	OrganizationId   string
+	RoutingTableId   string
+}
+
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates a route in a routing-table",
+		Long:  "Creates a route in a routing-table.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample("Create a route with CIDRv4 destination and IPv4 nexthop",
+				`stackit routing-table route create  \ 
+--routing-table-id xxx --organization-id yyy --network-area-id zzz \
+--destination-type cidrv4 --destination-value <ipv4-cidr> \
+--nexthop-type ipv4 --nexthop-value <ipv4-address>`),
+
+			examples.NewExample("Create a route with CIDRv6 destination and IPv6 nexthop",
+				`stackit routing-table route create \
+--routing-table-id xxx --organization-id yyy --network-area-id zzz \
+--destination-type cidrv6 --destination-value <ipv6-cidr> \
+--nexthop-type ipv6 --nexthop-value <ipv6-address>`),
+
+			examples.NewExample("Create a route with CIDRv6 destination and Nexthop Internet",
+				`stackit routing-table route create \
+--routing-table-id xxx --organization-id yyy --network-area-id zzz \
+--destination-type cidrv6 --destination-value <ipv6-cidr> \
+--nexthop-type internet`),
+		),
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			ctx := context.Background()
+			model, err := parseInput(params.Printer, cmd, nil)
+			if err != nil {
+				return err
+			}
+
+			apiClient, err := client.ConfigureClient(params.Printer, params.CliVersion)
+			if err != nil {
+				return err
+			}
+
+			prompt := fmt.Sprintf("Are you sure you want to create a route for routing-table with id %q?", model.RoutingTableId)
+			err = params.Printer.PromptForConfirmation(prompt)
+			if err != nil {
+				return err
+			}
+
+			req, err := buildRequest(ctx, model, apiClient)
+			if err != nil {
+				return err
+			}
+
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("create route request failed: %w", err)
+			}
+
+			return outputResult(params.Printer, model.OutputFormat, resp.GetItems())
+		},
+	}
+	configureFlags(cmd)
+	return cmd
+}
+
+func configureFlags(cmd *cobra.Command) {
+	cmd.Flags().Var(flags.CIDRFlag(), destinationValueFlag, "Destination value")
+	cmd.Flags().String(nextHopValueFlag, "", "NextHop value")
+	cmd.Flags().Var(flags.UUIDFlag(), networkAreaIdFlag, "Network-Area ID")
+	cmd.Flags().Var(flags.UUIDFlag(), organizationIdFlag, "Organization ID")
+	cmd.Flags().Var(flags.UUIDFlag(), routingTableIdFlag, "Routing-Table ID")
+
+	cmd.Flags().Var(
+		flags.EnumFlag(true, "", destTypeCIDRv4, destTypeCIDRv6),
+		destinationTypeFlag,
+		"Destination type")
+
+	cmd.Flags().Var(
+		flags.EnumFlag(true, "", nextHopTypeIPv4, nextHopTypeIPv6, nextHopTypeInternet, nextHopTypeBlackhole),
+		nextHopTypeFlag,
+		"Next hop type")
+
+	cmd.Flags().StringToString(labelFlag, nil, "Key=value labels")
+
+	err := flags.MarkFlagsRequired(cmd, organizationIdFlag, networkAreaIdFlag, routingTableIdFlag, destinationTypeFlag, destinationValueFlag, nextHopTypeFlag)
+	cobra.CheckErr(err)
+}
+
+func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel, error) {
+	globalFlags := globalflags.Parse(p, cmd)
+
+	model := &inputModel{
+		GlobalFlagModel:  globalFlags,
+		DestinationType:  flags.FlagToStringValue(p, cmd, destinationTypeFlag),
+		DestinationValue: flags.FlagToStringPointer(p, cmd, destinationValueFlag),
+		Labels:           flags.FlagToStringToStringPointer(p, cmd, labelFlag),
+		NetworkAreaId:    flags.FlagToStringValue(p, cmd, networkAreaIdFlag),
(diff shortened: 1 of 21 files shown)


internal/cmd/network/create/create_test.go
```diff
@@ -469,6 +473,14 @@ func TestParseInput(t *testing.T) {
 			}),
 			isValid: true,
 		},
+		{
+			description: "routing-table id invalid",
```
rubenhoenle, line 477: since the `routing-table-id` flag is optional, could we please have a test case without the flag being set or being invalid?

internal/cmd/network/update/update_test.go
```diff
@@ -49,6 +50,7 @@ func fixtureFlagValues(mods ...func(flagValues map[string]string)) map[string]st
 		ipv6DnsNameServersFlag: "2001:4860:4860::8888,2001:4860:4860::8844",
 		ipv6GatewayFlag:        "2001:4860:4860::8888",
 		labelFlag:              "key=value",
+		routingTableIdFlag:     testRoutingTableId,
```
rubenhoenle, line 53: same here, routing table id flag is optional to there must be at least 1 test case IMO which doesn't have it set

internal/cmd/routingtable/create/create.go
```diff
@@ -0,0 +1,186 @@
...
+	return apiClient.AddRoutingTableToArea(
+		ctx,
+		model.OrganizationId,
+		model.NetworkAreaId,
+		model.Region,
+	).AddRoutingTableToAreaPayload(payload), nil
+}
+
+func outputResult(p *print.Printer, outputFormat string, routingTable *iaas.RoutingTable) error {
+	if routingTable == nil {
```
rubenhoenle, line 148: This condition can be moved into the callback func for the pretty output, so the JSON/YAML output is the actual API response

internal/cmd/routingtable/create/create.go
```diff
@@ -0,0 +1,186 @@
...
+		return fmt.Errorf("create routing-table response is empty")
+	}
+
+	if routingTable.Id == nil {
```
rubenhoenle, line 152: same here

internal/cmd/routingtable/route/update/update.go
```diff
@@ -0,0 +1,153 @@
...
+		RouteId:         routeId,
+		RoutingTableId:  flags.FlagToStringValue(p, cmd, routingTableIdFlag),
+	}
+
+	p.DebugInputModel(model)
+	return &model, nil
+}
+
+func outputResult(p *print.Printer, outputFormat, routingTableId, networkAreaId string, route *iaas.Route) error {
+	if route == nil {
```
rubenhoenle: check can be removed, is already present in the pretty output callback func

internal/cmd/routingtable/route/update/update.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("update route response is empty")
+	}
+
+	if route.Id == nil || *route.Id == "" {
```
rubenhoenle: check can be removed, is already present in the pretty output callback func

internal/cmd/routingtable/update/update.go
```diff
@@ -0,0 +1,169 @@
...
+	p.DebugInputModel(model)
+	return &model, nil
+}
+
+func outputResult(p *print.Printer, outputFormat, networkAreaId string, routingTable *iaas.RoutingTable) error {
+	if routingTable == nil {
+		return fmt.Errorf("update routing-table response is empty")
+	}
+
+	if routingTable.Id == nil {
```
rubenhoenle: check belongs into pretty output callback func, JSON/YAML output won't have problems with this being nil

internal/cmd/network/update/update_test.go
h3adex: Added a test case

internal/cmd/routingtable/create/create.go
```diff
@@ -0,0 +1,186 @@
...
+
+	return apiClient.AddRoutingTableToArea(
+		ctx,
+		model.OrganizationId,
+		model.NetworkAreaId,
+		model.Region,
+	).AddRoutingTableToAreaPayload(payload), nil
+}
+
+func outputResult(p *print.Printer, outputFormat string, routingTable *iaas.RoutingTable) error {
```
rubenhoenle, line 147: Recognized during testing: It would be new to me that we show tables for `create` commands. I would also like to stick to that, because I don't see any benefit for the user: He just entered all information when executing the `create` command, so why show it all again.

For create we usually just print something like `Created routing table with ID xxx in STACKIT Network Area xxx".

Could you elaborate why you've chosen to implement a table output here? 😅

rubenhoenle: In fact the current approach with the table is already questionable for me since the ID of the routing table I just created isn't visible anymore after I exited the pager showing the table output.

internal/cmd/routingtable/update/update.go
```diff
@@ -0,0 +1,169 @@
...
+				`Updates the name of a routing-table with ID "xxx" in organization with ID "yyy" and network-area with ID "zzz"`,
+				"$ stackit routing-table update xxx --name foo --organization-id yyy --network-area-id zzz",
+			),
+			examples.NewExample(
+				`Updates the description of a routing-table with ID "xxx" in organization with ID "yyy" and network-area with ID "zzz"`,
+				"$ stackit routing-table update xxx --description foo --organization-id yyy --network-area-id zzz",
+			),
+			examples.NewExample(
+				`Disables the dynamic routes of a routing-table with ID "xxx" in organization with ID "yyy" and network-area with ID "zzz"`,
+				"$ stackit routing-table update xxx --organization-id yyy --network-area-id zzz --disable-dynamic-routes",
```
rubenhoenle: Your examples are using a non-existent `--disable-dynamic-routes` flag.

Image

internal/pkg/services/routingtable/utils/utils.go
```diff
@@ -0,0 +1,66 @@
...
+			routeDetails.HopValue = utils.PtrString(route.Nexthop.NexthopIPv6.Value)
+		} else if route.Nexthop.NexthopInternet != nil {
+			routeDetails.HopType = utils.PtrString(route.Nexthop.NexthopInternet.Type)
+		} else if route.Nexthop.NexthopBlackhole != nil {
+			routeDetails.HopType = utils.PtrString(route.Nexthop.NexthopBlackhole.Type)
+		}
+	}
+
+	if route.Labels != nil && len(*route.Labels) > 0 {
+		var labels []string
```
rubenhoenle:
```suggestion
		routeDetails.Labels = JoinStringMap(tt.input, ": ", "\n")
```

This should do everything you need 😅 

https://github.com/stackitcloud/stackit-cli/blob/92c48cbeb188b18ec9e2ee6a75d0b842e7c703b9/internal/pkg/utils/strings_test.go#L34-L68

internal/cmd/routingtable/create/create.go
h3adex: Not sure why I did that to be honest. I implemented your suggestion

internal/cmd/routingtable/create/create.go
```diff
@@ -0,0 +1,160 @@
...
+
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates a routing-table",
+		Long:  "Creates a routing-table.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				"Create a routing-table with name `rt`",
```
marceljk: nitpick, in our existing examples we use always double quotes instead of backticks for strings in the example description. So you basically should swap here the use of (") and (`)
```suggestion
				`Create a routing-table with name "rt"`,
```
Please adjust it also in the following example descriptions

internal/cmd/routingtable/create/create_test.go
```diff
@@ -0,0 +1,348 @@
...
+			Verbosity: globalflags.VerbosityDefault,
+			Region:    testRegion,
+		},
+		OrganizationId: testOrgId,
+		NetworkAreaId:  testNetworkAreaId,
+		Name:           testRoutingTableName,
+		Description:    utils.Ptr(testRoutingTableDescription),
+		SystemRoutes:   testSystemRoutesFlag,
+		DynamicRoutes:  testDynamicRoutesFlag,
+		Labels:         utils.Ptr(*testLabels),
```
marceljk: can be simplified
```suggestion
		Labels:         testLabels,
```

internal/cmd/routingtable/delete/delete.go
```diff
@@ -0,0 +1,106 @@
...
+
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   fmt.Sprintf("delete %s", routingTableIdArg),
+		Short: "Deletes a routing-table",
+		Long:  "Deletes a routing-table",
+		Args:  args.SingleArg(routingTableIdArg, utils.ValidateUUID),
+		Example: examples.Build(
+			examples.NewExample(
+				`Deletes a a routing-table`,
```
marceljk:
```suggestion
				`Delete a routing-table with ID "xxx"`,
```

internal/cmd/routingtable/describe/describe_test.go
```diff
@@ -0,0 +1,229 @@
...
+			description: "invalid routing-table ID - format",
+			argValues:   []string{"invalid-uuid"},
+			flagValues:  fixtureFlagValues(),
+			isValid:     false,
+		},
+		{
+			description: "invalid routing-table ID - empty",
+			argValues:   []string{testRoutingTableId},
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[routingTableIdArg] = ""
```
marceljk: this test is failing because the flag `routingTableIdArg` doesn't exist. the routing-table id is actually set here in the argValues.

internal/cmd/routingtable/route/create/create.go
```diff
@@ -0,0 +1,277 @@
...
+		Labels:           flags.FlagToStringToStringPointer(p, cmd, labelFlag),
+		NetworkAreaId:    flags.FlagToStringValue(p, cmd, networkAreaIdFlag),
+		NextHopType:      flags.FlagToStringPointer(p, cmd, nextHopTypeFlag),
+		NextHopValue:     flags.FlagToStringPointer(p, cmd, nextHopValueFlag),
+		OrganizationId:   flags.FlagToStringValue(p, cmd, organizationIdFlag),
+		RoutingTableId:   flags.FlagToStringValue(p, cmd, routingTableIdFlag),
+	}
+
+	// Next Hop validation logic
+	switch strings.ToLower(*model.NextHopType) {
```
marceljk: I would set the type of NextHopType to `string` instead `*string`. The flag is marked as required so it should be always set. When it's set to a pointer string, it could more likely result somewhere in a nil pointer. With the string type we can prevent this.

internal/cmd/routingtable/route/create/create.go
```diff
@@ -0,0 +1,277 @@
...
+	case "internet", "blackhole":
```
marceljk: please add some consts for the different nexthopeTypes instead of writing them all as plain strings. this can prevent future issues if somewhere is a typo

internal/cmd/routingtable/route/create/create.go
```diff
@@ -0,0 +1,277 @@
...
+				routeDetails.CreatedAt,
+				routeDetails.UpdatedAt,
+			)
+		}
+		err := table.Display(p)
+		if err != nil {
+			return fmt.Errorf("render table: %w", err)
+		}
+		return nil
+	})
```
marceljk, line 267: Same as mentioned by Ruben in the routingtable create command, we usually don't print a table in the create command. Instead a small info message is enough, e.g. "Created route with ID xxx in routing table yyy"

internal/cmd/routingtable/route/create/create_test.go
```diff
@@ -0,0 +1,701 @@
...
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[labelFlag] = "key=value"
+			}),
+			expectedModel: fixtureInputModel(func(model *inputModel) {
+				model.Labels = utils.Ptr(map[string]string{"key": "value"})
+			}),
+			isValid: true,
+		},
+		{
+			description:   "argument value missing",
```
marceljk: This command actually doesn't support arguments. So if the argument would be really missing, it should pass
```suggestion
			description:   "argument value is empty string",
```

internal/cmd/routingtable/route/delete/delete.go
```diff
@@ -0,0 +1,111 @@
...
+				return err
+			}
+
+			// Configure API client
+			apiClient, err := client.ConfigureClient(params.Printer, params.CliVersion)
+			if err != nil {
+				return err
+			}
+
+			prompt := fmt.Sprintf("Are you sure you want to delete the route %q in routing-table %q for network-area-id %q?", model.RouteID, model.RoutingTableId, model.OrganizationId)
```
marceljk:
```suggestion
			prompt := fmt.Sprintf("Are you sure you want to delete the route %q in routing-table %q for network area id %q?", model.RouteID, model.RoutingTableId, model. NetworkAreaId)
```

internal/cmd/routingtable/describe/describe_test.go
h3adex: removed the test case

internal/cmd/routingtable/route/create/create.go
```diff
@@ -252,25 +260,8 @@ func outputResult(p *print.Printer, outputFormat string, routes []iaas.Route) er
...
-				routeDetails.HopValue,
-				routeDetails.Labels,
-				routeDetails.CreatedAt,
-				routeDetails.UpdatedAt,
-			)
-		}
-		err := table.Display(p)
-		if err != nil {
-			return fmt.Errorf("render table: %w", err)
+			p.Outputf("Created route with ID %q\n", utils.PtrString(route.Id))
```
h3adex, line 264: We iterate over it, but the response only has one item.

h3adex: Image

Issue #291, opened 2024-05-07
Windows package missing


sammetb:
We are working on Windows machines, not on MACs and not on Linux.
A installation package would be great to have.

vicentepinto98:
Hello @sammetb, we will support distribution with [WinGet](https://learn.microsoft.com/en-us/windows/package-manager/) in the next couple of weeks

PR #1311, opened 2026-02-23
feat: Document scoop as installation option for Windows

JorTurFer:
## Description

Document how to install the cli using [Scoop](https://scoop.sh/) on Windows

relates to https://github.com/stackitcloud/stackit-cli/issues/291

## Checklist

- [x] Issue was linked above

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -188,7 +188,22 @@ Alternatively, you can install via [Homebrew](https://brew.sh/) or refer to one 
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


JorTurFer:
> @JorTurFer, for my understanding: https://github.com/ScoopInstaller/Main/blob/master/bucket/stackit.json looks like `autoupdate` will automatically generate entries for new versions?

That's correct, if the cli follows the same semver approach (not changing to add more digits, with/without extra letters, etc), it'll be 100% transparent for this repo and their tool will manage the things automatically.

Other manual actions needed on their repo are things like binary name change or new architectures supported.

But in general, no further actions are needed just to update the versions there

PR #1312, opened 2026-02-24
feat(resourcemanager): List and describe organizations

Manuelvaas:
## Description

relates to STACKITCLI-325

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/organization/list/list.go
+++ b/internal/cmd/organization/list/list.go
@@ -92,10 +92,10 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel, 
 	globalFlags := globalflags.Parse(p, cmd)
 
 	limit := flags.FlagToInt64Pointer(p, cmd, limitFlag)
-	if limit != nil && *limit < 1 {
+	if limit != nil && (*limit < 1 || *limit > 100) {
 		return nil, &errors.FlagValidationError{
 			Flag:    limitFlag,
-			Details: "must be greater than 0",
+			Details: "must be between 0 and 100",
 		}
 	}
 


internal/cmd/organization/describe/describe.go
```diff
@@ -0,0 +1,119 @@
...
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	OrganizationId string
+}
+
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe",
+		Short: "Show a organization",
```
cgoetz-inovex: typo
```suggestion
		Short: "Show an organization",
```

internal/cmd/organization/describe/describe.go
```diff
@@ -0,0 +1,119 @@
...
+		Long:  "Show a organization.",
```
cgoetz-inovex:
```suggestion
		Long:  "Show an organization.",
```

internal/cmd/organization/list/list.go
```diff
@@ -0,0 +1,145 @@
...
+			model.Member = auth.GetProfileEmail(activeProfile)
+
+			// Configure API client
+			apiClient, err := client.ConfigureClient(params.Printer, params.CliVersion)
+			if err != nil {
+				return err
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
```
cgoetz-inovex, line 70: ListOrganizations takes a limit and offset: https://docs.api.stackit.cloud/documentation/resource-manager/version/unversioned#tag/Organizations/operation/get-all-organizations

min 0, max 100, default 50

I'd assume if I specify `--limit 110` on the command line, the API would just return the first 100. We should loop here and build requests with max limit and offset until the `--limit 110` flag is satisfied.

internal/cmd/organization/list/list.go
```diff
@@ -0,0 +1,145 @@
...
+
+	p.DebugInputModel(model)
+	return &model, nil
+}
+
+func buildRequest(ctx context.Context, model *inputModel, apiClient *resourcemanager.APIClient) resourcemanager.ApiListOrganizationsRequest {
+	req := apiClient.ListOrganizations(ctx)
+	req = req.Member(model.Member)
+	if model.Limit != nil {
+		req = req.Limit(float32(*model.Limit))
```
cgoetz-inovex, line 115: unrelated to your PR but maybe we should investigate why the generator uses f32 here. It also uses f32 for offset. If we try to use a very large offset we'll get funky results.

internal/cmd/organization/list/list.go
Manuelvaas: Thought about that as well, but the use case to have more than 100 organizations is not present.

internal/cmd/organization/list/list.go
Manuelvaas: It's not specified, what "number" the API uses, so the generator uses something generic. It should be updated in the API Specs.

internal/cmd/organization/list/list.go
cgoetz-inovex: Then I'd expect at least an input validation.
if the usecase is not present, why does the API support it?

PR #1363, opened 2026-04-09
refac(io) set IO once in main to allow overriding with in memory io i…

cgoetz-inovex:
…n tests

`print.Printer` had a reference to a `cobra.Command` for using its IO streams. Each Command also used a Printer, resulting in an awkward circular dependency.
Refactored Printer to use IO streams directly. When using the application these are set in `main`, when used in tests these can be set to `bytes.Buffer`s.
Also replaced usages of `os.Args` with just a string slice. Also set in `main`.
`cobra.Command`s `Args` and IO-streams are set in `NewRootCmd` with `traverseCommands`.
`CmdParams` also has an `fs.FS`, currently unused but will allow using the real FS during regular use and an in-memory-FS during tests.

This change prepares the application for integrative testing while keeping good isolation. Generally speaking the goal is to move all things with side effects into main (compare with https://grafana.com/blog/how-i-write-http-services-in-go-after-13-years/#func-main-only-calls-run)

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/pkg/print/print_test.go
+++ b/internal/pkg/print/print_test.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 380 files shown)


internal/pkg/print/print_test.go
```diff
@@ -724,9 +707,8 @@ func TestIsVerbosityDebug(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
-			cmd := &cobra.Command{}
+			// TODO ask team if we really need this test
```
marceljk: Some open TODOs in this file

marceljk (approved):

PR #1366, opened 2026-04-14
feat(secret flag) add reusable secret flag implementation, update guide

cgoetz-inovex:
Had to split implementation into two parts: `Set()` and `SecretFlagToSP`. Set is only called when an argument is specified on the command line. So moving the `PromptForPassword` logic into `Set` does not work.

I'm not sure if the current solution with a specialized `*ToStringPointer` func is the way to go. So I've only refactored `obs-credentials add` as an example.

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above


internal/pkg/flags/secret.go
```diff
@@ -0,0 +1,78 @@
...
+	f.value = value
+	return nil
+}
+
+func (f *secretFlag) Type() string {
+	return "string"
+}
+
+func (f *secretFlag) Usage() string {
+	return fmt.Sprintf("%s. Can be a string (deprecated) or a file path, if prefixed with '@' (example: @./secret.txt). Will be read from stdin when empty.", f.name)
```
marceljk: nitpick: I suggest to set the first letter of the flag name to uppercase to keep the docs consistent

marceljk (dismissed):

internal/pkg/flags/secret_test.go
```diff
@@ -0,0 +1,124 @@
...
+		want       *string
+		file       *testFile
+		stdin      string
+		wantErr    bool
+		wantStdErr string
+	}{
+		{
+			name:  "no value: prompts",
+			value: "",
+			want:  utils.Ptr("from stdin"),
```
marceljk, line 34:
```suggestion
			want:  new("from stdin"),
```

cgoetz-inovex: go 1.26 feature (https://go.dev/doc/go1.26#language), cli ist still on 1.25. When we upgrade to 1.26 I'd run `go fix` on the repo to autofix this.

internal/pkg/flags/secret.go
cgoetz-inovex: fixed

marceljk (approved):

PR #1392, opened 2026-04-29
feat(curl): add option to debug call

h3adex:
## Description

Small PR to add similar feature as `curl -v`

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/curl/curl.go
+++ b/internal/cmd/curl/curl.go
@@ -31,6 +31,7 @@ const (
 	includeResponseHeadersFlag = "include"
 	failOnHTTPErrorFlag        = "fail"
 	outputFileFlag             = "output"
+	verboseFlag                = "verbose"
 )
 
 const (
@@ -45,6 +46,7 @@ type inputModel struct {
 	IncludeResponseHeaders bool
 	FailOnHTTPError        bool
 	OutputFile             *string
+	Verbose                bool
 }
 
 func NewCmd(params *types.CmdParams) *cobra.Command {
@@ -69,6 +71,10 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				`Get all the DNS zones for project with ID xxx via GET request to https://dns.api.stackit.cloud/v1/projects/xxx/zones, with header "Authorization: Bearer yyy", fail if server returns error (such as 403 Forbidden)`,
 				`$ stackit curl https://dns.api.stackit.cloud/v1/projects/xxx/zones -X POST -H "Authorization: Bearer yyy" --fail`,
 			),
+			examples.NewExample(
+				"Get all the DNS zones via GET with detailed information about the request and the response",
+				"$ stackit curl https://dns.api.stackit.cloud/v1/projects/xxx/zones --verbose",
+			),
 		),
 		Args: args.SingleArg(urlArg, utils.ValidateURLDomain),
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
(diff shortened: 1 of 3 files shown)


internal/cmd/curl/curl.go
```diff
@@ -87,13 +93,24 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				return err
 			}
 
+			if model.Verbose {
+				requestDump, _ := httputil.DumpRequest(req, true)
+				params.Printer.Outputln(fmt.Sprintf("--- REQUEST ---\n%s", string(requestDump)))
```
cgoetz-inovex: When dumping a request/response with a binary body this has the chance to bring the terminal into a weird state.
But looking at `outputResponse` this doesn't seem to matter

internal/cmd/curl/curl.go
```diff
@@ -87,13 +93,24 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
 				Timeout: 30 * time.Second,
 			}
 			resp, err := client.Do(req)
 			if err != nil {
 				return fmt.Errorf("do request: %w", err)
 			}
+
+			if model.Verbose {
+				responseDump, _ := httputil.DumpResponse(resp, false)
+				params.Printer.Outputln(fmt.Sprintf("--- RESPONSE ---\n%s", string(responseDump)))
```
cgoetz-inovex: what's the difference between `stackit curl --include` and `stackit curl --verbose` for the response?

cgoetz-inovex (approved):

internal/cmd/curl/curl.go
h3adex: Debug information from --verbose (request and response headers) is now sent to stderr, while the actual API response (body) remains on stdout. This allows users to pipe or redirect the response body without debug logs interfering with the data.

Added >  (request) and <  (response) prefixes to verbose output and improved line-break handling for a cleaner debugging experience.

h3adex: The --include flag continues to prepend response headers to stdout

internal/cmd/curl/curl.go
h3adex: I don't think we can solve this problem for binaries.

h3adex:
Test commands:

```sh
#!/bin/bash

WORKING_URL="https://iaas.api.stackit.cloud/v2/networks/public-ip-ranges"

# (Debug on stderr, Body on stdout)
bin/stackit curl "$WORKING_URL" --verbose | jq

# both works together
bin/stackit curl "$WORKING_URL" --verbose --include

bin/stackit curl "$WORKING_URL" \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"name": "test-network"}' \
  --verbose > /dev/null

echo '{"name": "test-network-from-file"}' > payload.json
bin/stackit curl "$WORKING_URL" \
  -X POST \
  -H "Content-Type: application/json" \
  --data @payload.json \
  --verbose > /dev/null
rm payload.json

head -c 1024 /dev/urandom > test.bin
bin/stackit curl "$WORKING_URL" \
  -X POST \
  -H "Content-Type: application/octet-stream" \
  --data @test.bin \
  --verbose > /dev/null
rm test.bin
```

cgoetz-inovex (dismissed):

internal/cmd/curl/curl.go
rubenhoenle:
```suggestion
				params.Printer.Outputf("--- REQUEST ---\n%s\n", string(requestDump))
```

internal/cmd/curl/curl.go
rubenhoenle: same here

rubenhoenle (dismissed):

internal/cmd/curl/curl.go
h3adex: I've changed it to use curl -v `< & >` for request/response design

internal/cmd/curl/curl.go
h3adex: see above

internal/cmd/curl/curl.go
h3adex: this also respects stdrr stdout 

`bin/stackit curl "https://iaas.api.stackit.cloud/v2/networks/public-ip-ranges" -v`

PR #1399, opened 2026-05-07
feat(sfs): onboard snapshot policy commands

GokceGK:
## Description

relates to STACKITCLI-395

## Checklist

- [x] Issue was linked above
- [] Examples were added / adjusted (see e.g. [here](https://github.com/stackitcloud/stackit-cli/blob/ef291d1683ca5b0d719ec0a26ecb999a32685117/internal/cmd/ske/cluster/create/create.go#L49-L63))


internal/cmd/beta/sfs/snapshot-policy/list/list.go
```diff
@@ -0,0 +1,161 @@
...
+
+	p.DebugInputModel(model)
+	return &model, nil
+}
+
+func buildRequest(ctx context.Context, model *inputModel, apiClient *sfs.APIClient) sfs.ApiListSnapshotPoliciesRequest {
+	req := apiClient.DefaultAPI.ListSnapshotPolicies(ctx, model.ProjectId)
+	if model.Immutable {
+		req = req.Immutable(true)
+	}
```
cgoetz-inovex: The `immutable` parameter works on the API level as a tri-state-boolean. We get different results for `null|true|false`.
When sending an explicit `false` to the API, it will only return non-immutable policies. When `null` it will return all.
You can try this out with the tf provider:
- branch: feat/STACKITTPR-630_snapshot-policy-list-resource
- config:
```hcl
data "stackit_sfs_snapshot_policies" "ps" {
  project_id = var.project_id
  immutable = true
}

output "ps" {
  value = data.stackit_sfs_snapshot_policies.ps
}
```
try setting immutable to `false` and removing it completely.
- `terraform plan`

Currently the CLI implementation does not support listing only mutable policies.

Should we model the immutable flag as an enum flag with values like `all`, `immutable-only`, `mutable-only`?

GokceGK: yes, you are right. Setting the field to enum sounds good to me, very nice suggestion.

GokceGK: @rubenhoenle would that be okay for you? Or any other suggestions?

cgoetz-inovex (approved):

PR #1421, opened 2026-05-27
refac(cdn): Migrate to new SDK structure

Fyusel:
STACKITCLI-356

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/pkg/services/cdn/client/client.go
+++ b/internal/pkg/services/cdn/client/client.go
@@ -2,7 +2,7 @@ package client
 
 import (
 	"github.com/spf13/viper"
-	"github.com/stackitcloud/stackit-sdk-go/services/cdn"
+	cdn "github.com/stackitcloud/stackit-sdk-go/services/cdn/v1api"
 
 	"github.com/stackitcloud/stackit-cli/internal/pkg/config"
 	genericclient "github.com/stackitcloud/stackit-cli/internal/pkg/generic-client"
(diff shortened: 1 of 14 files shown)


docs/stackit_beta_cdn_distribution_create.md
```diff
@@ -50,7 +50,7 @@ stackit beta cdn distribution create [flags]
       --loki-username string                      Username for log sink
       --monthly-limit-bytes int                   Monthly limit in bytes for the CDN distribution
       --optimizer                                 Enable optimizer for the CDN distribution (paid feature).
-      --regions strings                           Regions in which content should be cached, multiple of: ["EU" "US" "AF" "SA" "ASIA"] (default [])
+      --regions strings                           Regions in which content should be cached, multiple of: ["EU" "US" "AF" "SA" "ASIA" "unknown_default_open_api"] (default [])
```
marceljk: I think we need a helper function, what avoids that the fallback enum is part of the input flag and description. Similar like this in Terraform 

https://github.com/stackitcloud/terraform-provider-stackit/blob/d28e06e85538cf7b4e756f22b0cc8a71cb8391af/stackit/internal/utils/utils.go#L142-L152

internal/pkg/services/cdn/client/client.go
```diff
@@ -2,7 +2,7 @@ package client
 
 import (
 	"github.com/spf13/viper"
-	"github.com/stackitcloud/stackit-sdk-go/services/cdn"
+	cdn "github.com/stackitcloud/stackit-sdk-go/services/cdn/v1api"
 
 	"github.com/stackitcloud/stackit-cli/internal/pkg/config"
 	genericclient "github.com/stackitcloud/stackit-cli/internal/pkg/generic-client"
```
marceljk, line 8: Can't add a comment in the correct line here, but in the ConfigureClient, the useRegion param needs to be set from true to false. Otherwise the commands won't work

marceljk (approved):

PR #1424, opened 2026-06-01
feat(auth): add Workload Identity Federation (OIDC) support

philross:
## Description

Adds Workload Identity Federation (OIDC) support so CI/CD pipelines can authenticate without long-lived service account key files. The CLI exchanges a short-lived OIDC token for a STACKIT access token using the go SDK's WIF flow.

- New `internal/pkg/auth/auth.go`  with env-var helpers (`STACKIT_USE_OIDC`, `STACKIT_SERVICE_ACCOUNT_EMAIL`, `STACKIT_SERVICE_ACCOUNT_FEDERATED_TOKEN`)
- Auto-detects token source:  env var → GitHub Actions → Azure Pipelines
- No credentials are written to disk in WIF mode
- Updated `AUTHENTICATION.md` with setup guide

relates to #1327

## Checklist

- [x] Issue was linked above
-  [x] Examples were added / adjusted (see e.g. [here](https://github.com/stackitcloud/stackit-cli/blob/ef291d1683ca5b0d719ec0a26ecb999a32685117/internal/cmd/ske/cluster/create/create.go#L49-L63))


philross:
Before I continue working on this feature: Is the introduction of environment variables (`STACKIT_USE_OIDC`, `STACKIT_SERVICE_ACCOUNT_EMAIL`, `STACKIT_SERVICE_ACCOUNT_FEDERATED_TOKEN`) mentioned in #1327  the right approach, or should CLI flags (e.g.` --use-oidc`) be added instead?

philross:
Not stale, @GokceGK could you help route this to the right person/team to get an initial response on the question above?

GokceGK:
Hi @philross, 

thanks for starting this contribution. I will check your question and get back to you asap. 

Regards,
Gökce from the STACKIT Developer Tools Team

GokceGK:
Hi @philross, 

you can use env variables, but please align them with the STACKIT Terraform Provider.

Regards,
Gökce from the STACKIT Developer Tools Team

philross:
Thanks @GokceGK,

According to https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/guides/workload_identity_federation, `STACKIT_USE_OIDC` and ` STACKIT_SERVICE_ACCOUNT_EMAIL` are already compliant with the STACKIT Terraform Provider convention. In the case of `STACKIT_SERVICE_ACCOUNT_FEDERATED_TOKEN`, the docs just mention using `service_account_federated_token` in the provider configuration.  But since the naming is consistent I think keeping `STACKIT_SERVICE_ACCOUNT_FEDERATED_TOKEN` should be fine.

philross:
Not stale

internal/cmd/auth/activate-service-account/activate_service_account_test.go
```diff
@@ -101,6 +101,17 @@ func TestParseInput(t *testing.T) {
 				model.OnlyPrintAccessToken = false
 			}),
 		},
+		{
+			description: "oidc_mode_no_key_required",
```
Fyusel: This name is confusing here: OIDC is not set but the name implies that it is set

philross: You are right, the test does not make too much sense as is. 
`TestParseInput` currently validates flag parsing only and does not consider environment variables, which would be needed to test the OIDC setup. Would you prefer that we extend this function/test scope to include env vars, or would it be better to keep `parseInput` focused on flags and add a separate OIDC-focused test in `activate_service_account_test.go` ?

Fyusel: I would keep _TestParseInput_ as is. You already added tests to _oidc_test.go_ to check the correct handling of the oidc flag, right?

internal/cmd/auth/activate-service-account/activate_service_account.go
```diff
@@ -59,13 +59,22 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				`Only print the corresponding access token by using the service account token. This access token can be stored as environment variable (STACKIT_ACCESS_TOKEN) in order to be used for all subsequent commands.`,
 				"$ stackit auth activate-service-account --service-account-token my-service-account-token --only-print-access-token",
 			),
+			examples.NewExample(
+				`Authenticate via Workload Identity Federation (OIDC) and print the short-lived access token. Set STACKIT_USE_OIDC=1 and STACKIT_SERVICE_ACCOUNT_EMAIL; no service account key file is required.`,
+				"$ STACKIT_USE_OIDC=1 STACKIT_SERVICE_ACCOUNT_EMAIL=ci@sa.stackit.cloud stackit auth activate-service-account --only-print-access-token",
+			),
 		),
```
Fyusel, line 68: I checked the TFP implementation against this implementation and found one thing which is different.

The TFP implementation allows to provide a `use_oidc` in the provider configuration. If this is set this has higher priority (than the environment variable).

See:  https://github.com/stackitcloud/terraform-provider-stackit/blob/main/docs/index.md#authentication
```
1. Explicit configuration, e.g. by setting the field `use_oidc` in the provider block (see example below)
2. Environment variable, e.g. by setting `STACKIT_USE_OIDC`
3. Credentials file
```

In order to use the same mechanism here we need to add a cli flag as well (if provided -> cli flag has prio, if not environment variable is taken).

philross: I added supported for `--use-oidc` and also `STACKIT_FEDERATED_TOKEN_FILE`

internal/cmd/auth/activate-service-account/activate_service_account_test.go
philross: Since we now have `--use-oidc` , adding test cases here makes more sense now.

internal/pkg/auth/oidc.go
```diff
@@ -0,0 +1,80 @@
...
+	"os"
+
+	"github.com/stackitcloud/stackit-sdk-go/core/oidcadapters"
+)
+
+const (
+	EnvUseOIDC                      = "STACKIT_USE_OIDC"
+	EnvServiceAccountEmail          = "STACKIT_SERVICE_ACCOUNT_EMAIL"
+	EnvServiceAccountFederatedToken = "STACKIT_SERVICE_ACCOUNT_FEDERATED_TOKEN" //nolint:gosec // linter false positive
+	EnvFederatedTokenFile           = "STACKIT_FEDERATED_TOKEN_FILE"
```
GokceGK: I receive linter errors for this line. We might need to add nolint comment also in here

internal/cmd/auth/activate-service-account/activate_service_account_test.go
```diff
@@ -80,8 +87,29 @@ func TestParseInput(t *testing.T) {
...
 			},
 		},
+		{
+			description: "use_oidc_true",
+			flagValues: fixtureFlagValues(func(flagValues map[string]string) {
+				flagValues[useOIDCFlag] = "true"
+			}),
+			isValid: true,
+			expectedModel: fixtureInputModel(func(model *inputModel) {
+				model.UseOIDC = boolPtr(true)
```
GokceGK: is there a reason why we dont use the utils.Ptr() instead of boolPtr

internal/pkg/auth/oidc_test.go
```diff
@@ -0,0 +1,203 @@
...
+		{
+			description: "uses env when override is nil",
+			envUseOIDC:  "1",
+			override:    nil,
+			expected:    true,
+		},
+		{
+			description: "override true wins over env false",
+			envUseOIDC:  "0",
+			override:    boolPtr(true),
```
GokceGK: same here. Cant we use utils.Ptr

internal/pkg/auth/oidc.go
philross: Fixed.

internal/cmd/auth/activate-service-account/activate_service_account_test.go
philross: No, I just missed that internal function. Fixed.

internal/pkg/auth/oidc_test.go
philross: Also fixed here.

Fyusel (approved):

GokceGK (approved):

PR #1433, opened 2026-06-09
feat(vpn): basic connection commands

cgoetz-inovex:
- implement create, describe, list, delete for vpn connection
- add helpers for string based enum flags
- make JoinStringPtr generic to accept string based enum slices

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/beta/vpn/connection/connection.go
+++ b/internal/cmd/beta/vpn/connection/connection.go
@@ -1,0 +1,31 @@
+package connection
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/vpn/connection/create"
+	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/vpn/connection/delete"
+	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/vpn/connection/describe"
+	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/vpn/connection/list"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/types"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+)
+
+func NewCmd(p *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "connection",
+		Short: "Provides functionality for VPN connections",
+		Long:  "Provides functionality for VPN connections.",
+		Args:  args.NoArgs,
+		Run:   utils.CmdHelp,
+	}
+	addSubcommands(cmd, p)
+	return cmd
+}
+
+func addSubcommands(cmd *cobra.Command, p *types.CmdParams) {
+	cmd.AddCommand(create.NewCmd(p))
+	cmd.AddCommand(delete.NewCmd(p))
+	cmd.AddCommand(describe.NewCmd(p))
+	cmd.AddCommand(list.NewCmd(p))
+}
--- a/internal/cmd/beta/vpn/connection/connection.go
+++ b/internal/cmd/beta/vpn/connection/connection.go
@@ -2,6 +2,7 @@ package connection
 
 import (
 	"github.com/spf13/cobra"
+
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/vpn/connection/create"
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/vpn/connection/delete"
 	"github.com/stackitcloud/stackit-cli/internal/cmd/beta/vpn/connection/describe"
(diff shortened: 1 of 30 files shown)


internal/cmd/beta/vpn/connection/create/create.go
```diff
@@ -0,0 +1,427 @@
...
+
+	err := flags.MarkFlagsRequired(
+		cmd,
+		gatewayIdFlag, displayNameFlag,
+		tunnel1RemoteAddressFlag,
+		tunnel1PreSharedKeyFlag,
+		tunnel1Phase1EncryptionAlgorithmsFlag.Name(), tunnel1Phase1IntegrityAlgorithmsFlag.Name(),
+		tunnel1Phase2EncryptionAlgorithmsFlag.Name(), tunnel1Phase2IntegrityAlgorithmsFlag.Name(),
+		tunnel2RemoteAddressFlag,
+		tunnel2PreSharedKeyFlag,
```
Fyusel, line 269: This is not marked as required in the API docs but the text in the API docs says it is required. What is true?
Will they update the API docs?

(same for tunnel1PreSharedKeyFlag)

internal/cmd/beta/vpn/connection/connection.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	addSubcommands(cmd, p)
+	return cmd
+}
+
+func addSubcommands(cmd *cobra.Command, p *types.CmdParams) {
+	cmd.AddCommand(create.NewCmd(p))
+	cmd.AddCommand(delete.NewCmd(p))
+	cmd.AddCommand(describe.NewCmd(p))
+	cmd.AddCommand(list.NewCmd(p))
```
Fyusel, line 32: status command is missing

internal/pkg/flags/string_enum.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func StringEnumDefaultValue[T ~string](value T) StringEnumFlagOption[T] {
+	return func(f *stringEnumFlag[T]) {
+		f.value = value
+		f.valueSet = true
+	}
+}
+
+func StringEnumFlag[T ~string](name string, possibleValues []T, docs string, opts ...StringEnumFlagOption[T]) *stringEnumFlag[T] {
```
Fyusel, line 35: There is a similar function under utils.go (based on the one of the TFP)
_func FormatPossibleValues(values ...string) []string {_

You can remove the one from utils.go and add a comment to this function here.

internal/cmd/beta/vpn/connection/create/create.go
```diff
@@ -0,0 +1,427 @@
...
+	Tunnel2Phase1IntegrityAlgorithms  []vpn.PhaseIntegrityAlgorithmsInner
+	Tunnel2Phase1RekeyTime            *int32
+	Tunnel2Phase2DhGroups             []vpn.PhaseDhGroupsInner
+	Tunnel2Phase2EncryptionAlgorithms []vpn.PhaseEncryptionAlgorithmsInner
+	Tunnel2Phase2IntegrityAlgorithms  []vpn.PhaseIntegrityAlgorithmsInner
+	Tunnel2Phase2RekeyTime            *int32
+	Tunnel2Phase2DpdAction            *vpn.TunnelConfigurationPhase2AllOfDpdAction
+	Tunnel2Phase2StartAction          *vpn.TunnelConfigurationPhase2AllOfStartAction
+	Tunnel2PreSharedKey               string
+	Tunnel2RemoteAddress              string
```
Fyusel: Is it possible to use a _tunnelModel_ here?
Both blocks are the same, aren't they?

e.g. something like:
```
Tunnel1 tunnelInputModel
Tunnel2 tunnelInputModel
```

internal/pkg/flags/string_enum.go
cgoetz-inovex: If these two new StringEnum*Flag thingies make it through review, I'd the replace all usages of the old non generic enum flags with this one. At a first glance, it looks like FormatPossibleValues would then be unused.

internal/cmd/beta/vpn/connection/create/create.go
cgoetz-inovex: From the docs:
> A Pre-Shared Key for authentication. Required in create-requests, optional in update-requests and omitted in every response.

For me it looks like they reuse the same component for POST, PUT and GET and use docs for the differences. I've messaged them with the suggestion to introduce separate components to allow to model this in the schema. Same for the differences between policy-, route- and bgp gateway based connections.
But no idea if this will be done (no answer as of today).

internal/cmd/beta/vpn/connection/connection.go
cgoetz-inovex: good catch, thanks

internal/cmd/beta/vpn/connection/create/create.go
cgoetz-inovex: good idea this should let us reuse some code in `buildRequest`

Fyusel (dismissed):

Fyusel (approved):

PR #1436, opened 2026-06-10
fix(iaas): update examples and fix the payload for boot volumes

GokceGK:
## Description

relates to STACKITCLI-404

## Checklist

- [x] Issue was linked above


internal/cmd/server/create/create.go
```diff
@@ -194,6 +182,11 @@ func configureFlags(cmd *cobra.Command) {
 	cmd.MarkFlagsMutuallyExclusive(imageIdFlag, bootVolumeSourceTypeFlag)
 	cmd.MarkFlagsMutuallyExclusive(networkIdFlag, networkInterfaceIdsFlag)
 	cmd.MarkFlagsOneRequired(networkIdFlag, networkInterfaceIdsFlag)
+
+	// hide image-id flag from help command to prevent users from using it alone which leads to errors
```
marceljk: nitpick: I would add here the reason, so in case someone comes across this in the future, it's clear what the limitation is
```suggestion
	// hide image-id flag from help command to prevent users from using it alone which leads to errors, because it works only with small images
```

marceljk (approved):

PR #1454, opened 2026-07-02
SDK migration Secrets-Manager

SerseusWasTaken:
## Description
Updates the Secrets-Manager SDK

relates to STACKITCLI-377

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/secrets-manager/instance/list/list.go
+++ b/internal/cmd/secrets-manager/instance/list/list.go
@@ -7,7 +7,7 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/pkg/types"
 
 	"github.com/spf13/cobra"
-	"github.com/stackitcloud/stackit-sdk-go/services/secretsmanager"
+	secretsmanager "github.com/stackitcloud/stackit-sdk-go/services/secretsmanager/v1api"
 
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
@@ -18,7 +18,6 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/pkg/projectname"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/services/secrets-manager/client"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/tables"
-	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
 )
 
 const (
@@ -67,7 +66,7 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				return fmt.Errorf("get Secrets Manager instances: %w", err)
 			}
 
-			if resp.Instances == nil || len(*resp.Instances) == 0 {
+			if len(resp.Instances) == 0 {
 				projectLabel, err := projectname.GetProjectName(ctx, params.Printer, params.CliVersion, cmd)
 				if err != nil {
 					params.Printer.Debug(print.ErrorLevel, "get project name: %v", err)
@@ -76,14 +75,13 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				params.Printer.Info("No instances found for project %q\n", projectLabel)
 				return nil
 			}
-			instances := *resp.Instances
 
 			// Truncate output
-			if model.Limit != nil && len(instances) > int(*model.Limit) {
-				instances = instances[:*model.Limit]
+			if model.Limit != nil && len((resp.Instances)) > int(*model.Limit) {
+				(resp.Instances) = resp.Instances[:*model.Limit]
 			}
 
-			return outputResult(params.Printer, model.OutputFormat, instances)
+			return outputResult(params.Printer, model.OutputFormat, (resp.Instances))
 		},
 	}
 
@@ -119,7 +117,7 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel, 
 }
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *secretsmanager.APIClient) secretsmanager.ApiListInstancesRequest {
-	req := apiClient.ListInstances(ctx, model.ProjectId)
+	req := apiClient.DefaultAPI.ListInstances(ctx, model.ProjectId)
 	return req
 }
 
@@ -130,10 +128,10 @@ func outputResult(p *print.Printer, outputFormat string, instances []secretsmana
 		for i := range instances {
 			instance := instances[i]
 			table.AddRow(
-				utils.PtrString(instance.Id),
-				utils.PtrString(instance.Name),
-				utils.PtrString(instance.State),
-				utils.PtrString(instance.SecretCount),
+				instance.Id,
+				instance.Name,
+				instance.State,
+				instance.SecretCount,
 			)
 		}
 		err := table.Display(p)
--- a/internal/cmd/secrets-manager/instance/list/list.go
+++ b/internal/cmd/secrets-manager/instance/list/list.go
@@ -77,11 +77,11 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 			}
 
 			// Truncate output
-			if model.Limit != nil && len((resp.Instances)) > int(*model.Limit) {
-				(resp.Instances) = resp.Instances[:*model.Limit]
+			if model.Limit != nil && len(resp.Instances) > int(*model.Limit) {
+				resp.Instances = resp.Instances[:*model.Limit]
 			}
 
-			return outputResult(params.Printer, model.OutputFormat, (resp.Instances))
+			return outputResult(params.Printer, model.OutputFormat, resp.Instances)
 		},
 	}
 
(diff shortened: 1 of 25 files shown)


internal/cmd/secrets-manager/instance/list/list.go
```diff
@@ -76,14 +75,13 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
 
 			// Truncate output
-			if model.Limit != nil && len(instances) > int(*model.Limit) {
-				instances = instances[:*model.Limit]
+			if model.Limit != nil && len((resp.Instances)) > int(*model.Limit) {
+				(resp.Instances) = resp.Instances[:*model.Limit]
 			}
 
-			return outputResult(params.Printer, model.OutputFormat, instances)
+			return outputResult(params.Printer, model.OutputFormat, (resp.Instances))
```
cgoetz-inovex: superfluous parens around `resp.Instances`

internal/cmd/secrets-manager/instance/update/update.go
```diff
@@ -162,37 +165,39 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
-			KeyRingId:           model.KmsKeyringId,
-			KeyVersion:          model.KmsKeyVersion,
-			ServiceAccountEmail: model.KmsServiceAccountEmail,
+		payload := secretsmanager.UpdateInstancePayload{
+			Name: *model.InstanceName,
 		}
-	}
 
-	req = req.UpdateInstancePayload(payload)
+		if model.KmsKeyId != nil && model.KmsKeyringId != nil && model.KmsKeyVersion != nil && model.KmsServiceAccountEmail != nil {
```
cgoetz-inovex: nitpick: conditional inconsistent with `create`.  In this file we configure the flags:
```
	cmd.MarkFlagsRequiredTogether(kmsKeyIdFlag, kmsKeyringIdFlag, kmsKeyVersionFlag, kmsServiceAccountEmailFlag)
```
we could do it here like in create and only check for `model.KmsKeyId != nil` or also change `create`

internal/cmd/secrets-manager/instance/update/update.go
```diff
@@ -162,37 +165,39 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
+				KeyVersion:          *model.KmsKeyVersion,
+				ServiceAccountEmail: *model.KmsServiceAccountEmail,
+			}
+		}
+		req = req.UpdateInstancePayload(payload)
 
-	return req
+		return req, nil
+	}
+	return secretsmanager.ApiUpdateInstanceRequest{}, fmt.Errorf("instanceName must not be null")
```
cgoetz-inovex: nitpick: imo it would be more idiomatic to flip the conditional at the top and return early with this error if `InstanceName` is nil.
See: https://go.dev/doc/effective_go#if:~:text=This%20is%20an,no%20else%20statements.

internal/cmd/secrets-manager/instance/update/update_test.go
```diff
@@ -384,7 +386,7 @@ func TestBuildUpdateInstanceRequest(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
-			request := buildUpdateInstanceRequest(testCtx, tt.model, testClient)
+			request, _ := buildUpdateInstanceRequest(testCtx, tt.model, testClient)
```
cgoetz-inovex:
```suggestion
			request, err := buildUpdateInstanceRequest(testCtx, tt.model, testClient)
			if (err != nil) != tt.wantErr {
				t.Fatalf("buildUpdateInstanceRequest() error = %v, wantErr %v", err, tt.wantErr)
			}
```

with extended definition of test:
```go
	tests := []struct {
		description     string
		model           *inputModel
		expectedRequest secretsmanager.ApiUpdateInstanceRequest
		wantErr         bool
	}{
```

and new test case:
```go
		{
			description: "nil instance name",
			model: fixtureInputModel(func(model *inputModel) {
				model.InstanceName = nil
			}),
			wantErr: true,
		},
```

internal/cmd/secrets-manager/user/create/create.go
```diff
@@ -119,10 +119,18 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel,
...
+		description = *model.Description
+	}
+	if model.Write != nil {
+		write = *model.Write
+	}
 	req = req.CreateUserPayload(secretsmanager.CreateUserPayload{
-		Description: model.Description,
-		Write:       model.Write,
+		Description: description,
+		Write:       write,
```
cgoetz-inovex: flags configured as:
```go
	cmd.Flags().String(descriptionFlag, "", "A user chosen description to differentiate between multiple users")
	cmd.Flags().Bool(writeFlag, false, "User write access to the secrets engine. If unset, user is read-only")
```

When executing test case 'no description specified' or 'no write flag given' we get a pointer to `""` or `false` in `model.Description/Write`.
So these conditional assignements will never be skipped.

Suggesting to change the definition of model to (removing pointers):
```go
type inputModel struct {
	*globalflags.GlobalFlagModel

	InstanceId  string
	Description string
	Write       bool
}
```
Then we can remove the `utils.Ptr` calls in `parseInput`:
```go
		Description:     flags.FlagToStringValue(p, cmd, descriptionFlag),
		Write:           flags.FlagToBoolValue(p, cmd, writeFlag),

```

and finally we can leave this block as it were.

internal/cmd/secrets-manager/user/list/list.go
```diff
@@ -69,23 +68,22 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
 				params.Printer.Info("No users found for instance %q\n", instanceLabel)
 				return nil
 			}
-			users := *resp.Users
 
 			// Truncate output
-			if model.Limit != nil && len(users) > int(*model.Limit) {
-				users = users[:*model.Limit]
+			if model.Limit != nil && len((resp.Users)) > int(*model.Limit) {
+				(resp.Users) = resp.Users[:*model.Limit]
```
cgoetz-inovex: nitpick: superfluous parens around `resp.Users`

internal/pkg/services/secrets-manager/utils/utils.go
```diff
@@ -4,38 +4,38 @@ import (
...
-	"github.com/stackitcloud/stackit-sdk-go/services/secretsmanager"
+	secretsmanager "github.com/stackitcloud/stackit-sdk-go/services/secretsmanager/v1api"
 )
 
 type SecretsManagerClient interface {
-	GetInstanceExecute(ctx context.Context, projectId, instanceId string) (*secretsmanager.Instance, error)
-	GetUserExecute(ctx context.Context, projectId string, instanceId string, userId string) (*secretsmanager.User, error)
+	GetInstance(ctx context.Context, projectId, instanceId string) secretsmanager.ApiGetInstanceRequest
+	GetUser(ctx context.Context, projectId string, instanceId string, userId string) secretsmanager.ApiGetUserRequest
 }
```
cgoetz-inovex, line 13: another option would be to remove this interface and use `secretsmanager.DefaultAPI` in the methods down below. But I'm not sure if this is truly better than this approach.

internal/pkg/services/secrets-manager/utils/utils_test.go
```diff
@@ -69,12 +74,12 @@ func TestGetInstanceName(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
-			client := &secretsManagerClientMocked{
+			mockOptions := &apiClientMockOptions{
 				getInstanceFails: tt.getInstanceFails,
 				getInstanceResp:  tt.getInstanceResp,
 			}
 
-			output, err := GetInstanceName(context.Background(), client, testProjectId, testInstanceId)
+			output, err := GetInstanceName(context.Background(), newAPIClient(*mockOptions).DefaultAPI, testProjectId, testInstanceId)
```
cgoetz-inovex: you can save some pointer things here
```suggestion
			mockOptions := apiClientMockOptions{
getInstanceFails: tt.getInstanceFails,
getInstanceResp:  tt.getInstanceResp,
}

			output, err := GetInstanceName(context.Background(), newAPIClient(mockOptions).DefaultAPI, testProjectId, testInstanceId)
```

cgoetz-inovex (dismissed):

internal/cmd/secrets-manager/user/create/create.go
SerseusWasTaken: Yes, that is true. I just wanted the code to be more defensive to make sure the function cannot run into a nil pointer dereference. 
I intentionally left the flags as pointers because they are not mandatory, and i've read that we may want mandatory flags to be values and optional ones pointers in another PR. But i agree that there is practically no reason for these flags to be pointers.  
I'm happy to change these to values.

internal/cmd/secrets-manager/instance/update/update.go
SerseusWasTaken: Ah, good point. I will also update the condition in `create`

internal/pkg/services/secrets-manager/utils/utils.go
SerseusWasTaken: This decision should probably be made in a proper discussion, since this pattern is used in multiple places. So i would suggest to keep the status-quo here

cgoetz-inovex (approved):

PR #1457, opened 2026-07-06
feat(ske): Check service enablement only in error case

Fyusel:
STACKITCLI-287

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/ske/cluster/create/create.go
+++ b/internal/cmd/ske/cluster/create/create.go
@@ -95,21 +95,20 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				return err
 			}
 
-			// Check if the project is enabled before trying to create
-			enabled, err := serviceEnablementUtils.ProjectEnabled(ctx, serviceEnablementApiClient, model.ProjectId, model.Region)
-			if err != nil {
-				return err
-			}
-			if !enabled {
-				return &errors.ServiceDisabledError{
-					Service: "ske",
-				}
-			}
-
 			// Check if cluster exists
 			exists, err := skeUtils.ClusterExists(ctx, apiClient.DefaultAPI, model.ProjectId, model.Region, model.ClusterName)
 			if err != nil {
-				return err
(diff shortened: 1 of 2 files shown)


internal/cmd/ske/cluster/create/create.go
```diff
@@ -95,20 +95,19 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
+				enabled, err := serviceEnablementUtils.ProjectEnabled(ctx, serviceEnablementApiClient, model.ProjectId, model.Region)
+				if err != nil {
+					return fmt.Errorf("check if project is enabled failed: %w", err)
+				}
+				if !enabled {
+					return &errors.ServiceDisabledError{
+						Service: "ske",
+					}
+				}
 				return err
```
cgoetz-inovex: it's a bit late, but wouldn't this hide the original error from line 99 (`ClusterExists`) when service is enabled and the call to `ProjectEnabled` would return a nil error?
So sth. like
```go
			exists, err := skeUtils.ClusterExists(ctx, apiClient.DefaultAPI, model.ProjectId, model.Region, model.ClusterName)
            // err is now sth. like 'server is down'
			if err != nil {
				// Check if the project is enabled
				enabled, err := serviceEnablementUtils.ProjectEnabled(ctx, serviceEnablementApiClient, model.ProjectId, model.Region)
                 // let's assume call succeeds, enabled=true and err=nil now
				if err != nil {
					return fmt.Errorf("check if project is enabled failed: %w", err)
				}
				if !enabled {
					return &errors.ServiceDisabledError{
						Service: "ske",
					}
				}
				return err
                // we return nil here instead of 'server is down'
```

Fyusel: Good finding! Thanks!

Fyusel: Updated

cgoetz-inovex (approved):

PR #1467, opened 2026-07-09
feat(mongodbflex): migrate to v2 API

SerseusWasTaken:
## Description
Migrates mongodbflex to V2 API

relates to STACKITCLI-379

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/mongodbflex/backup/describe/describe.go
+++ b/internal/cmd/mongodbflex/backup/describe/describe.go
@@ -119,10 +119,10 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *mongodbflex
 }
 
 func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup *mongodbflex.Backup) error {
-	if backup == nil {
-		return fmt.Errorf("API response is nil")
-	}
 	return p.OutputResult(outputFormat, backup, func() error {
+		if backup == nil {
+			return fmt.Errorf("API response is nil")
+		}
 		table := tables.NewTable()
 		table.AddRow("ID", utils.PtrString(backup.Id))
 		table.AddSeparator()
--- a/internal/cmd/mongodbflex/backup/describe/describe.go
+++ b/internal/cmd/mongodbflex/backup/describe/describe.go
@@ -119,9 +119,8 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *mongodbflex
 }
 
 func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup *mongodbflex.Backup) error {
-	// allows passing empty backups but avoids having to pass large backup by value
 	if backup == nil {
-		backup = &mongodbflex.Backup{}
+		return fmt.Errorf("API response is nil")
 	}
 	return p.OutputResult(outputFormat, backup, func() error {
 		table := tables.NewTable()
--- a/internal/cmd/mongodbflex/backup/describe/describe.go
+++ b/internal/cmd/mongodbflex/backup/describe/describe.go
@@ -7,7 +7,7 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/pkg/types"
 
 	"github.com/spf13/cobra"
-	"github.com/stackitcloud/stackit-sdk-go/services/mongodbflex"
+	mongodbflex "github.com/stackitcloud/stackit-sdk-go/services/mongodbflex/v2api"
 
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
@@ -61,7 +61,7 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				return err
 			}
 
-			instanceLabel, err := mongoUtils.GetInstanceName(ctx, apiClient, model.ProjectId, model.InstanceId, model.Region)
+			instanceLabel, err := mongoUtils.GetInstanceName(ctx, apiClient.DefaultAPI, model.ProjectId, model.InstanceId, model.Region)
 			if err != nil {
 				params.Printer.Debug(print.ErrorLevel, "get instance name: %v", err)
 				instanceLabel = model.InstanceId
@@ -75,13 +75,13 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				return fmt.Errorf("describe backup for MongoDB Flex instance: %w", err)
 			}
 
-			restoreJobs, err := apiClient.ListRestoreJobs(ctx, model.ProjectId, model.InstanceId, model.Region).Execute()
+			restoreJobs, err := apiClient.DefaultAPI.ListRestoreJobs(ctx, model.ProjectId, model.InstanceId, model.Region).Execute()
 			if err != nil {
 				return fmt.Errorf("get restore jobs for MongoDB Flex instance %q: %w", instanceLabel, err)
 			}
 
 			restoreJobState := mongoUtils.GetRestoreStatus(model.BackupId, restoreJobs)
-			return outputResult(params.Printer, model.OutputFormat, restoreJobState, *resp.Item)
+			return outputResult(params.Printer, model.OutputFormat, restoreJobState, resp.Item)
 		},
 	}
 	configureFlags(cmd)
@@ -114,11 +114,15 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
 }
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *mongodbflex.APIClient) mongodbflex.ApiGetBackupRequest {
-	req := apiClient.GetBackup(ctx, model.ProjectId, model.InstanceId, model.BackupId, model.Region)
+	req := apiClient.DefaultAPI.GetBackup(ctx, model.ProjectId, model.InstanceId, model.BackupId, model.Region)
 	return req
 }
 
-func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup mongodbflex.Backup) error {
+func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup *mongodbflex.Backup) error {
+	// allows passing empty backups but avoids having to pass large backup by value
+	if backup == nil {
+		backup = &mongodbflex.Backup{}
+	}
 	return p.OutputResult(outputFormat, backup, func() error {
 		table := tables.NewTable()
 		table.AddRow("ID", utils.PtrString(backup.Id))
--- a/internal/cmd/mongodbflex/backup/restore-jobs/restore_jobs.go
+++ b/internal/cmd/mongodbflex/backup/restore-jobs/restore_jobs.go
@@ -7,7 +7,7 @@ import (
 	"github.com/stackitcloud/stackit-cli/internal/pkg/types"
 
 	"github.com/spf13/cobra"
-	"github.com/stackitcloud/stackit-sdk-go/services/mongodbflex"
+	mongodbflex "github.com/stackitcloud/stackit-sdk-go/services/mongodbflex/v2api"
 
 	"github.com/stackitcloud/stackit-cli/internal/pkg/args"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/errors"
@@ -63,7 +63,7 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				return err
 			}
 
-			instanceLabel, err := mongodbflexUtils.GetInstanceName(ctx, apiClient, model.ProjectId, *model.InstanceId, model.Region)
+			instanceLabel, err := mongodbflexUtils.GetInstanceName(ctx, apiClient.DefaultAPI, model.ProjectId, *model.InstanceId, model.Region)
 			if err != nil {
 				params.Printer.Debug(print.ErrorLevel, "get instance name: %v", err)
 				instanceLabel = *model.InstanceId
@@ -75,11 +75,11 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("get restore jobs for MongoDB Flex instance %q: %w", instanceLabel, err)
 			}
-			if resp.Items == nil || len(*resp.Items) == 0 {
+			if len(resp.Items) == 0 {
 				cmd.Printf("No restore jobs found for instance %q\n", instanceLabel)
 				return nil
 			}
-			restoreJobs := *resp.Items
+			restoreJobs := resp.Items
 
 			// Truncate output
 			if model.Limit != nil && len(restoreJobs) > int(*model.Limit) {
@@ -127,7 +127,7 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel, 
 }
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *mongodbflex.APIClient) mongodbflex.ApiListRestoreJobsRequest {
-	req := apiClient.ListRestoreJobs(ctx, model.ProjectId, *model.InstanceId, model.Region)
+	req := apiClient.DefaultAPI.ListRestoreJobs(ctx, model.ProjectId, *model.InstanceId, model.Region)
 	return req
 }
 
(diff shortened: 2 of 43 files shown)


internal/cmd/mongodbflex/backup/describe/describe.go
```diff
@@ -114,11 +114,15 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *mongodbflex.APIClient) mongodbflex.ApiGetBackupRequest {
-	req := apiClient.GetBackup(ctx, model.ProjectId, model.InstanceId, model.BackupId, model.Region)
+	req := apiClient.DefaultAPI.GetBackup(ctx, model.ProjectId, model.InstanceId, model.BackupId, model.Region)
 	return req
 }
 
-func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup mongodbflex.Backup) error {
+func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup *mongodbflex.Backup) error {
+	// allows passing empty backups but avoids having to pass large backup by value
```
rubenhoenle: Don't quite get the intention of this change here tbh 😅 Can you explain it please?

Even if we decide to keep it, can this be moved into the callback for the pretty output in the `p.OutputResult` call?

internal/cmd/mongodbflex/backup/restore-jobs/restore_jobs.go
```diff
@@ -75,11 +75,11 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("get restore jobs for MongoDB Flex instance %q: %w", instanceLabel, err)
 			}
-			if resp.Items == nil || len(*resp.Items) == 0 {
+			if len(resp.Items) == 0 {
 				cmd.Printf("No restore jobs found for instance %q\n", instanceLabel)
```
rubenhoenle: You can please quick-fix this here directly with the migration: We have an invalid JSON/YAML output here.

See https://github.com/stackitcloud/stackit-cli/issues/893 (and the linked PRs in it) for reference.

internal/cmd/mongodbflex/instance/create/create_test.go
```diff
@@ -481,13 +480,13 @@ func TestBuildRequest(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
-			client := &mongoDBFlexClientMocked{
+			settings := mockSettings{
```
rubenhoenle: Wouldn't it just be an option to have a `mockSettings` field as part of the table test struct at that point? 😅

internal/cmd/mongodbflex/instance/create/create_test.go
```diff
@@ -496,8 +495,9 @@ func TestBuildRequest(t *testing.T) {
 			}
 
 			diff := cmp.Diff(request, tt.expectedRequest,
-				cmp.AllowUnexported(tt.expectedRequest),
+				cmp.AllowUnexported(tt.expectedRequest, mongodbflex.DefaultAPIService{}),
 				cmpopts.EquateComparable(testCtx),
+				cmpopts.IgnoreFields(tt.expectedRequest, "ApiService"),
```
rubenhoenle, line 506: Isn't this redundant to this line above?

```golang
cmp.AllowUnexported(tt.expectedRequest, mongodbflex.DefaultAPIService{}),
```

internal/cmd/mongodbflex/instance/describe/describe.go
```diff
@@ -121,7 +121,7 @@ func outputResult(p *print.Printer, outputFormat string, instance *mongodbflex.I
 		table.AddRow("VERSION", utils.PtrString(instance.Version))
 		table.AddSeparator()
 		if instance.HasAcl() {
-			aclsArray := *instance.Acl.Items
+			aclsArray := instance.Acl.Items
 			acls := strings.Join(aclsArray, ",")
```
rubenhoenle:
```suggestion
			acls := strings.Join(instance.Acl.Items, ",")
```

Why create a variable here at all? ^^

internal/cmd/mongodbflex/options/options.go
```diff
@@ -223,7 +223,8 @@ func buildFlavorsTable(flavors []mongodbflex.InstanceFlavor) tables.Table {
 			utils.PtrString(f.Cpu),
 			utils.PtrString(f.Memory),
 			utils.PtrString(f.Description),
-			utils.PtrString(f.Categories),
+			//TODO: check if this is ok
```
rubenhoenle: resolve open TODO

internal/cmd/mongodbflex/options/options.go
rubenhoenle:
```suggestion
	for _, storageClass := range storagesResp.StorageClasses {
```

Saves you the line below and one above 😉

internal/cmd/mongodbflex/options/options_test.go
```diff
@@ -253,13 +255,13 @@ func TestBuildAndExecuteRequest(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
 			params := testparams.NewTestParams()
-			client := &mongoDBFlexClientMocked{
+			settings := mockClientSettings{
```
rubenhoenle: same here

internal/cmd/mongodbflex/user/create/create.go
```diff
@@ -152,7 +152,8 @@ func outputResult(p *print.Printer, outputFormat, instanceLabel string, user *mo
 		p.Outputf("Created user for instance %q. User ID: %s\n\n", instanceLabel, utils.PtrString(user.Id))
 		p.Outputf("Username: %s\n", utils.PtrString(user.Username))
 		p.Outputf("Password: %s\n", utils.PtrString(user.Password))
-		p.Outputf("Roles: %v\n", utils.PtrString(user.Roles))
+		//TODO: check if this is ok
```
rubenhoenle: resolve open TODO

internal/cmd/mongodbflex/user/describe/describe.go
```diff
@@ -116,7 +116,8 @@ func outputResult(p *print.Printer, outputFormat string, user mongodbflex.Instan
 		table.AddSeparator()
 		table.AddRow("USERNAME", utils.PtrString(user.Username))
 		table.AddSeparator()
-		table.AddRow("ROLES", utils.PtrString(user.Roles))
+		//TODO: check if this is ok
```
rubenhoenle: same here

internal/cmd/mongodbflex/backup/describe/describe.go
SerseusWasTaken: `mongodbflex.Backup` became a larger struct with the SDK version bump, so the motivation was to make the linter happy (to avoid passing large values by value).
My idea was to use the opportunity to make sure this does not run into a nil pointer dereference.

And yes sure, this could also be moved into the callback

internal/cmd/mongodbflex/instance/create/create_test.go
SerseusWasTaken: Yeah fair point 😄  i will move it into the test struct

internal/cmd/mongodbflex/instance/create/create_test.go
SerseusWasTaken: It is, good catch

internal/cmd/mongodbflex/backup/restore-jobs/restore_jobs.go
SerseusWasTaken: will do

internal/cmd/mongodbflex/backup/describe/describe.go
rubenhoenle: Ok, got it. Thanks for the explanation.

Then I would like to stick to our regular CLI behavior: 
- For JSON/YAML just output what the API provided. 
- For the pretty output throw errors if the API response is nil do prevent nil pointer dereferences. See below.

```golang
func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup *mongodbflex.Backup) error {
    return return p.OutputResult(outputFormat, backup, func() error {
         if backup == nil {
              return fmt.Errorf("API response is nil")
         }

         // build the table and show it ...

        return nil
    }
}
```

internal/cmd/mongodbflex/instance/create/create.go
```diff
@@ -250,19 +250,19 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient MongoDBFlexC
 	}
 
 	req = req.CreateInstancePayload(mongodbflex.CreateInstancePayload{
-		Name:           model.InstanceName,
-		Acl:            &mongodbflex.CreateInstancePayloadAcl{Items: model.ACL},
-		BackupSchedule: model.BackupSchedule,
-		FlavorId:       flavorId,
-		Replicas:       &replicas,
-		Storage: &mongodbflex.Storage{
+		Name:           *model.InstanceName,
```
rubenhoenle: Potential nil pointer dereference.

According to the [API docs](https://docs.api.stackit.cloud/documentation/mongodb-flex-service/version/v2#tag/instance/paths/~1v2~1projects~1%7BprojectId%7D~1regions~1%7Bregion%7D~1instances/post) the instance name is required. So there's need to have the instance name as a string pointer type in the `inputModel` struct. 

So I suggest to do the following change:

```diff
type inputModel struct {
    // ...
-   InstanceName *string
+   InstanceName string
}
```

Same applies for the BackupSchedule and the Version field.

internal/cmd/mongodbflex/instance/create/create.go
```diff
@@ -250,19 +250,19 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient MongoDBFlexC
...
+		Acl:            mongodbflex.ACL{Items: *model.ACL},
```
rubenhoenle: Same here for the ACL field, it's required on API side. No need to have it as a pointer type in the inputModel then.

internal/cmd/mongodbflex/instance/update/update.go
```diff
@@ -233,7 +233,7 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient MongoDBFlexC
 			return req, err
 		}
 	} else if model.FlavorId != nil {
-		err := mongodbflexUtils.ValidateFlavorId(*model.FlavorId, flavors.Flavors)
+		err := mongodbflexUtils.ValidateFlavorId(*model.FlavorId, &flavors.Flavors)
```
rubenhoenle: For legacy reasons `mongodbflexutils.ValidateFlavorId` takes a slice pointer parameter (because the old SDK layer used always pointers on slices).

Since a slice in Golang is already a pointer and can be nil, types like `*[]string` are basically a unneeded double-pointer.

Long story short: Since we got rid of the old SDK layer now, I suggest to do the following change in the implementation:

```diff
- func ValidateFlavorId(flavorId string, flavors *[]mongodbflex.InstanceFlavor) error {
+ func ValidateFlavorId(flavorId string, flavors *[]mongodbflex.InstanceFlavor) error {
```

This way you don't have to pass the pointer to the flavors slice here.

Kind of a nitpick, but let's make this code nice and readable again 😄

internal/cmd/mongodbflex/instance/update/update_test.go
```diff
@@ -590,6 +588,7 @@ func TestBuildRequest(t *testing.T) {
 			diff := cmp.Diff(request, tt.expectedRequest,
 				cmp.AllowUnexported(tt.expectedRequest),
 				cmpopts.EquateComparable(testCtx),
+				cmpopts.IgnoreFields(tt.expectedRequest, "ApiService"),
```
rubenhoenle, line 591: same as here: https://github.com/stackitcloud/stackit-cli/pull/1467#discussion_r3552170597

internal/cmd/mongodbflex/options/options.go
rubenhoenle:
```suggestion
	Flavors  []mongodbflex.InstanceFlavor `json:"flavors,omitempty"`
```

Same here, you can get rid of these double-pointers 😄 These slices should already nil by default when you initialize the struct.

rubenhoenle (dismissed):

internal/cmd/mongodbflex/backup/describe/describe.go
SerseusWasTaken: Alright, yeah thats better :) i will do that

internal/cmd/mongodbflex/instance/create/create.go
SerseusWasTaken: `InstanceName` is marked as a required flag, and `BackupSchedule` has a default value, so i don't think there is any actual risk of nil pointer dereference. And a default value for `version` is fetched if not specified by the user. So this implementation looks safe to me. 
However, i do think it is worth reducing the unneeded pointers in the models, so i would be willing to do that for this PR

internal/cmd/mongodbflex/instance/update/update_test.go
SerseusWasTaken: I don't see an issue here, as `cmp.AllowUnexported` call does not include the `mongodbflex.DefaultAPIService{}`

rubenhoenle: fair point, sorry my bad :smile:

internal/cmd/mongodbflex/instance/create/create.go
```diff
@@ -251,18 +251,18 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient MongoDBFlexC
...
-		Storage: &mongodbflex.Storage{
+		FlavorId:       *flavorId,
+		Replicas:       replicas,
+		Storage: mongodbflex.Storage{
 			Class: model.StorageClass,
 			Size:  model.StorageSize,
 		},
-		Version: model.Version,
-		Options: utils.Ptr(map[string]string{
+		Version: *model.Version,
```
rubenhoenle: I think you missed it for Version. It's also required on creation and shouldn't be a pointer type then 😅

internal/cmd/mongodbflex/backup/describe/describe.go
```diff
@@ -114,11 +114,14 @@ func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inpu
...
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *mongodbflex.APIClient) mongodbflex.ApiGetBackupRequest {
-	req := apiClient.GetBackup(ctx, model.ProjectId, model.InstanceId, model.BackupId, model.Region)
+	req := apiClient.DefaultAPI.GetBackup(ctx, model.ProjectId, model.InstanceId, model.BackupId, model.Region)
 	return req
 }
 
-func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup mongodbflex.Backup) error {
+func outputResult(p *print.Printer, outputFormat, restoreStatus string, backup *mongodbflex.Backup) error {
+	if backup == nil {
```
rubenhoenle: As said, this nil check can be moved to the pretty output.

The JSON/YAML output can handle the nil value just fine so I don't see a reason to throw an error there.

internal/cmd/mongodbflex/instance/create/create.go
SerseusWasTaken: I didn't change it since it is an optional argument that the CLI has to check if it has been set. And i personally find a nil check cleaner than a zero value check here: 
```go
// Fill in version, if needed
if model.Version == nil {
  version, err := mongodbflexUtils.GetLatestMongoDBVersion(ctx, apiClient.DefaultAPI, model.ProjectId, model.Region)
  if err != nil {
    return fmt.Errorf("get latest MongoDB version: %w", err)
  }
  model.Version = &version
}
```
Is it a shared understanding in the team that all values that are required for the API call should be values?

internal/cmd/mongodbflex/backup/describe/describe.go
SerseusWasTaken: My bad, i somehow missed that.

internal/cmd/mongodbflex/instance/create/create.go
rubenhoenle: Ah got it, the CLI fetches the latest version and sets that one in case the flag isn't set. Makes sense. My bad, sorry.

rubenhoenle (approved):

PR #1449, opened 2026-06-29
Feat/region adjustment rabbitmq

SerseusWasTaken:
## Description
Migrates to rabbitmq api v2 and implements appropriate region adjustments

relates to STACKITCLI-407 and STACKITCLI-375

## Checklist

- [x] Issue was linked above
- [x] Code format was applied: `make fmt`
- [ ] Examples were added / adjusted (see e.g. [here](https://github.com/stackitcloud/stackit-cli/blob/ef291d1683ca5b0d719ec0a26ecb999a32685117/internal/cmd/ske/cluster/create/create.go#L49-L63))
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI)
- [x] Unit tests got implemented or updated
- [x] Unit tests are passing: `make test` (will be checked by CI)
- [x] No linter issues: `make lint` (will be checked by CI)

PR #1474, opened 2026-07-13
feat(opensearch): region adjustments and migration to new SDK structure

SerseusWasTaken:
## Description
Migrate opensearch to v2 SDK API

relates to STACKITCLI-372

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/opensearch/credentials/create/create.go
+++ b/internal/cmd/opensearch/credentials/create/create.go
@@ -76,7 +76,7 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				return fmt.Errorf("create OpenSearch credentials: %w", err)
 			}
 
-			return outputResult(params.Printer, model.OutputFormat, model.ShowPassword, instanceLabel, resp)
+			return outputResult(params.Printer, *model, instanceLabel, resp)
 		},
 	}
 	configureFlags(cmd)
@@ -112,23 +112,26 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *opensearch.
 	return req
 }
 
-func outputResult(p *print.Printer, outputFormat string, showPassword bool, instanceLabel string, resp *opensearch.CredentialsResponse) error {
+func outputResult(p *print.Printer, model inputModel, instanceLabel string, resp *opensearch.CredentialsResponse) error {
+	if model.GlobalFlagModel == nil {
+		return fmt.Errorf("no global flags defined")
+	}
 	if resp == nil || resp.Raw == nil {
 		return fmt.Errorf("response or response content is nil")
 	}
 
-	if !showPassword {
+	if !model.ShowPassword {
 		resp.Raw.Credentials.Password = "hidden"
 	}
 
-	return p.OutputResult(outputFormat, resp, func() error {
+	return p.OutputResult(model.OutputFormat, resp, func() error {
 		p.Outputf("Created credentials for instance %q. Credentials ID: %s\n\n", instanceLabel, resp.Id)
 		// The username field cannot be set by the user so we only display it if it's not returned empty
 		if resp.HasRaw() {
 			if username := resp.Raw.Credentials.Username; username != "" {
 				p.Outputf("Username: %s\n", username)
 			}
-			if !showPassword {
+			if !model.ShowPassword {
 				p.Outputf("Password: <hidden>\n")
 			} else {
 				p.Outputf("Password: %s\n", resp.Raw.Credentials.Password)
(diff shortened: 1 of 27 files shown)


internal/pkg/services/opensearch/utils/utils.go
```diff
@@ -29,59 +29,56 @@ func ValidatePlanId(planId string, offerings *opensearch.ListOfferingsResponse)
...
 		return "", fmt.Errorf("get OpenSearch instance: %w", err)
 	}
-	return *resp.Name, nil
+	return resp.Name, nil
 }
 
-func GetCredentialsUsername(ctx context.Context, apiClient OpenSearchClient, projectId, instanceId, credentialsId string) (string, error) {
-	resp, err := apiClient.GetCredentialsExecute(ctx, projectId, instanceId, credentialsId)
+func GetCredentialsUsername(ctx context.Context, apiClient OpenSearchClient, projectId, region, instanceId, credentialsId string) (string, error) {
+	resp, err := apiClient.GetCredentials(ctx, projectId, instanceId, region, credentialsId).Execute()
```
GokceGK: region must be added before instanceId

internal/pkg/services/opensearch/utils/utils.go
```diff
@@ -29,59 +29,56 @@ func ValidatePlanId(planId string, offerings *opensearch.ListOfferingsResponse)
...
 	}
 	details := fmt.Sprintf("You provided plan_name %q for version %s, which is invalid. Available plan names for that version are: %s", planName, version, availablePlanNames)
-	return nil, &errors.DSAInvalidPlanError{
+	return "", &errors.DSAInvalidPlanError{
 		Service: service,
 		Details: details,
 	}
 }
 
 type OpenSearchClient interface {
```
GokceGK: I think we can remove this client and use the apiClient directly. Like we have done for other services 

https://github.com/stackitcloud/stackit-cli/pull/1449/changes#diff-fbfa081e40e62c1d42769762ff3cbb0db379d43baed6ec03fcbe53a4b7f628b7L32

internal/pkg/services/opensearch/utils/utils_test.go
```diff
@@ -5,14 +5,15 @@ import (
...
 	"github.com/google/uuid"
-	"github.com/stackitcloud/stackit-sdk-go/services/opensearch"
+	opensearch "github.com/stackitcloud/stackit-sdk-go/services/opensearch/v2api"
+
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
 )
 
 var (
 	testProjectId     = uuid.NewString()
+	testRegion        = "eu01"
```
GokceGK: can you shift this to const area?

internal/cmd/opensearch/credentials/create/create.go
```diff
@@ -109,35 +108,35 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel,
 }
 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *opensearch.APIClient) opensearch.ApiCreateCredentialsRequest {
-	req := apiClient.CreateCredentials(ctx, model.ProjectId, model.InstanceId)
+	req := apiClient.DefaultAPI.CreateCredentials(ctx, model.ProjectId, model.Region, model.InstanceId)
 	return req
 }
 
 func outputResult(p *print.Printer, outputFormat string, showPassword bool, instanceLabel string, resp *opensearch.CredentialsResponse) error {
```
GokceGK: nitpick: not introduced by this PR so you can just skip it :) but I think instead of outputFormat and showPassword, we can parse here the model and then read the values in the function.

SerseusWasTaken: Out of curiosity: what would be the upside to that? I think it's better to keep the arguments precise. Since the function only needs these two args from the model, i think passing the whole model here would make it less clear on what the function depends on from the model.
If the function changes in the future and more fields are needed from the model then i would see the argument. But i think two args is reasonable

internal/pkg/services/opensearch/utils/utils.go
SerseusWasTaken: Good catch!

internal/pkg/services/opensearch/utils/utils.go
SerseusWasTaken: will do

internal/cmd/opensearch/credentials/create/create.go
GokceGK: I am also fine with 2 args :) It is just to align with similar service implementations. We added some checks like this https://github.com/stackitcloud/stackit-cli/blob/f4e58587b1abafd34f57e93d97430b32e00cbaf3/internal/cmd/redis/credentials/create/create.go#L117

SerseusWasTaken: Ah i see. In that case i'm going to adjust it

GokceGK (approved):

PR #1476, opened 2026-07-15
feat(postgresflex): migrate to V2 api

SerseusWasTaken:
## Description
Migrates postgresflex to the v2 API

relates to STACKITCLI-374

## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/postgresflex/backup/describe/describe_test.go
+++ b/internal/cmd/postgresflex/backup/describe/describe_test.go
@@ -8,7 +8,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/uuid"
-	"github.com/stackitcloud/stackit-sdk-go/services/postgresflex"
+	postgresflex "github.com/stackitcloud/stackit-sdk-go/services/postgresflex/v2api"
 
 	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
 	"github.com/stackitcloud/stackit-cli/internal/pkg/testparams"
@@ -18,7 +18,7 @@ import (
 type testCtxKey struct{}
 
 var testCtx = context.WithValue(context.Background(), testCtxKey{}, "foo")
-var testClient = &postgresflex.APIClient{}
+var testClient = &postgresflex.APIClient{DefaultAPI: &postgresflex.DefaultAPIService{}}
 var testProjectId = uuid.NewString()
 var testInstanceId = uuid.NewString()
 var testBackupId = "backupID"
@@ -63,7 +63,7 @@ func fixtureInputModel(mods ...func(model *inputModel)) *inputModel {
 }
 
 func fixtureRequest(mods ...func(request *postgresflex.ApiGetBackupRequest)) postgresflex.ApiGetBackupRequest {
-	request := testClient.GetBackup(testCtx, testProjectId, testRegion, testInstanceId, testBackupId)
+	request := testClient.DefaultAPI.GetBackup(testCtx, testProjectId, testRegion, testInstanceId, testBackupId)
 	for _, mod := range mods {
 		mod(&request)
 	}
@@ -231,7 +231,7 @@ func TestBuildRequest(t *testing.T) {
 
 			diff := cmp.Diff(request, tt.expectedRequest,
 				cmp.AllowUnexported(tt.expectedRequest),
-				cmpopts.EquateComparable(testCtx),
+				cmpopts.EquateComparable(testCtx, postgresflex.DefaultAPIService{}),
 			)
 			if diff != "" {
 				t.Fatalf("Data does not match: %s", diff)
@@ -243,7 +243,7 @@ func TestBuildRequest(t *testing.T) {
 func Test_outputResult(t *testing.T) {
 	type args struct {
 		outputFormat string
-		backup       postgresflex.Backup
+		backup       *postgresflex.Backup
 	}
 	tests := []struct {
 		name    string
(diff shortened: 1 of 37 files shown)


internal/cmd/postgresflex/backup/list/list.go
```diff
@@ -80,11 +80,11 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("get backups for PostgreSQL Flex instance %q: %w", instanceLabel, err)
 			}
-			if resp.Items == nil || len(*resp.Items) == 0 {
+			if len(resp.Items) == 0 {
 				params.Printer.Outputf("No backups found for instance %q", instanceLabel)
```
rubenhoenle: invalid JSON/YAML output

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -196,11 +190,10 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient PostgreSQLFl
 }
 
 func outputResult(p *print.Printer, outputFormat string, async bool, instanceLabel, instanceId string, resp *postgresflex.CloneInstanceResponse) error {
-	if resp == nil {
-		return fmt.Errorf("response not set")
-	}
-
 	return p.OutputResult(outputFormat, resp, func() error {
+		if resp == nil {
```
rubenhoenle, line 194: Nice, thanks 😅

internal/cmd/postgresflex/instance/list/list.go
```diff
@@ -77,7 +77,7 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 				params.Printer.Info("No instances found for project %q\n", projectLabel)
```
rubenhoenle: Invalid JSON/YAML output

internal/cmd/postgresflex/options/options.go
```diff
@@ -233,12 +227,11 @@ func buildVersionsTable(versions []string) tables.Table {
 }
 
 func buildStoragesTable(storagesResp postgresflex.ListStoragesResponse) tables.Table {
-	storages := *storagesResp.StorageClasses
 	table := tables.NewTable()
 	table.SetTitle("Storages")
 	table.SetHeader("MINIMUM", "MAXIMUM", "STORAGE CLASS")
-	for i := range storages {
-		sc := storages[i]
+	for i := range storagesResp.StorageClasses {
```
rubenhoenle:
```suggestion
	for _, sc := range storagesResp.StorageClasses {
```

internal/cmd/postgresflex/user/list/list.go
```diff
@@ -69,16 +69,16 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-			if resp.Items == nil || len(*resp.Items) == 0 {
-				instanceLabel, err := postgresflexUtils.GetInstanceName(ctx, apiClient, model.ProjectId, model.Region, *model.InstanceId)
+			if len(resp.Items) == 0 {
+				instanceLabel, err := postgresflexUtils.GetInstanceName(ctx, apiClient.DefaultAPI, model.ProjectId, model.Region, model.InstanceId)
 				if err != nil {
 					params.Printer.Debug(print.ErrorLevel, "get instance name: %v", err)
-					instanceLabel = *model.InstanceId
+					instanceLabel = model.InstanceId
 				}
 				params.Printer.Info("No users found for instance %q\n", instanceLabel)
```
rubenhoenle: Invalid JSON/YAML output

internal/cmd/postgresflex/backup/list/list.go
SerseusWasTaken: Oh yeah, that went completely over my head since it was not in the `outputResult` function 😄 

I also noticed here that the json serializer prints nil slices as `null` instead of `[]`. And apparently the SDK returns a nil slice when no backup exists. This does not seem to happen when trying to list users for a database without any. I'm not sure why the SDK seems to behave that way.

I'll push a workaround with a regression test.

internal/cmd/postgresflex/backup/describe/describe_test.go
```diff
@@ -243,19 +243,19 @@ func TestBuildRequest(t *testing.T) {
...
-		backup       postgresflex.Backup
+		backup       *postgresflex.Backup
 	}
 	tests := []struct {
 		name    string
 		args    args
 		wantErr bool
 	}{
 		{"empty", args{}, true},
-		{"standard", args{outputFormat: "", backup: postgresflex.Backup{StartTime: utils.Ptr(time.Now().Format(time.RFC3339))}}, false},
```
rubenhoenle, line 254: Not your fault, this was a mess already before but maybe let's format this table test properly so it's readable again:

```golang
{
			name:    "empty",
			args:    args{},
			wantErr: true,
		},
		{
			name: "standard",
			args: args{
				outputFormat: "",
				backup: &postgresflex.Backup{
					StartTime: utils.Ptr(time.Now().Format(time.RFC3339)),
				},
			},
			wantErr: false,
		},
		{
			name: "complete",
			args: args{
				outputFormat: "",
				backup: &postgresflex.Backup{
					EndTime:   utils.Ptr(time.Now().Format(time.RFC3339)),
					Id:        utils.Ptr("id"),
					Labels:    []string{"foo", "bar", "baz"},
					Name:      utils.Ptr("name"),
					Options:   &map[string]string{"test1": "test1", "test2": "test2"},
					Size:      utils.Ptr(int64(42)),
					StartTime: utils.Ptr(time.Now().Format(time.RFC3339)),
				},
			},
			wantErr: false,
		},
```

internal/cmd/postgresflex/backup/list/list.go
rubenhoenle: > I also noticed here that the json serializer prints nil slices as null instead of []. 

I think this is correct and should be left like it is.

> And apparently the SDK returns a nil slice when no backup exists. This does not seem to happen when trying to list users for a database without any. I'm not sure why the SDK seems to behave that way.

Did you check how the API behaves here? Maybe the SDK is behaving just fine but the API is responding with 

```json
{
  "items": null
}
```

and with

```json
{
  "items": []
}
```

for the other endpoint. That would explain the behavior of our SDK 😄

SerseusWasTaken: That's a good point, that must be it 😄 I'll leave it as it is in that case.

rubenhoenle (approved):

PR #1477, opened 2026-07-15
feat(ske): support stateless WIF kubeconfig login

Galin-Karabadzhakov:
## Description

Enable `stackit ske kubeconfig login --idp` to operate as a stateless Kubernetes exec credential provider using STACKIT Workload Identity Federation.

Kubernetes clients do not always include cluster information in `KUBERNETES_EXEC_INFO`, and in-cluster workloads should not require persisted CLI credentials or cache state.

This change:

- Adds explicit cluster and organization flags as fallbacks when exec cluster information is omitted.
- Uses the existing global project and region configuration.
- Automatically detects the projected SKE workload identity token or `STACKIT_FEDERATED_TOKEN_FILE`.
- Obtains a STACKIT access token through the SDK WIF flow.
- Exchanges the access token for an SKE cluster IdP token without requiring authentication storage.
- Supports stateless authentication through `STACKIT_ACCESS_TOKEN`.
- Preserves existing kubeconfig and cache behavior for non-stateless authentication.
- Adds documentation and unit coverage for the new flows.

Relates to #1327.

This is complementary to #1424, which introduces general WIF authentication support.

## Validation

- `make fmt`
- `make generate-docs`
- `make lint`
- `make build`
- `go test ./internal/pkg/auth ./internal/cmd/ske/kubeconfig/login`
- Live SKE validation using a projected workload identity, read-only root filesystem, and no persisted home directory

The repository-wide `make test` currently reports two unrelated failures that were reproduced on an unchanged `main` checkout.

## Checklist

- [x] Issue was linked above
- [x] Examples were added or adjusted
- [x] Docs are up-to-date: `make generate-docs`
- [x] Unit tests were implemented or updated
- [x] Affected unit tests are passing
- [x] No linter issues: `make lint`


Galin-Karabadzhakov:
Hi team, could you please confirm whether this approach aligns with the intended direction for non-interactive SKE authentication?
I have validated the implementation with workload identity federation and credential renewal in a Kubernetes exec credential flow. It allows stateless, short-lived authentication without storing bearer tokens or service-account keys.
If a different command structure or authentication integration is preferred, I would be happy to adjust the implementation. A brief direction from the maintainers would help us move this forward and avoid duplicating functionality outside the CLI.

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -148,33 +171,115 @@ func parseClusterConfig(p *print.Printer, cmd *cobra.Command, idpMode bool) (*cl
...
+	if organizationID := flags.FlagToStringValue(p, cmd, organizationFlag); organizationID != "" {
+		clusterConfig.OrganizationID = organizationID
+	}
+	globalFlags := globalflags.Parse(p, cmd)
+	if clusterConfig.STACKITProjectID == "" {
+		clusterConfig.STACKITProjectID = globalFlags.ProjectId
+	}
+	if clusterConfig.Region == "" {
+		clusterConfig.Region = globalFlags.Region
+	}
```
SerseusWasTaken, line 210: It looks like `--cluster-name` and `--organization-id` **override** the values configured from the clusterConfig, while `--region` and `--project-id` are only used as **fallbacks** when the clusterConfig has not defined them. I think they should behave the same.

internal/pkg/auth/utils.go
```diff
@@ -60,16 +64,35 @@ func retrieveIDPWellKnownConfig(p *print.Printer) (*wellKnownConfig, error) {
...
 	if err != nil {
 		return nil, fmt.Errorf("parse IDP well-known configuration: %w", err)
 	}
 	return idpWellKnownConfig, nil
 }
 
+// GetIDPTokenEndpoint returns the configured IdP token endpoint without requiring
+// authentication storage. Persisted configuration is reused when available.
+func GetIDPTokenEndpoint(p *print.Printer) (string, error) {
+	tokenEndpoint, err := GetAuthField(IDP_TOKEN_ENDPOINT)
```
SerseusWasTaken, line 77: We should only call `GetAuthField` if we are in stateless mode, since `GetAuthField` can try to write to the system when no there is no auth keyring, since it calls `getAuthFieldFromEncodedTextFile` from `storage.go`.
I would suggest to implement another function like `GetIDPTokenEndpointStateless` that recieves the `tokenEndpoint` information like so:  `wellKnownConfig, err := retrieveIDPWellKnownConfigWithStorage(p, false)` and call this new function when we are in stateless mode, instead of `GetIDPTokenEndpoint`

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Good catch. I made the behavior consistent across all four values: cluster configuration from KUBERNETES_EXEC_INFO takes precedence, while CLI/global values are used only as fallbacks. I also added a regression test with conflicting values.

internal/pkg/auth/utils.go
Galin-Karabadzhakov: Agreed. I added GetIDPTokenEndpointStateless, which performs endpoint discovery without reading from or writing to authentication storage. Stateless WIF and STACKIT_ACCESS_TOKEN flows now use this function, while stateful flows continue using the stored endpoint when available. Regression coverage was added for this behavior.

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -118,6 +143,8 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(idpFlag, false, "Use the STACKIT IdP for authentication to the cluster.")
+	cmd.Flags().String(clusterNameFlag, "", "SKE cluster name. Used when the Kubernetes exec request does not provide cluster information.")
+	cmd.Flags().String(organizationFlag, "", "Organization ID. Used in IdP mode when the Kubernetes exec request does not provide cluster information.")
```
marceljk, line 152: Not sure if there is a common practice for this in Kubernetes tools, because I didn't work much with Kubernetes, but usually we prioritize in our tools the explicit config first and if this is not set, it falls back to the alternatives. (See for example also the authentication in the Terraform provider: https://github.com/stackitcloud/terraform-provider-stackit#authentication)

I would also suggest to have it like that here, but I'm also open for discussion.

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -315,30 +437,53 @@ func getAccessToken(params *types.CmdParams) (string, error) {
...
+	if os.Getenv(clients.FederatedTokenFileEnv) != "" {
+		return true
+	}
+	fileInfo, err := os.Stat(defaultFederatedTokenPath)
+	return err == nil && !fileInfo.IsDir()
+}
 
-	return accessToken, nil
+func getWorkloadIdentityAccessToken() (string, error) {
+	roundTripper, err := sdkAuth.SetupAuth(&sdkConfig.Configuration{WorkloadIdentityFederation: true})
```
marceljk: When in the configuration is no wif function configured, it fallbacks to the file based authentication, which works well for a kubernetes environment, but not for pipelines in Github or ADO.

I would suggest to use the oidcadapter from this function, which works also with Github or Azure, and pass it to the sdkConfig

https://github.com/stackitcloud/stackit-cli/blob/0aa5f4f3561b61bc5f5919b979cc179360bd8cea/internal/pkg/auth/oidc.go#L40-L80

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -315,30 +437,53 @@ func getAccessToken(params *types.CmdParams) (string, error) {
...
+func workloadIdentityConfigured() bool {
+	if os.Getenv(envServiceAccountEmail) == "" {
+		return false
 	}
+	if os.Getenv(clients.FederatedTokenFileEnv) != "" {
+		return true
+	}
+	fileInfo, err := os.Stat(defaultFederatedTokenPath)
+	return err == nil && !fileInfo.IsDir()
+}
```
marceljk:
```suggestion
    if auth.IsOIDCEnabled() {
        return true
    }
    _, err := auth.OIDCTokenFunc()
    return err == nil
}
```
The checks should be adjusted, so that it also takes care of the alternatives

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Agreed. I’ll change cluster configuration resolution so explicit flags/global configuration take precedence, with KUBERNETES_EXEC_INFO used as fallback, and update the conflict tests.

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Agreed. I’ll use auth.OIDCTokenFunc() for both workload-identity detection and the SDK WIF configuration. This will support projected Kubernetes tokens, explicit tokens, GitHub Actions, and Azure DevOps consistently.

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Agreed. I’ll retain the service-account email validation, then use auth.IsOIDCEnabled() and auth.OIDCTokenFunc() as suggested instead of checking only file-based token sources.

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Implemented in 309e0e81. Explicit cluster name, organization ID, project ID, and region now take precedence; KUBERNETES_EXEC_INFO remains the fallback. The tests cover both conflicting explicit values and exec-only fallback.

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Implemented in 0cce847c. The SKE login flow now passes the token function returned by auth.OIDCTokenFunc() into sdkConfig.Configuration, while a nil function preserves the SDK projected-token-file fallback. Coverage now exercises static tokens, configured files, GitHub Actions, and Azure DevOps.

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Implemented in 445e7b2d and completed with coverage in 0cce847c. Detection now honors explicit OIDC and all token sources exposed by auth.OIDCTokenFunc(), while still requiring a service-account email.

Galin-Karabadzhakov (commented):
Thanks for the review. All requested changes have now been addressed:
- Explicit flags/global configuration take precedence over KUBERNETES_EXEC_INFO, which remains the fallback.
- auth.OIDCTokenFunc() is used for workload-identity detection and SDK WIF configuration.
- Static tokens, configured token files, GitHub Actions, Azure DevOps, and Kubernetes projected-token fallback are covered.
- Stateless IdP endpoint discovery avoids authentication storage.
Validation completed successfully with focused unit tests, go vet, and make lint. Please re-review when convenient.

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -148,33 +173,115 @@ func parseClusterConfig(p *print.Printer, cmd *cobra.Command, idpMode bool) (*cl
...
+	if organizationID := flags.FlagToStringValue(p, cmd, organizationFlag); organizationID != "" {
+		clusterConfig.OrganizationID = organizationID
+	}
+	globalFlags := globalflags.Parse(p, cmd)
+	if globalFlags.ProjectId != "" {
+		clusterConfig.STACKITProjectID = globalFlags.ProjectId
+	}
+	if globalFlags.Region != "" {
+		clusterConfig.Region = globalFlags.Region
+	}
```
marceljk, line 210: I recognize that with the change that the flags should always override the clusterConfig, we ran into a new issue.
If this command is used in a setup where the STACKIT CLI config is set, this could lead to potential issues, because `globalflags.Parse()` reads the globalFlags either from the passed flags or reads them from the config. This can potentially make a valid clusterConfig invalid, because the correct projectId / region are overridden with the wrong one.

Our wanted behavior would be the following:

| flag     | global   | clusterConfig | result        |
|-------|--------|---------------|---------------|
| none   |  set     | none                | global        |
| set      |  set     | none               | flag          |
| none   |  set     | set                  | clusterConfig |
| set      |  set     | set                 | flag          |

- flag is the explicit flag `--project-id` or `--region`
- global is the value out of the CLI config file
- clusterConfig is the parsed value out of the env KUBERNETES_EXEC_INFO

It would be also great to document the behavior with the table in the help output of this command and cover this behavior also by some unit tests.

Sorry for the back and forth with these flags. 

I already prepared some changes locally to test how the change could look like. If you like, I can also commit the wanted change to this PR.

Galin-Karabadzhakov: Thanks, the precedence matrix makes the expected behavior clear. I agree that explicit --project-id and --region flags should override KUBERNETES_EXEC_INFO, while values from the global CLI config should only be used when the exec cluster configuration does not provide them.
I’ll update the resolution logic to distinguish explicitly changed Cobra flags from Viper-loaded configuration, document the precedence table in the command help, and add tests covering all four combinations. Since you already prepared the change, you’re also welcome to commit it directly to this PR.

Galin-Karabadzhakov (commented):
Implemented and pushed commit **3de94aad** to **PR #1477**.

Changes:

- Explicit flags override all sources.
- KUBERNETES_EXEC_INFO overrides global CLI configuration.
- Global configuration fills missing exec values.
- Added tests for all four precedence combinations.
- Added the precedence table to CLI help and generated docs.

SKE tests, vet, docs check, and lint pass. Full tests only encountered an unrelated ANSI-color snapshot failure in the CDN package.

internal/cmd/ske/kubeconfig/login/login.go
```diff
@@ -50,7 +59,14 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
 			"First you need to obtain a kubeconfig for use with the login command (first or second example).",
-			"Secondly you use the kubeconfig with your chosen Kubernetes client (third example), the client will automatically retrieve the credentials via the STACKIT CLI.",
+			"Secondly you use the kubeconfig with your chosen Kubernetes client (third example), the client will automatically retrieve the credentials via the STACKIT CLI.\n\n"+
+				"Project ID and region are resolved with the following precedence:\n\n"+
+				"| Explicit flag | Global config | Exec cluster config | Result |\n"+
+				"| --- | --- | --- | --- |\n"+
+				"| Not set | Set | Not set | Global config |\n"+
+				"| Set | Set | Not set | Explicit flag |\n"+
+				"| Not set | Set | Set | Exec cluster config |\n"+
+				"| Set | Set | Set | Explicit flag |",
```
marceljk, line 69: For the generated markdown this table looks good, but in the help output of the command it's a bit off. Fixed version:
```suggestion
				"| Explicit flag | Global config | Exec cluster config | Result              |\n"+
				"| ------------- | ------------- | ------------------- | ------------------- |\n"+
				"| Not set       | Set           | Not set             | Global config       |\n"+
				"| Set           | Set           | Not set             | Explicit flag       |\n"+
				"| Not set       | Set           | Set                 | Exec cluster config |\n"+
				"| Set           | Set           | Set                 | Explicit flag       |",
```

marceljk (commented):
Thanks for your contribution!
Left only one nitpick comment, and besides that, it looks good from my side.

internal/cmd/ske/kubeconfig/login/login.go
Galin-Karabadzhakov: Agreed, thanks. I’ll apply the aligned table formatting and regenerate the command documentation.

Galin-Karabadzhakov (commented):
Thanks for the review. Agreed on the final formatting nitpick. I’ll align the table for terminal output and regenerate the documentation. No functional changes are required.

jastBytes (approved):
Thank you for your contribution! Looks good.

SerseusWasTaken (approved):

PR #1503, opened 2026-07-27
chore(edge): refactoring and migration to new SDK structure

GokceGK:
## Description

relates to STACKITCLI-357

## Checklist

- [x] Issue was linked above


internal/cmd/beta/edge/instance/create/create.go
```diff
@@ -19,34 +18,41 @@ import (
...
+	descriptionFlag = "description"
+	planIdFlag      = "plan-id" // UUID
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+	DisplayName *string
+	Description string
+	PlanId      *string
+}
```
SerseusWasTaken, line 34: `description` is the only optional field in the API, so i think we should flip the pointers:

```suggestion
type inputModel struct {
	*globalflags.GlobalFlagModel
	DisplayName string
	Description *string
	PlanId      string
}
```

internal/cmd/beta/edge/instance/create/create.go
```diff
@@ -60,169 +66,100 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-
-	planIdValue := flags.FlagToStringPointer(p, cmd, commonInstance.PlanIdFlag)
-	if err := commonInstance.ValidatePlanId(planIdValue); err != nil {
-		return nil, err
-	}
-
-	descriptionValue := flags.FlagWithDefaultToStringValue(p, cmd, commonInstance.DescriptionFlag)
-	if err := commonInstance.ValidateDescription(descriptionValue); err != nil {
+	displayNameValue := flags.FlagToStringPointer(p, cmd, displayNameFlag)
+	planIdValue := flags.FlagToStringValue(p, cmd, planIdFlag)
```
SerseusWasTaken: Not sure if this is beyond the scope of this PR/ticket, but we could add validation of `planId` here, similar to what we have with `flavorId` for postgres for example

internal/cmd/beta/edge/instance/create/create.go
```diff
@@ -60,169 +66,100 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
 
-// Output result based on the configured output format
-func outputResult(p *print.Printer, outputFormat string, async bool, projectLabel string, instance *edge.Instance) error {
+func outputResult(p *print.Printer, model *inputModel, projectLabel string, instance *edge.Instance) error {
 	if instance == nil {
-		// This is only to prevent nil pointer deref.
-		// As long as the API behaves as defined by it's spec, instance can not be empty (HTTP 200 with an empty body)
-		return commonErr.NewNoInstanceError("")
+		return fmt.Errorf("instance response is empty")
 	}
```
SerseusWasTaken: Let's move this into the `OutputResult` lambda function below to fix this output for JSON/YAML

internal/cmd/beta/edge/instance/describe/describe.go
```diff
@@ -76,123 +59,62 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-}
-
-// Output result based on the configured output format
 func outputResult(p *print.Printer, outputFormat string, instance *edge.Instance) error {
 	if instance == nil {
-		// This is only to prevent nil pointer deref.
-		// As long as the API behaves as defined by it's spec, instance can not be empty (HTTP 200 with an empty body)
-		return commonErr.NewNoInstanceError("")
+		return fmt.Errorf("instance response is empty")
 	}
```
SerseusWasTaken: same here, move this into the function below

internal/cmd/beta/edge/instance/update/update.go
```diff
@@ -98,187 +73,98 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-					}
-					waiter := waiterFactory(concreteClient)
-					_, err = waiter.WaitWithContext(ctx)
+					_, err = wait.CreateOrUpdateInstanceWaitHandler(ctx, apiClient.DefaultAPI, model.ProjectId, model.Region, instanceId).WaitWithContext(ctx)
 					return err
 				})
-
 				if err != nil {
-					return fmt.Errorf("wait for edge instance update: %w", err)
+					return fmt.Errorf("wait for SQLServer Flex instance update: %w", err)
```
SerseusWasTaken:
```suggestion
					return fmt.Errorf("wait for edge cloud instance update: %w", err)
```

internal/cmd/beta/edge/kubeconfig/create/create.go
```diff
@@ -131,20 +91,58 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
+			if err != nil {
+				return fmt.Errorf("build kubeconfig create request: %w", err)
+			}
+			respKubeconfig, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("create kubeconfig for Edge Cloud instance: %w", err)
+			}
+			if respKubeconfig.Kubeconfig == nil {
+				return fmt.Errorf("no kubeconfig returned from the API")
+			}
```
SerseusWasTaken: I would check for the whole reply and not the map here, and let the `outputResult` function handle the nil map case

```suggestion
			if respKubeconfig == nil {
				return fmt.Errorf("no kubeconfig returned from the API")
			}
```

internal/cmd/beta/edge/instance/create/create.go
```diff
@@ -60,169 +66,100 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
+	req = apiClient.DefaultAPI.CreateInstance(ctx, model.ProjectId, model.Region)
 
-// buildRequest constructs the spec that can be tested.
-func buildRequest(ctx context.Context, model *inputModel, apiClient client.APIClient) (*createRequestSpec, error) {
-	req := apiClient.CreateInstance(ctx, model.ProjectId, model.Region)
-
-	// Build request payload
 	payload := edge.CreateInstancePayload{
-		DisplayName: &model.DisplayName,
+		DisplayName: *model.DisplayName,
```
rubenhoenle: Both the display name as the plan id are required flags. Why are they pointers in your `inputModel` struct then? I see only a risk for nil pointers here without any benefit 😅

internal/cmd/beta/edge/instance/create/create.go
```diff
@@ -60,169 +66,100 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-		Execute:   req.Execute,
-	}, nil
+
+	return req.CreateInstancePayload(payload), nil
 }
 
-// Output result based on the configured output format
-func outputResult(p *print.Printer, outputFormat string, async bool, projectLabel string, instance *edge.Instance) error {
+func outputResult(p *print.Printer, model *inputModel, projectLabel string, instance *edge.Instance) error {
 	if instance == nil {
```
rubenhoenle: Move this nil check into the callback of `p.OutputResult`. The JSON/YAML output can handle nil values just fine.

internal/cmd/beta/edge/instance/create/create_test.go
```diff
@@ -34,86 +27,17 @@ var (
 	testPlanId      = uuid.NewString()
 	testDescription = "Initial instance description"
```
rubenhoenle, line 32: Could be const

internal/cmd/beta/edge/instance/delete/delete.go
```diff
@@ -89,155 +59,67 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-				operationState = "Deleted"
 			}
 
-			params.Printer.Info("%s instance with %q %q of project %q.\n", operationState, model.identifier.Flag, model.identifier.Value, projectLabel)
-
+			operationState := "Deleted"
+			if model.Async {
+				operationState = "Triggered deletion of"
+			}
+			params.Printer.Info("%s instance %q of project %q.\n", operationState, instanceLabel, projectLabel)
```
rubenhoenle:
```suggestion
			params.Printer.Outputf("%s instance %q of project %q.\n", operationState, instanceLabel, projectLabel)
```

I hope this is the last time I will have to comment this. Stderr vs. stdout. 😉

internal/cmd/beta/edge/instance/describe/describe.go
```diff
@@ -76,123 +59,62 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-	return spec, nil
-}
-
-// Output result based on the configured output format
 func outputResult(p *print.Printer, outputFormat string, instance *edge.Instance) error {
 	if instance == nil {
-		// This is only to prevent nil pointer deref.
-		// As long as the API behaves as defined by it's spec, instance can not be empty (HTTP 200 with an empty body)
-		return commonErr.NewNoInstanceError("")
+		return fmt.Errorf("instance response is empty")
```
rubenhoenle: same here, move this nil check into the callback func please

internal/cmd/beta/edge/instance/describe/describe_test.go
```diff
@@ -32,544 +21,185 @@ var (
 	testProjectId = uuid.NewString()
 	testRegion    = "eu01"
```
rubenhoenle: could be const

internal/cmd/beta/edge/instance/describe/describe_test.go
```diff
@@ -32,544 +21,185 @@ var (
...
-			wantErr: &cliErr.RequestFailedError{},
-			args: args{
-				model: fixtureByNameInputModel(),
-				client: &mockAPIClient{
-					getInstanceByNameMock: &mockExecutable{
-						executeFails: true,
-					},
-				},
-			},
+			name:    "empty",
```
rubenhoenle:
```suggestion
			name:    "instance is nil",
```

The name is not really in sync what's done here.

Maybe also worth to write the test case explictly like this so it's better understandable from a first glance:

```golang
{
    name: "instance is nil",
    args: args{
        instance: nil,
    }
    wantErr: true,
}
```

internal/cmd/beta/edge/instance/list/list.go
```diff
@@ -120,71 +109,31 @@ func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
...
 		if len(instances) == 0 {
 			p.Outputf("No instances found for project %q\n", projectLabel)
 			return nil
 		}
 
-		// Display instances found for project in a table
 		table := tables.NewTable()
-		// List: only output the most important fields. Be sure to filter for any non-required fields.
 		table.SetHeader("ID", "NAME", "UI", "STATE")
 		for i := range instances {
```
rubenhoenle:
```suggestion
		for _, instance := range instances {
```

If we're already refactoring let's please to it properly 😄

internal/cmd/beta/edge/instance/update/update.go
```diff
@@ -98,187 +73,98 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-				operationState = "Updated"
 			}
 
-			params.Printer.Info("%s instance with %q %q of project %q.\n", operationState, model.identifier.Flag, model.identifier.Value, projectLabel)
-
+			operationState := "Updated"
+			if model.Async {
+				operationState = "Triggered update of"
+			}
+			params.Printer.Info("%s instance %q\n", operationState, instanceLabel)
```
rubenhoenle: Same here: stderr vs stdout
```suggestion
			params.Printer.Outputf("%s instance %q\n", operationState, instanceLabel)
```

internal/cmd/beta/edge/instance/update/update.go
```diff
@@ -98,187 +73,98 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-			err = run(ctx, model, apiClient)
+			req, err := buildRequest(ctx, model, apiClient)
 			if err != nil {
 				return err
 			}
+			err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("update Edge Cloud instance: %w", err)
+			}
+			instanceId := model.InstanceId
```
rubenhoenle: From a first glance this variable is used exactly once. What's the point in having it?

internal/cmd/beta/edge/instance/update/update.go
```diff
@@ -98,187 +73,98 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-	identifierFlags := []string{commonInstance.InstanceIdFlag, commonInstance.DisplayNameFlag}
-	cmd.MarkFlagsMutuallyExclusive(identifierFlags...) // InstanceId xor DisplayName
-	cmd.MarkFlagsOneRequired(identifierFlags...)
+	cmd.Flags().StringP(descriptionFlag, "d", "", "A user chosen description to distinguish multiple instances.")
+	cmd.Flags().String(planIdFlag, "", "Service plan configures the size of the Instance.")
 
 	// Make sure at least one updatable field is provided, otherwise it would be a no-op
-	updatedFields := []string{commonInstance.DescriptionFlag, commonInstance.PlanIdFlag}
+	updatedFields := []string{descriptionFlag, planIdFlag}
 	cmd.MarkFlagsOneRequired(updatedFields...)
```
rubenhoenle:
```suggestion
	cmd.MarkFlagsOneRequired(descriptionFlag, planIdFlag)
```

What's the point in having the `updatedFields` slice here?

internal/cmd/beta/edge/instance/update/update.go
```diff
@@ -98,187 +73,98 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
+func parseInput(p *print.Printer, cmd *cobra.Command, inputArgs []string) (*inputModel, error) {
+	instanceId := inputArgs[0]
+
 	globalFlags := globalflags.Parse(p, cmd)
 	if globalFlags.ProjectId == "" {
 		return nil, &cliErr.ProjectIdError{}
 	}
 
-	// Generate input model based on chosen flags
 	model := inputModel{
```
rubenhoenle, line 138:
```golang
planIdValue := flags.FlagToStringValue(p, cmd, planIdFlag)
err := utils.ValidateUUID(planIdValue)
if err != nil {
	return nil, err
}

model := inputModel{
	GlobalFlagModel: globalFlags,
    PlanId: &planIdValue,
    InstanceId: instanceId,
    Description: flags.FlagToStringPointer(p, cmd, descriptionFlag),
}
```

If you change the order here a little bit you get some actually readable code.

rubenhoenle: Even way shorter, but it's more about readability for me here.

internal/cmd/beta/edge/kubeconfig/create/create.go
```diff
@@ -21,106 +23,64 @@ import (
...
-// is guaranteed to be unique within a given project as of today. The chosen flag over args approach ensures we
-// won't need a breaking change of the CLI when we refactor the commands to take the identifier as arg at some point.
+// NewCmd https://aip.stackit.cloud/aip/general/0121/
+// We have decided to eliminate the usage of display name flag
+// To be the AIP compliant, and align with the standard CLI implementation, we will use the InstanceID arg
 func NewCmd(params *types.CmdParams) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create",
-		Short: "Creates or updates a local kubeconfig file of an edge instance",
+		Use:   fmt.Sprintf("create kubeconfig for %s", instanceIdArg),
```
rubenhoenle:
```suggestion
		Use:   "create",
```

This is no description field

internal/cmd/beta/edge/kubeconfig/create/create.go
```diff
@@ -392,6 +309,6 @@ func marshalKubeconfig(kubeconfigMap map[string]interface{}, format string) (str
 		}
 		return string(kubeconfigYAML), nil
 	default:
-		return "", fmt.Errorf("%w: %s", commonErr.NewNoIdentifierError(""), format)
+		return "", fmt.Errorf("format is not JSON or YAML: %s", format)
```
rubenhoenle, line 312: can't we use `p.OutputResult` here?

internal/cmd/beta/edge/token/create/create.go
```diff
@@ -19,94 +20,46 @@ import (
...
-// is guaranteed to be unique within a given project as of today. The chosen flag over args approach ensures we
-// won't need a breaking change of the CLI when we refactor the commands to take the identifier as arg at some point.
+// NewCmd https://aip.stackit.cloud/aip/general/0121/
+// We have decided to eliminate the usage of display name flag
+// To be the AIP compliant, and align with the standard CLI implementation, we will use the InstanceID arg
 func NewCmd(params *types.CmdParams) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create",
-		Short: "Creates a token for an edge instance",
+		Use:   fmt.Sprintf("create token for %s", instanceIdArg),
```
rubenhoenle:
```suggestion
		Use:   fmt.Sprintf("create"),
```

Same here. Furthermore we don't use arguments on create commands. The instance id would be a flag.

internal/cmd/beta/edge/token/create/create.go
```diff
@@ -19,94 +20,46 @@ import (
...
+		Short: "Creates a token for an Edge Cloud instance",
 		Long: fmt.Sprintf("%s\n\n%s\n%s",
 			"Creates a token for a STACKIT Edge Cloud (STEC) instance.",
-			fmt.Sprintf("An expiration time can be set for the token. The expiration time is set in seconds(s), minutes(m), hours(h), days(d) or months(M). Default is %d seconds.", commonKubeconfig.ExpirationSecondsDefault),
+			"An expiration time can be set for the token. The expiration time is set in seconds(s), minutes(m), hours(h), days(d) or months(M). Default is 3600(1h) seconds.",
 			"Note: the format for the duration is <value><unit>, e.g. 30d for 30 days. You may not combine units."),
-		Args: args.NoArgs,
+		Args: args.SingleArg(instanceIdArg, nil),
```
rubenhoenle: 1. We don't do that by convention. Instance ID would have to be a flag.
2. This is a breaking change.

internal/cmd/beta/edge/token/create/create.go
```diff
@@ -200,90 +175,20 @@ func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
...
-	}
+func buildRequest(ctx context.Context, model *inputModel, apiClient *edge.APIClient) (edge.ApiGetTokenByInstanceIdRequest, error) {
+	req := apiClient.DefaultAPI.GetTokenByInstanceId(ctx, model.ProjectId, model.Region, model.InstanceId)
+	return req.ExpirationSeconds(int64(model.Expiration)), nil // #nosec G115 ValidateExpiration ensures safe bounds, conversion is safe
 }
 
 // Output result based on the configured output format
 func outputResult(p *print.Printer, outputFormat string, token *edge.Token) error {
-	if token == nil || token.Token == nil {
+	if token == nil {
```
rubenhoenle: same as above, move this nil check into the callback for the pretty output

internal/cmd/beta/edge/token/create/create.go
```diff
@@ -200,90 +175,20 @@ func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
...
 		// This is only to prevent nil pointer deref.
 		// As long as the API behaves as defined by it's spec, instance can not be empty (HTTP 200 with an empty body)
 		return fmt.Errorf("no token returned from the API")
 	}
-	tokenString := *token.Token
+	tokenString := token.Token
 
 	return p.OutputResult(outputFormat, token, func() error {
 		p.Outputln(tokenString)
```
rubenhoenle:
```suggestion
		p.Outputln(token.Token)
```

Why a variable for everything?

rubenhoenle (changes requested):
Just a first glance, this thing is big. Will have to take a second look but I have to quit work for today.

internal/pkg/services/edge/utils/utils.go
```diff
@@ -0,0 +1,16 @@
+package utils
+
+import (
+	"context"
+	"fmt"
+
+	edge "github.com/stackitcloud/stackit-sdk-go/services/edge/v1beta1api"
+)
+
+func GetInstanceName(ctx context.Context, apiClient edge.DefaultAPI, projectId, instanceId, region string) (string, error) {
```
SerseusWasTaken: Nitpick: swap region and instanceId argument to be in the same order as the APIs expect:

```suggestion
func GetInstanceName(ctx context.Context, apiClient edge.DefaultAPI, projectId, region, instanceId string) (string, error) {
```

internal/cmd/beta/edge/instance/create/create.go
GokceGK: I have added a UUID validation already. I thought this is enough validation from the CLI perspective. But as postgres and also some DSA services have validations for flavors and plans, I can also add here one.

GokceGK: We have decided that we will leave validation to the API side as much as possible.

internal/cmd/beta/edge/kubeconfig/create/create.go
GokceGK: not really, here I am returning the marshalled data which can be written in a file or output depending on the given flags.

internal/cmd/beta/edge/instance/create/create.go
```diff
@@ -60,169 +64,96 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-	}
-
-	resp, err := spec.Execute()
-	if err != nil {
-		return nil, cliErr.NewRequestFailedError(err)
-	}
-
-	return resp, nil
-}
+func buildRequest(ctx context.Context, model *inputModel, apiClient *edge.APIClient) (req edge.ApiCreateInstanceRequest, err error) {
```
SerseusWasTaken: Nitpick: any reason why we use named return values here? they aren't actually by the function and pretty unusual in our codebase

internal/cmd/beta/edge/token/create/create.go
```diff
@@ -19,94 +20,46 @@ import (
...
+// To be the AIP compliant, and align with the standard CLI implementation, we will use the InstanceID arg
 func NewCmd(params *types.CmdParams) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create",
-		Short: "Creates a token for an edge instance",
+		Short: "Creates a token for an Edge Cloud instance",
 		Long: fmt.Sprintf("%s\n\n%s\n%s",
 			"Creates a token for a STACKIT Edge Cloud (STEC) instance.",
-			fmt.Sprintf("An expiration time can be set for the token. The expiration time is set in seconds(s), minutes(m), hours(h), days(d) or months(M). Default is %d seconds.", commonKubeconfig.ExpirationSecondsDefault),
+			"An expiration time can be set for the token. The expiration time is set in seconds(s), minutes(m), hours(h), days(d) or months(M). Default is 3600(1h) seconds.",
```
SerseusWasTaken: Nitpick: use the actual configured`expirationSecondsDefault` here instead of a hardcoded string value

internal/cmd/beta/edge/token/create/create.go
```diff
@@ -137,17 +122,15 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 }
 
 func configureFlags(cmd *cobra.Command) {
-	cmd.Flags().StringP(commonInstance.InstanceIdFlag, commonInstance.InstanceIdShorthand, "", commonInstance.InstanceIdUsage)
-	cmd.Flags().StringP(commonInstance.DisplayNameFlag, commonInstance.DisplayNameShorthand, "", commonInstance.DisplayNameUsage)
-	cmd.Flags().StringP(commonKubeconfig.ExpirationFlag, commonKubeconfig.ExpirationShorthand, "", commonKubeconfig.ExpirationUsage)
+	cmd.Flags().String(instanceIdFlag, "", "Edge Cloud instance ID")
+	cmd.Flags().StringP(expirationFlag, "e", "", "Expiration time for the token, e.g. 5d. By default, the token is valid for 1h.")
```
SerseusWasTaken: Here use `expirationSecondsDefault` as well instead of hardcoded string

internal/cmd/beta/edge/token/create/create.go
```diff
@@ -137,17 +122,15 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-	identifierFlags := []string{commonInstance.InstanceIdFlag, commonInstance.DisplayNameFlag}
-	cmd.MarkFlagsMutuallyExclusive(identifierFlags...) // InstanceId xor DisplayName
-	cmd.MarkFlagsOneRequired(identifierFlags...)
+	err := flags.MarkFlagsRequired(cmd, instanceIdFlag)
+	cobra.CheckErr(err)
 }
 
 // Parse user input (arguments and/or flags)
-func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
+func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel, error) {
```
SerseusWasTaken: The last `_ []string` parameter seems to be unnecassary

GokceGK: I removed it and adapted the related unit test. Changed also the kubeconfig create command as same implementation was in there.

internal/cmd/beta/edge/instance/create/create.go
GokceGK: ah yes, you are right. We sometimes used that style but here it is not necessary at all, removed

internal/cmd/beta/edge/kubeconfig/create/create.go
SerseusWasTaken: yeah if we wanted to use `p.OutputResult` here we would need to change the logic of `outputResult`

internal/cmd/beta/edge/token/create/create.go
GokceGK: thx, moved :)

internal/cmd/beta/edge/instance/create/create.go
```diff
@@ -60,169 +64,94 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
+	planIdValue := flags.FlagToStringValue(p, cmd, planIdFlag)
+	descriptionValue := flags.FlagToStringPointer(p, cmd, descriptionFlag)
 
 	model := inputModel{
 		GlobalFlagModel: globalFlags,
-		DisplayName:     *displayNameValue,
+		DisplayName:     displayNameValue,
 		Description:     descriptionValue,
-		PlanId:          *planIdValue,
+		PlanId:          planIdValue,
```
rubenhoenle:
```suggestion
		PlanId:          flags.FlagToStringValue(p, cmd, planIdFlag),
```

Nitpick: You're creating 3 variables above to use them once here, why not to everything directly when initializing the struct? 😅

internal/cmd/beta/edge/instance/delete/delete.go
```diff
@@ -89,155 +59,67 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
-	case commonInstance.DisplayNameFlag:
-		factory := func(c *edge.APIClient) instanceWaiter {
-			return wait.DeleteInstanceByNameWaitHandler(ctx, c, model.ProjectId, model.Region, model.identifier.Value)
-		}
-		return factory, nil
-	default:
-		return nil, commonErr.NewInvalidIdentifierError(model.identifier.Flag)
-	}
+func buildRequest(ctx context.Context, model *inputModel, apiClient *edge.APIClient) edge.ApiDeleteInstanceRequest {
+	req := apiClient.DefaultAPI.DeleteInstance(ctx, model.ProjectId, model.Region, model.InstanceId)
```
rubenhoenle:
```suggestion
	return apiClient.DefaultAPI.DeleteInstance(ctx, model.ProjectId, model.Region, model.InstanceId)
```

Creating a variable for returning it in the next line seems odd.

internal/cmd/beta/edge/instance/list/list.go
```diff
@@ -77,17 +62,23 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
 			if err != nil {
 				params.Printer.Debug(print.ErrorLevel, "get project name: %v", err)
-				// If project label can't be determined, fall back to project ID
 				projectLabel = model.ProjectId
 			}
 
 			// Call API
-			resp, err := run(ctx, model, apiClient)
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
```
rubenhoenle:
```suggestion
			resp, err := buildRequest(ctx, model, apiClient).Execute()
```

internal/cmd/beta/edge/instance/list/list.go
```diff
@@ -120,71 +109,30 @@ func parseInput(p *print.Printer, cmd *cobra.Command) (*inputModel, error) {
...
-	req := apiClient.ListInstances(ctx, model.ProjectId, model.Region)
-
-	return &listRequestSpec{
-		ProjectID: model.ProjectId,
-		Region:    model.Region,
-		Limit:     model.Limit,
-		Execute:   req.Execute,
-	}, nil
+func buildRequest(ctx context.Context, model *inputModel, apiClient *edge.APIClient) edge.ApiListInstancesRequest {
+	req := apiClient.DefaultAPI.ListInstances(ctx, model.ProjectId, model.Region)
```
rubenhoenle:
```suggestion
	return apiClient.DefaultAPI.ListInstances(ctx, model.ProjectId, model.Region)
```

same here

internal/cmd/beta/edge/kubeconfig/create/create.go
rubenhoenle: Fair point, my bad. Sorry. 

Then you can just resolve this comment 😊

rubenhoenle (dismissed):

internal/cmd/beta/edge/instance/create/create.go
GokceGK: ah you are totally right 😅 before we had some validations where variables were used, but after cleanup I forgot to remove the variables

SerseusWasTaken (approved):

PR #1505, opened 2026-07-28
feat(postgresflex): migrate to v3 API

rubenhoenle:
## Description

relates to STACKITCLI-415

## Checklist

- [ ] Issue was linked above


internal/cmd/postgresflex/instance/create/create.go
```diff
@@ -168,98 +233,122 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel,
...
-		Acl:            postgresflex.ACL{Items: model.ACL},
-		BackupSchedule: model.BackupSchedule,
-		FlavorId:       flavorId,
-		Replicas:       replicas,
-		Storage: postgresflex.Storage{
-			Class: model.StorageClass,
-			Size:  model.StorageSize,
+		Network: postgresflex.InstanceNetworkCreate{
+			Acl: model.ACL,
 		},
```
marceljk: There is also a new additional property `accessScope`. AFAIK this is needed for the private endpoints feature

internal/cmd/postgresflex/backup/describe/describe.go
```diff
@@ -88,13 +83,18 @@ func configureFlags(cmd *cobra.Command) {
...
 
 	globalFlags := globalflags.Parse(p, cmd)
 	if globalFlags.ProjectId == "" {
 		return nil, &errors.ProjectIdError{}
 	}
 
+	backupId, err := strconv.ParseInt(backupIdStr, 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("invalid backup id format, must be an integer: %w", err)
+	}
```
marceljk, line 96: nitpick: you can also pass a validate function in the call of `args.SingleArg(backupIdArg, <function>)` and check there if the input is valid.
But it's anyways required to parse here the string to int, so the validate function wouldn't have a big benefit.

marceljk: Feel free to just close the comment, if you don't want to add the validate function

internal/cmd/postgresflex/backup/list/list.go
```diff
@@ -129,29 +125,32 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *postgresfle
...
 		if len(backups) == 0 {
 			p.Outputf("No backups found for instance %q", instanceLabel)
 			return nil
 		}
+
 		table := tables.NewTable()
-		table.SetHeader("ID", "CREATED AT", "EXPIRES AT", "BACKUP SIZE")
-		for i := range backups {
-			backup := backups[i]
+		table.SetHeader("ID", "CREATED AT", "RETAINED UNTIL", "BACKUP SIZE")
```
marceljk:
```suggestion
		table.SetHeader("ID", "COMPLETED AT", "RETAINED UNTIL", "BACKUP SIZE")
```

internal/cmd/postgresflex/instance/clone/clone.go
```diff
@@ -115,8 +130,10 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 
 func configureFlags(cmd *cobra.Command) {
 	cmd.Flags().String(recoveryTimestampFlag, "", "Recovery timestamp for the instance, in a date-time with the layout format YYYY-MM-DDTHH:mm:ss±HH:mm, e.g. 2006-01-02T15:04:05-07:00")
-	cmd.Flags().String(storageClassFlag, "", "Storage class. If not specified, storage class from the existing instance will be used.")
-	cmd.Flags().Int64(storageSizeFlag, 0, "Storage size (in GB). If not specified, storage size from the existing instance will be used.")
+	cmd.Flags().String(storageClassFlag, "", "Storage class. If not specified, storage class from the existing instance will be used. This flag will be required after 2027-01-31.")
```
marceljk, line 133: Didn't test the endpoint here, but based on the API Docs only storageSize is required, but not storageClass. Do we need to make this required in the future or what will happen if isn't just not set?

internal/cmd/postgresflex/instance/create/create.go
```diff
@@ -149,17 +193,38 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
 	typeFlag.Register(cmd.Flags())
 
+	// after 2027-01-31: add backup-schedule,storage-size,version,flavor-id,retention-days
 	err := flags.MarkFlagsRequired(cmd, instanceNameFlag, aclFlag)
 	cobra.CheckErr(err)
+
+	cmd.MarkFlagsRequiredTogether(encryptionKekKeyIdFlag, encryptionKekKeyringIdFlag, encryptionKekKeyVersionFlag, encryptionServiceAccountFlag)
+
+	// remove after 2027-01-31
+	err = cmd.Flags().MarkDeprecated("type", fmt.Sprintf("Will be removed after 2027-01-31. Use the --%s flag instead.", flavorIdFlag))
```
marceljk:
```suggestion
	err = cmd.Flags().MarkDeprecated(typeFlag.Name(), fmt.Sprintf("Will be removed after 2027-01-31. Use the --%s flag instead.", flavorIdFlag))
```

internal/cmd/postgresflex/instance/update/update.go
```diff
@@ -133,12 +140,28 @@ func configureFlags(cmd *cobra.Command) {
...
 	cmd.Flags().String(versionFlag, "", "Version")
+	cmd.Flags().String(retentionDaysFlag, "", "The days for how long the backup files should be stored before cleaned up (32 to 90).")
 	typeFlag.Register(cmd.Flags())
+
+	// remove after 2027-01-31
+	err := cmd.Flags().MarkDeprecated(storageClassFlag, "This flag has no effect and will be removed after 2027-01-31.")
+	cobra.CheckErr(err)
+
+	// remove after 2027-01-31
+	err = cmd.Flags().MarkDeprecated("type", fmt.Sprintf("Will be removed after 2027-01-31. Use the --%s flag instead.", flavorIdFlag))
```
marceljk:
```suggestion
	err = cmd.Flags().MarkDeprecated(typeFlag.Name(), fmt.Sprintf("Will be removed after 2027-01-31. Use the --%s flag instead.", flavorIdFlag))
```

internal/pkg/services/postgresflex/utils/utils.go
```diff
@@ -71,14 +71,18 @@ func ValidateFlavorId(flavorId string, flavors []postgresflex.Flavor) error {
 	}
 }
 
-func ValidateStorage(storageClass *string, storageSize *int64, storages *postgresflex.ListStoragesResponse, flavorId string) error {
-	if storages == nil {
-		return fmt.Errorf("nil storages")
+/*func ValidateStorage(storageClass *string, storageSize *int64, flavors []postgresflex.ListFlavors, flavorId string) error {
```
marceljk: leftover

internal/pkg/services/postgresflex/utils/utils.go
```diff
@@ -96,30 +100,27 @@ func ValidateStorage(storageClass *string, storageSize *int64, storages *postgre
...
-		if f.Id == nil || f.Cpu == nil || f.Memory == nil {
-			continue
-		}
-		if *f.Cpu == cpu && *f.Memory == ram {
-			return f.Id, nil
+		if f.Cpu == cpu && f.Memory == ram {
+			return &f.Id, nil
 		}
-		availableFlavors = fmt.Sprintf("%s\n- %d CPU, %d GB RAM", availableFlavors, *f.Cpu, *f.Cpu)
+		availableFlavors = fmt.Sprintf("%s\n- %d CPU, %d GB RAM", availableFlavors, f.Cpu, f.Cpu)
```
marceljk:
```suggestion
		availableFlavors = fmt.Sprintf("%s\n- %d CPU, %d GB RAM", availableFlavors, f.Cpu, f.Memory)
```

marceljk: availableFlavors isn't used at all, so this can be removed

internal/cmd/postgresflex/instance/update/update.go
```diff
@@ -104,24 +110,25 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
...
+			err = req.Execute()
 			if err != nil {
 				return fmt.Errorf("update PostgreSQL Flex instance: %w", err)
 			}
-			instanceId := *resp.Item.Id
+
+			var waitResp *postgresflex.GetInstanceResponse
 
 			// Wait for async operation, if async mode not enabled
 			if !model.Async {
```
marceljk: When the update command is executed with the flag --async, the waitResp remains nil and results in an error output

internal/cmd/postgresflex/backup/list/list.go
```diff
@@ -80,7 +76,7 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("get backups for PostgreSQL Flex instance %q: %w", instanceLabel, err)
 			}
-			backups := resp.Items
+			backups := resp.Backups
 
 			// Truncate output
 			if model.Limit != nil && len(backups) > int(*model.Limit) {
```
marceljk: The used endpoint supports setting a page size. I think at this point we can just set the page size with the limit. Then we don't need to truncate the output manually and fetch the right amount of results

marceljk: same applies to the other list endpoints

internal/cmd/postgresflex/instance/clone/clone.go
rubenhoenle: Storage class is also required, it's wrong in the API docs. I was in contact with the postgresflex team about this.

internal/pkg/services/postgresflex/utils/utils.go
GokceGK: Client is not used anymore, we can remove it

internal/pkg/services/postgresflex/utils/utils.go
```diff
@@ -71,55 +71,24 @@ func ValidateFlavorId(flavorId string, flavors []postgresflex.Flavor) error {
 	}
 }
```
GokceGK: GetInstanceReplicas, GetInstanceType and ValidateFlavorId functions are not used anymore. We can clean up

internal/pkg/services/postgresflex/utils/utils_test.go
GokceGK: This test can be cleaned up after related util function is deleted

internal/pkg/services/postgresflex/utils/utils_test.go
```diff
@@ -57,128 +56,20 @@ func newAPIMockClient(s mockSettings) postgresflex.DefaultAPI {
...
-				t.Fatalf("should not have failed: %v", err)
-			}
-			if !tt.isValid && err == nil {
-				t.Fatalf("should have failed")
-			}
-		})
-	}
-}
-
 func TestValidateFlavorId(t *testing.T) {
```
GokceGK: This test can be cleaned up after related util function is deleted

internal/pkg/services/postgresflex/utils/utils.go
rubenhoenle: you're right, thanks

internal/pkg/services/postgresflex/utils/utils.go
rubenhoenle: You're right, thanks.

internal/pkg/services/postgresflex/utils/utils_test.go
rubenhoenle: should be done :)

internal/pkg/services/postgresflex/utils/utils_test.go
rubenhoenle: should be done :)

GokceGK (approved):

PR #1507, opened 2026-07-29
fix(network): fix nil pointer dereference when dns name servers are n…

SerseusWasTaken:
…ot set during update

## Description
Fixes a nil pointer dereference when no dns servers are configured for the update.

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/network/update/update.go
+++ b/internal/cmd/network/update/update.go
@@ -180,8 +180,10 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *iaas.APICli
 	var payloadIPv6 *iaas.UpdateNetworkIPv6Body
 
 	if model.IPv6DnsNameServers != nil || model.NoIPv6Gateway || model.IPv6Gateway != nil {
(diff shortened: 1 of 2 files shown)


internal/cmd/network/update/update.go
```diff
@@ -192,8 +194,10 @@ func buildRequest(ctx context.Context, model *inputModel, apiClient *iaas.APICli
 	}
 
 	if model.IPv4DnsNameServers != nil || model.NoIPv4Gateway || model.IPv4Gateway != nil {
-		payloadIPv4 = &iaas.UpdateNetworkIPv4Body{
-			Nameservers: *model.IPv4DnsNameServers,
+		payloadIPv4 = &iaas.UpdateNetworkIPv4Body{}
+
+		if model.IPv4DnsNameServers != nil {
+			payloadIPv4.Nameservers = *model.IPv4DnsNameServers
```
marceljk, line 200: We can add here an additional unit test, to cover this issue with the nil pointer dereference, to avoid that this happens again in the future

marceljk (approved):

Issue #1478, opened 2026-07-16
`project create` fails when only `STACKIT_ACCESS_TOKEN` is set


philross:
## Description

When using `STACKIT_ACCESS_TOKEN` in container environments, certain CLI commands (e.g. `project create`) fail despite the token being valid.

## Steps to reproduce
1. Set `STACKIT_ACCESS_TOKEN` as an environment variable with a valid token
2. Pass this token to a container, where the stackit-cli ist installed, but not configured otherwise.
3. Run `stackit project create --parent-id xxxx --name my-project`

## Actual behavior

The command fails with:
`Error: build project creation request: get authentication flow: read from keyring: exec: "dbus-launch": executable file not found in $PATH, read from encoded file as fallback: value not found`

Even though `STACKIT_ACCESS_TOKEN` is set and API calls to other endpoints work fine.

## Expected behavior

When `STACKIT_ACCESS_TOKEN` is set, all cli command should be possible.

## Environment
 - OS: alpine:3.24.0
 - Version of STACKIT CLI (see `stackit --version`): `0.63.0`

**Additional information**

After looking into it, I saw that  `project create` additionally calls [`auth.GetAuthFlow()`](https://github.com/stackitcloud/stackit-cli/blob/8c9edfa66974a266a4b13f1d9460205ea4e2b1bd/internal/cmd/project/create/create.go#L151) directly in `buildRequest()` to determine the owner email for the project membership payload. Seems that `GetAuthFlow()` is not aware of `STACKIT_ACCESS_TOKEN` and always reads `auth_flow_type` exclusively from the auth storage backend (keyring or fallback file).

Fyusel:
Hi @philross,
thank you for reporting this issue. We will have a look at this.

philross:
Closing as issue should be fixed with #1521

PR #1521, opened 2026-08-04
fix: fix auth flow when creating project with access token

SerseusWasTaken:
## Description
Fixes the auth flow for creating projects to be able to use STACKIT_ACCESS_TOKEN for project creation

Relates to STACKITTPR-761 and #1478 

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/internal/cmd/project/create/create.go
+++ b/internal/cmd/project/create/create.go
@@ -148,29 +148,9 @@ func parseInput(p *print.Printer, cmd *cobra.Command, _ []string) (*inputModel, 
 func buildRequest(ctx context.Context, model *inputModel, apiClient *resourcemanager.APIClient) (resourcemanager.ApiCreateProjectRequest, error) {
 	req := apiClient.DefaultAPI.CreateProject(ctx)
 
-	authFlow, err := auth.GetAuthFlow()
+	email, err := auth.GetAuthEmail()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/pkg/auth/storage.go
```diff
@@ -234,6 +234,9 @@ func GetAuthFieldMap(keyMap map[authFieldKey]string) error {
 }
 
 func GetAuthFlow() (AuthFlow, error) {
+	if accessToken := os.Getenv(envAccessTokenName); accessToken != "" {
+		return AUTH_FLOW_SERVICE_ACCOUNT_TOKEN, nil
```
cgoetz-inovex: I'm not quite sure about this one. Decisions on which AuthFlow to use seem to be done in functions where `SetAuthFlow()` is called.

What's the reason to include this change?

cgoetz-inovex (dismissed):

internal/pkg/auth/storage.go
SerseusWasTaken: My intention was to guarantee that the access token is used when it is set. Before, it as possible to have the token set and still be treated unauthenticated by the CLI which why the original issue came up. I personally don't see a reason not to use the token when it is set.

With this change, `GetAuthFlow` is now aware of the token's existence. Does that make sense?

cgoetz-inovex (approved):

PR #1530, opened 2026-08-11
feat(valkey): onboard valkey

SerseusWasTaken:
## Description

Relates to STACKITCLI-428

## Checklist

- [ ] Issue was linked above


internal/cmd/root.go
```diff
@@ -203,6 +204,7 @@ func addSubcommands(cmd *cobra.Command, params *types.CmdParams) {
 	cmd.AddCommand(affinityGroups.NewCmd(params))
 	cmd.AddCommand(git.NewCmd(params))
 	cmd.AddCommand(kms.NewCmd(params))
+	cmd.AddCommand(valkey.NewCmd(params))
```
Fyusel, line 209: Currently the valkey service is registered as beta command and "normal" command. Therefore the documentation is also doubled. The service should only be present either in beta or root.

Since this is just a copy of the redis I would just move it out of beta and put it in root

internal/cmd/beta/valkey/instance/create/create.go
```diff
@@ -0,0 +1,266 @@
...
+	cmd.Flags().Int32(metricsFrequencyFlag, 0, "Metrics frequency in seconds")
+	cmd.Flags().String(metricsPrefixFlag, "", "Metrics prefix")
+	cmd.Flags().Var(flags.UUIDFlag(), monitoringInstanceIdFlag, "Monitoring instance ID")
+	cmd.Flags().Var(flags.CIDRSliceFlag(), sgwAclFlag, "List of IP networks in CIDR notation which are allowed to access this instance")
+	cmd.Flags().StringSlice(syslogFlag, []string{}, "Syslog")
+	cmd.Flags().Var(flags.UUIDFlag(), planIdFlag, "Plan ID")
+	cmd.Flags().String(planNameFlag, "", "Plan name")
+	cmd.Flags().String(versionFlag, "", "Instance Valkey version")
+
+	cmd.Flags().Int32(minReplicasToWriteFlag, 0, "Minimum number of replicas that must acknowledge a write for it to be accepted (Valkey only)")
```
Fyusel:
```suggestion
	cmd.Flags().Int32(minReplicasToWriteFlag, 0, "Minimum number of replicas that must acknowledge a write for it to be accepted")
```

internal/cmd/beta/valkey/instance/create/create.go
```diff
@@ -0,0 +1,266 @@
...
+	cmd.Flags().String(replBacklogSizeFlag, "", "Replication backlog size (e.g. \"1mb\") (Valkey only)")
```
Fyusel:
```suggestion
	cmd.Flags().String(replBacklogSizeFlag, "", "Replication backlog size (e.g. \"1mb\")")
```

internal/cmd/valkey/instance/create/create.go
```diff
@@ -0,0 +1,266 @@
...
+			ReplBacklogSize:      model.ReplBacklogSize,
+		},
+		PlanId: planId,
+	})
+	return req, nil
+}
+
+func outputResult(p *print.Printer, model *inputModel, projectLabel, instanceId string, resp *valkey.CreateInstanceResponse) error {
+	return p.OutputResult(model.OutputFormat, resp, func() error {
+		if model == nil {
```
Fyusel: This check should be made before calling this function otherwise model.OutputFormat would crash if model is nil

internal/cmd/beta/valkey/plans/plans.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/types"
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+}
+
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "offerings",
```
Fyusel: This is a rename of the existing command. I see that in the API docs this is called offerings instead of plans.
Currently this is a mismatch (file name "plans" versus command name "offerings").

We should check why this was renamed or why this was called "plans" in the first place. Independent on this the file name and the command name should match

internal/cmd/valkey/plans/plans.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
+}
+
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "offerings",
+		Short: "Lists all Valkey service plans",
```
Fyusel, line 34: Here it is still called plans (same for the rest of the file)

internal/cmd/valkey/plans/plans.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/globalflags"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/print"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/projectname"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/services/valkey/client"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/tables"
+	"github.com/stackitcloud/stackit-cli/internal/pkg/types"
+)
+
+type inputModel struct {
+	*globalflags.GlobalFlagModel
```
Fyusel, line 27: Why was the limit flag removed?

internal/cmd/beta/valkey/plans/plans.go
Fyusel: As far as I can see right now, _plans_ was used since the _offerings_ have a _plans_ array which is displayed here.

internal/cmd/valkey/instance/create/create.go
SerseusWasTaken: Where would it crash? I don't think `model` is accessed between passing and this check

internal/cmd/beta/valkey/plans/plans.go
SerseusWasTaken: Yeah good point, i didn't double check that after going with the API

internal/cmd/valkey/plans/plans.go
SerseusWasTaken: my bad, i'll re-add it

internal/cmd/valkey/instance/create/create.go
Fyusel: If _model_ is nil, then the call to this function would crash by passing _model.OutputFormat_

SerseusWasTaken:
Currently not testable due to API-issue. We will have to wait for a fix

internal/cmd/valkey/credentials/create/create.go
```diff
@@ -0,0 +1,163 @@
...
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates credentials for a Valkey instance",
+		Long:  "Creates credentials (username and password) for a Valkey instance.",
+		Args:  args.NoArgs,
+		Example: examples.Build(
+			examples.NewExample(
+				`Create credentials for a Valkey instance with ID "xxx"`,
+				"$ stackit beta valkey credentials create --instance-id xxx"),
```
Fyusel: The examples need some updates: "beta" needs to be removed (all other files as well)

SerseusWasTaken: good catch

Fyusel (approved):

PR #1533, opened 2026-08-13
feat(albwaf): onboard albwaf custom rule group

SerseusWasTaken:
## Description
Adds albwaf custom rule group support

Relates to STACKITCLI-443

## Checklist

- [ ] Issue was linked above


internal/cmd/beta/albwaf/albwaf.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/stackitcloud/stackit-cli/internal/pkg/utils"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmd(params *types.CmdParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "alb-waf",
+		Short: "Provides functionality for Application Load Balancer Web Application Firwall",
+		Long:  "Provides functionality for Application Load Balancer Web Application Firwall.",
```
cgoetz-inovex:
```suggestion
		Short: "Provides functionality for Application Load Balancer Web Application Firewall",
		Long:  "Provides functionality for Application Load Balancer Web Application Firewall.",
```

internal/pkg/config/config.go
```diff
@@ -25,6 +25,7 @@ const (
 
 	AuthorizationCustomEndpointKey     = "authorization_custom_endpoint"
 	AlbCustomEndpoint                  = "alb_custom _endpoint"
+	AlbWafCustomEndpointKey            = "albwaf_custom_endpoint"
```
cgoetz-inovex: There's some things missing:
- `internal/cmd/config/set/set.go` should call `viper.BindPFlag`
- `internal/cmd/config/unset.go` shoudl call `viper.Set(..., "")`
- in this file there's `setConfigDefaults()`, which should call `viper.SetDefault()`

internal/cmd/beta/albwaf/custom-rule-group/list/list.go
```diff
@@ -0,0 +1,183 @@
...
+	for {
+		want := min(int64(maxPageSize), limit-received)
+		request := buildRequest(ctx, model, apiClient, pageId, want)
+		response, err := request.Execute()
+		if err != nil {
+			return nil, fmt.Errorf("list custom rule groups: %w", err)
+		}
+		if response.Items != nil {
+			items = append(items, response.Items...)
+		}
```
cgoetz-inovex: nitpick, conditional not needed

```suggestion
		items = append(items, response.Items...)
```

```go
func TestNilSplat(t *testing.T) {
	var f []string
	var s []string = nil
	f = append(f, s...)
}
```

internal/cmd/beta/alb-waf/custom-rule-group/list/list.go
```diff
@@ -0,0 +1,183 @@
...
+			return nil, fmt.Errorf("list custom rule groups: %w", err)
+		}
+		if response.Items != nil {
+			items = append(items, response.Items...)
+		}
+		pageId = ""
+		if response.NextPageId != nil {
+			pageId = *response.NextPageId
+		}
+		received += want
```
cgoetz-inovex, line 144:
```suggestion
		received += len(response.Items)
```

internal/cmd/beta/alb-waf/custom-rule-group/list/list.go
```diff
@@ -0,0 +1,183 @@
...
+func buildRequest(ctx context.Context, model *inputModel, apiClient *albwaf.APIClient, pageId string, pageSize int64) albwaf.ApiListCustomRuleGroupRequest {
+	req := apiClient.DefaultAPI.ListCustomRuleGroup(ctx, model.ProjectId, model.Region)
+	req = req.PageSize(strconv.FormatInt(pageSize, 10))
+	if pageId != "" {
+		req = req.PageId(pageId)
+	}
+	return req
+}
+
+func fetchCustomRuleGroups(ctx context.Context, model *inputModel, apiClient *albwaf.APIClient) ([]albwaf.GetCustomRuleGroupResponse, error) {
```
cgoetz-inovex, line 123: this looks pagination AIP compliant, we could add a comment to replace this with the utils function we want to do in STACKITSDK-525

internal/cmd/beta/albwaf/custom-rule-group/list/list.go
```diff
@@ -0,0 +1,183 @@
...
+	req = req.PageSize(strconv.FormatInt(pageSize, 10))
+	if pageId != "" {
+		req = req.PageId(pageId)
+	}
+	return req
+}
+
+func fetchCustomRuleGroups(ctx context.Context, model *inputModel, apiClient *albwaf.APIClient) ([]albwaf.GetCustomRuleGroupResponse, error) {
+	var pageId string
+	var items []albwaf.GetCustomRuleGroupResponse
```
cgoetz-inovex:
```suggestion
	items := make([]albwaf.GetCustomRuleGroupResponse, 0)
```
this is passed to `outputResult`, for `-o json` it makes a difference if we pass an empty slice vs a nil slice. `null` vs `[]`. The empty slice is nicer to handle in scripts.

internal/cmd/beta/albwaf/custom-rule-group/create/create.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			_, err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("create ALB WAF custom rule group: %w", err)
+			}
+
+			params.Printer.Outputf("Created custom rule group with name %q \n", model.Payload.Name)
```
cgoetz-inovex: call `params.Printer.OutputResult`, this currently breaks scripting with `-o json/yaml`

internal/cmd/beta/albwaf/custom-rule-group/delete/delete.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			_, err = req.Execute()
+			if err != nil {
+				return fmt.Errorf("delete ALB WAF custom rule group: %w", err)
+			}
+
+			params.Printer.Outputf("Custom rule group %q deleted.\n", model.Name)
```
cgoetz-inovex:
```suggestion
			params.Printer.Info("Custom rule group %q deleted.\n", model.Name)
```

internal/cmd/beta/albwaf/custom-rule-group/update/update.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+
+			// Call API
+			req := buildRequest(ctx, model, apiClient)
+			resp, err := req.Execute()
+			if err != nil {
+				return fmt.Errorf("update ALB WAF custom rule group: %w", err)
+			}
+
+			params.Printer.Outputf("Updated custom rule group %q.\n", resp.Name)
```
cgoetz-inovex: `params.Printer.OutputResult` see create.go comment

internal/cmd/beta/alb-waf/custom-rule-group/list/list.go
SerseusWasTaken: Oh yeah, good idea!

internal/pkg/config/config.go
SerseusWasTaken: good catch, thanks

internal/cmd/beta/alb-waf/custom-rule-group/list/list.go
SerseusWasTaken: Not sure about that one. I've applied the same logic as in `CDN distribution list` and it seems to work correctly. If we change it here, we should also change it for CDN.

Since we will replace it with a shared sdk function anyway in the near future, i would suggest to keep it like CDN has it for now.

cgoetz-inovex: Suggestion with the shared function, sounds pragmatic.
Current impl can only lead to weird behaviour if the API behaves in a buggy way, so waiting is fine for me

cgoetz-inovex (dismissed):

cgoetz-inovex (approved):

PR #1588, opened 2026-09-08
feat(valkey): add Key Value Store to valkey descriptions

GokceGK:
## Description

relates to STACKITTPR-845
## Checklist

- [x] Issue was linked above

### DIFF
--- a/internal/cmd/redis/redis.go
+++ b/internal/cmd/redis/redis.go
@@ -15,7 +15,7 @@ import (
 func NewCmd(params *types.CmdParams) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:        "redis",
-		Deprecated: "\nRedis commands `stackit redis` have been deprecated and will be removed after 2027-08-31. Please use valkey (Key Value Store) commands `stackit valkey` instead.",
+		Deprecated: "\nRedis commands `stackit redis` have been deprecated and will be removed after 2027-08-31. Please use Key Value Store (valkey) commands `stackit valkey` instead.",
 		Short:      "Provides functionality for Redis",
 		Long:       "Provides functionality for Redis.",
 		Args:       args.NoArgs,
--- a/internal/cmd/valkey/credentials/create/create.go
+++ b/internal/cmd/valkey/credentials/create/create.go
@@ -40,10 +40,10 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 		Args:  args.NoArgs,
 		Example: examples.Build(
 			examples.NewExample(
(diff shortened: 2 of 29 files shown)


internal/cmd/valkey/credentials/create/create.go
```diff
@@ -75,7 +75,7 @@ func NewCmd(params *types.CmdParams) *cobra.Command {
 			req := buildRequest(ctx, model, apiClient.DefaultAPI)
 			resp, err := req.Execute()
 			if err != nil {
-				return fmt.Errorf("create Valkey credentials: %w", err)
+				return fmt.Errorf("create Key Value Store (valkey) credentials: %w", err)
```
rubenhoenle: Same as in the Terraform provider.

I would suggest to adjust only the command **descriptions** and leave everything else as it is. No need to pollute log and error messages with this.

internal/cmd/valkey/credentials/create/create.go
```diff
@@ -35,15 +35,15 @@ type inputModel struct {
...
+		Long:  "Creates credentials (username and password) for a Key Value Store (valkey) instance.",
 		Args:  args.NoArgs,
 		Example: examples.Build(
 			examples.NewExample(
-				`Create credentials for a Valkey instance with ID "xxx"`,
+				`Create credentials for a Key Value Store (valkey) instance with ID "xxx"`,
 				"$ stackit valkey credentials create --instance-id xxx"),
 			examples.NewExample(
-				`Create credentials for a Valkey instance and show the password in the output`,
+				`Create credentials for a Key Value Store (valkey) instance and show the password in the output`,
```
rubenhoenle: When the description already explains that `valkey = Key Value Store` we don't have to repeat this in each and every example IMO.

Keep CLI output compact and readable, don't pollute it.

rubenhoenle (changes requested):

internal/cmd/redis/redis.go
```diff
@@ -15,7 +15,7 @@ import (
 func NewCmd(params *types.CmdParams) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:        "redis",
-		Deprecated: "\nRedis commands `stackit redis` have been deprecated and will be removed after 2027-08-31. Please use valkey (Key Value Store) commands `stackit valkey` instead.",
+		Deprecated: "\nRedis commands `stackit redis` have been deprecated and will be removed after 2027-08-31. Please use Key Value Store (valkey) commands `stackit valkey` instead.",
```
rubenhoenle, line 18: Keep in mind that your terminal won't render the backtick markdown syntax in most cases. @GokceGK 

Image

PR #1602, opened 2026-09-14
feat(auth): add no browser open flag to login

niri81:
## Description

Add an additional option to only print the authorization URL and not open it directly in the standard browser. Especially useful if another browser or an incognito tab need to be used to sign-in with different accounts and bypass automatic SSO login.

## Checklist

- [ ] Issue was linked above


niri81:
Review pending, not stale

